PR #70, opened 2023-05-06
Bugfix/components can't be clicked when zoom out

princepride:
When I zoom out the browser, the components can't be clicked. I noticed that a possible issue might be within the useEffect hook of MainLayout. The code toggles the sidebar state when the window size changes, which may be causing the issue I am encountering, so I comment out the useEffect the problem has solved.


chungyau97:
Hey there @princepride , could you share a detail video of the bug that you're talking about?

princepride:
@chungyau97 

https://github.com/FlowiseAI/Flowise/assets/29850264/6b0d7104-2d51-4543-b4a6-5ac9d418856c

princepride:
As you see, when I zoom out the browser, all components can't be clicked

chungyau97:
After further investigation I found that the `MuiDrawer-modal` drawer is blocking it.
Need to take another approach on solving this.

HenryHengZJ:
I removed the code instead of commenting out, @chungyau97 do you have better solution?

HenryHengZJ:
Downside is that when expand back the drawer will not automatically opened up

HenryHengZJ (approved):

chungyau97 (approved):
adding `setTimeout` solve the issue.

PR #94, opened 2023-05-12
Feature/Add metadata filter

HenryHengZJ:
When upserting, include metadata filter on document loader:
image

When loading from existing index, you can filter by metadata:
image


chungyau97 (approved):
Test Result:
image
image
image

Issue #131, opened 2023-05-20
[BUG] Missing escape character in CURL example


tcpxenos:
**Describe the bug**
The example CURL code is missing a backslash to escape the newline

**To Reproduce**
Open the API Endpoint dialog, open the CURL tab, and choose an API key

**Screenshots**
Example output, missing a backslash after `how are you?"}'`
```
curl http://myip:3000/api/v1/prediction/myendpoint \
     -X POST \
     -d '{"question": "Hey, how are you?"}'
     -H "Authorization: Bearer myapikey"
```

**Setup**

-   OS: Docker on Debian
-   Browser: Chrome

Pathholder1806:
Hey @tcpxenos I would like to contribute to this issue!

PR #138, opened 2023-05-20
Update APICodeDialog.js

ShivangRawat30:
Added backslash after how are you?"}'

solves #131


HenryHengZJ:
@ShivangRawat30 thanks for the fix, can you sign the CLA?

HenryHengZJ:
Hey @ShivangRawat30 unfortunately i dont think this is the correct fix. I'll have a look at it tomorrow. At first glance, this is changing the config example, which is not the bug was

HenryHengZJ (changes requested):
Hey @ShivangRawat30 unfortunately i dont think this is the correct fix. I'll have a look at it tomorrow. At first glance, this is changing the config example, which is not the bug was

HenryHengZJ:
Opened another PR to fix the issue - https://github.com/FlowiseAI/Flowise/pull/145

Update: merged

Issue #146, opened 2023-05-22
Dont forget you need Nodejs previously installed


alonsofv0123:
In order to install flowise you will need to have Nodejs previously installed in your computer. 
https://nodejs.org/en/download
if not, this prompt: npm install -g flowise    could not run

VeryKumar:
I can take this on

PR #148, opened 2023-05-22
Update README.md

abhinav2712:
Added the requirement and instructions for installing node as well
This solves #146


HenryHengZJ:
can you also specify node version has to be larger or equal to 18.15.0?

53jk1 (approved):

PR #155, opened 2023-05-23
Pretty terminal commands 

PrabhuUdurg:
Just added '$' in the beginning of terminal commands
I think it looks better. The project is amazing!


HenryHengZJ:
hey thanks for the idea. However we'd prefer not to add `$` as it makes it hard for other OS like Windows to copy paste the command

PR #159, opened 2023-05-23
feature/BufferWindowMemory

chungyau97:
Results:
image


packages/components/nodes/memory/BufferWindowMemory/BufferWindowMemory.ts
```diff
@@ -0,0 +1,61 @@
...
+                label: 'Input Key',
+                name: 'inputKey',
+                type: 'string',
+                default: 'input'
+            },
+            {
+                label: 'Size',
+                name: 'k',
+                type: 'number',
+                default: '4'
```
HenryHengZJ, line 40: can add a `description: 'window of size k to surface the last k back-and-forths to use as memory.'` so people can understand whats this k for

HenryHengZJ (approved):

PR #162, opened 2023-05-24
style: improve AddNodes click

xianjianlf2:
When I click on a node, there is a gap that causes click on an empty space
like this : 
---
bugfix
---
modified :
bugfixDone


HenryHengZJ (approved):
looks good! thanks for the fix

PR #190, opened 2023-05-26
feature: changes to support ApiChain

vjsai:
- Trying to add support for ApiChain from langchainjs
- Preview of the flow created
Screenshot 2023-05-26 at 1 48 23 PM
- Right now the support is to read the api docs via the document loader.

Things to do additionally:
- Add support for custom headers
- Make changes in the langchain library to support post calls as well and make changes here to support the same.
- Also make changes on langchain to support retriever and make changes for the same here.

This is my first Open Source PR so let me know anything more needs to be added or removed. Thanks in advance.


packages/components/nodes/chains/ApiChain/ApiChain.ts
```diff
@@ -0,0 +1,78 @@
...
+        } else {
+            const res = await chain.run(input)
+            return res
+        }
+    }
+}
+
+const getOpenAPIChain = async (documents: Document[], llm: BaseLanguageModel, options: any = {}) => {
+    const texts = documents.map(({ pageContent }) => pageContent);
+    const apiResponsePrompt = new PromptTemplate({
```
HenryHengZJ, line 66: do you see an improvement from using custom `apiResponsePrompt` instead of the original prompt used in langchain library?

HenryHengZJ:
you can do `yarn lint-fix` to fix all the linting issues

packages/components/nodes/chains/ApiChain/ApiChain.ts
vjsai: @HenryHengZJ I was able to see less number tokens used and was getting meaningful results. May be we can give user an option to choose this and default one? what do you think ?

HenryHengZJ: ah okay i think its fine, just curious thats all. 

I'll make update once this PR is merged to get the POST API in. Thanks for making a head start!

HenryHengZJ:
can we also add headers in? you can refer here - https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/tools/RequestsPost/RequestsPost.ts#L52

vjsai:
Sure will add that functionality and update the PR.

HenryHengZJ (approved):
Looks good! thanks for the PR @vjsai

PR #199, opened 2023-05-27
includes user/pass from .env in docker-compose config

alvin:
custom user/pass config in .env didn't work for me until I added this to docker_compose.yml 

not sure if there are any downsides to this, but changing the values in .env wasn't working for me without it (oddly it still required auth, it just "stuck" on the default example of user / 1234 -- perhaps cached in the docker workflow somehow?)

### DIFF
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -6,6 +6,8 @@ services:
         restart: always
         environment:
             - PORT=${PORT}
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
@alvin if you pull from latest, should be working fine now.

Refer to [here](https://github.com/FlowiseAI/Flowise/blob/main/docker/README.md)

PR #200, opened 2023-05-27
feature: Added support for Faiss vector Index

vjsai:
- Added support for Faiss
Screenshot 2023-05-27 at 1 59 31 PM


packages/server/src/utils/index.ts
```diff
@@ -325,7 +325,10 @@ export const getVariableValue = (paramValue: string, reactFlowNodes: IReactFlowN
  * @returns {INodeData}
  */
 export const resolveVariables = (reactFlowNodeData: INodeData, reactFlowNodes: IReactFlowNode[], question: string): INodeData => {
-    const flowNodeData = cloneDeep(reactFlowNodeData)
+    //skipping cloneDeep of instances
+    //as cloning them gives "Illegal invocation" Exception
+    const flowNodeDataWithoutInstance = cloneDeep(omit(reactFlowNodeData, ['instance']))
+    const flowNodeData = merge(flowNodeDataWithoutInstance, { instance: reactFlowNodeData.instance })
```
vjsai, line 331: This should be ok as instances are created only during init phase. your toughts @HenryHengZJ ?

HenryHengZJ: yeah tested and it works fine

HenryHengZJ:
don't we need any libraries installed - https://js.langchain.com/docs/modules/indexes/vector_stores/integrations/faiss#setup ?

HenryHengZJ:
can you also sign the cla? thanks!

vjsai:
@HenryHengZJ Yes we do need libraries.Should we leave note to end user or should i add dependency ? I have signed cla with one of the user ,the other in my vscode is using git config which hasn't signed.I will fix it.

HenryHengZJ:
You can add the dependency to `packages/components`

vjsai:
@HenryHengZJ Have added the dependency you can review and merge. Thanks!!!

packages/server/src/utils/index.ts
HenryHengZJ: somehow when I tried the branch locally, this caused streaming stopped working:
image

need to look into this

vjsai: @HenryHengZJ In local with this code are you still getting Illegal Invocation Exception ?

HenryHengZJ: yes. I think i'll make another PR on top of this as I'll have to disable streaming when FaiSS is being used.

HenryHengZJ:
Closing this as https://github.com/FlowiseAI/Flowise/pull/209 opened here

Thanks again @vjsai for making a head start!

PR #252, opened 2023-06-04
[server] check `nodeModule.nodeClass` is not undefined before trying to intialize

yaredtsy:


HenryHengZJ:
Thanks @yaredtsy can you sign the CLA?

PR #264, opened 2023-06-06
Add basePath configuration option to support OpenAI mirror 

try-to-fly:
This pull request is aimed at adding a new configuration option, `basePath`, to the project. This is motivated by the need to support an OpenAI mirror site in China. (such as https://api2d.com/)

The `basePath` will allow users in China to conveniently redirect their requests to a local mirror, thus significantly improving access speed and reliability.

Please consider integrating this change to enhance the user experience for a large number of users in China.


HenryHengZJ:
@try-to-fly can you add the same to `OpenAI` node as well?

PR #274, opened 2023-06-08
Feature/Return SourceDocuments

HenryHengZJ:
Add ability to return source documents when using ConversationalRetrievalQAChain:

https://github.com/FlowiseAI/Flowise/assets/26460777/2ba35091-04fa-440b-b3ca-04d7347eb6e4


trepont:
I've been able to

convert my docs into embeddings and stored into pinecone index/namespace
use that index/namespace embedding in pinecone to power a conversational agent
and SEPERATELY, create a number of conversational agents which use collected web browser (directed) or Serp (search) web results to chain them and use ChatGPT for inference
bu i am facing issue with retun SourceDocuments  in this scenario,how to use return SourceDocuments  for multple chains scenario
is that possible either now or in the future?
Thanks in advance.

segelmache:
Hi there, it would be very helpful, if the returned sources would be limited to the sources, the model actually used for its answer. Would that be possible? Otherwise, if it's getting 50 documents from the vector database, all sources of thouse 50 documents would be displayed.

PR #284, opened 2023-06-10
feature/SupportZepMemoryAndAutoSummary

wangerzi:
Support https://docs.getzep.com/sdk/langchain/ in Flowise.

Some limits:
- the unique session id should be changed manually in every conversion, because zep server doesn't support session id reuse now

image

### DIFF
--- a/packages/components/nodes/memory/ZepMemory/ZepMemory.ts
+++ b/packages/components/nodes/memory/ZepMemory/ZepMemory.ts
@@ -39,7 +39,7 @@ class ZepMemory_Memory implements INode {
                 label: 'Session Id',
                 name: 'sessionId',
                 type: 'string',
-                placeholder: 'if empty, chatId will be used automatically',
+                description: 'if empty, chatId will be used automatically',
                 default: '',
                 additionalParams: true
             },
--- a/packages/components/nodes/memory/ZepMemory/ZepMemory.ts
+++ b/packages/components/nodes/memory/ZepMemory/ZepMemory.ts
@@ -36,6 +36,14 @@ class ZepMemory_Memory implements INode {
                 default: true
             },
             {
+                label: 'Session Id',
+                name: 'sessionId',
+                type: 'string',
+                placeholder: 'if empty, chatId will be used automatically',
+                default: '',
+                additionalParams: true
+            },
+            {
                 label: 'Auto Summary Template',
                 name: 'autoSummaryTemplate',
                 type: 'string',
@@ -88,11 +96,13 @@ class ZepMemory_Memory implements INode {
         const inputKey = nodeData.inputs?.inputKey as string
         const autoSummaryTemplate = nodeData.inputs?.autoSummaryTemplate as string
         const autoSummary = nodeData.inputs?.autoSummary as boolean
-        const sessionId = options?.chatId as string
+        const sessionId = nodeData.inputs?.sessionId as string
+
+        const chatId = options?.chatId as string
 
         const obj: ZepMemoryInput = {
             baseURL,
-            sessionId,
+            sessionId: sessionId ? sessionId : chatId,
             aiPrefix,
             humanPrefix,
             returnMessages: true,
--- a/packages/components/nodes/memory/ZepMemory/ZepMemory.ts
+++ b/packages/components/nodes/memory/ZepMemory/ZepMemory.ts
@@ -2,6 +2,7 @@ import { SystemChatMessage } from 'langchain/schema'
 import { INode, INodeData, INodeParams } from '../../../src/Interface'
 import { getBaseClasses } from '../../../src/utils'
 import { ZepMemory, ZepMemoryInput } from 'langchain/memory/zep'
+import { ICommonObject } from '../../../src'
 
 class ZepMemory_Memory implements INode {
     label: string
@@ -27,13 +28,6 @@ class ZepMemory_Memory implements INode {
                 name: 'baseURL',
                 type: 'string',
                 default: 'http://127.0.0.1:8000'
-            },
-            {
-                label: 'Session Id',
-                name: 'sessionId',
-                type: 'string',
-                placeholder: 'unique and manually change in every conversion!',
-                default: ''
             },
             {
                 label: 'Auto Summary',
@@ -86,15 +80,15 @@ class ZepMemory_Memory implements INode {
         ]
     }
 
-    async init(nodeData: INodeData): Promise<any> {
+    async init(nodeData: INodeData, _: string, options: ICommonObject): Promise<any> {
         const baseURL = nodeData.inputs?.baseURL as string
-        const sessionId = nodeData.inputs?.sessionId as string
         const aiPrefix = nodeData.inputs?.aiPrefix as string
         const humanPrefix = nodeData.inputs?.humanPrefix as string
         const memoryKey = nodeData.inputs?.memoryKey as string
         const inputKey = nodeData.inputs?.inputKey as string
         const autoSummaryTemplate = nodeData.inputs?.autoSummaryTemplate as string
         const autoSummary = nodeData.inputs?.autoSummary as boolean
+        const sessionId = options?.chatId as string
 
         const obj: ZepMemoryInput = {
             baseURL,
--- a/packages/components/nodes/memory/ZepMemory/ZepMemory.ts
+++ b/packages/components/nodes/memory/ZepMemory/ZepMemory.ts
@@ -1,0 +1,136 @@
+import { SystemChatMessage } from 'langchain/schema'
+import { INode, INodeData, INodeParams } from '../../../src/Interface'
+import { getBaseClasses } from '../../../src/utils'
+import { ZepMemory, ZepMemoryInput } from 'langchain/memory/zep'
+
+class ZepMemory_Memory implements INode {
+    label: string
+    name: string
+    description: string
+    type: string
+    icon: string
+    category: string
+    baseClasses: string[]
+    inputs: INodeParams[]
+
+    constructor() {
+        this.label = 'Zep Memory'
+        this.name = 'ZepMemory'
+        this.type = 'ZepMemory'
+        this.icon = 'memory.svg'
+        this.category = 'Memory'
+        this.description = 'Summarizes the conversation and stores the memory in zep server'
+        this.baseClasses = [this.type, ...getBaseClasses(ZepMemory)]
+        this.inputs = [
+            {
+                label: 'Base URL',
+                name: 'baseURL',
+                type: 'string',
+                default: 'http://127.0.0.1:8000'
+            },
+            {
+                label: 'Session Id',
+                name: 'sessionId',
+                type: 'string',
+                placeholder: 'unique and manually change in every conversion!',
+                default: ''
+            },
+            {
+                label: 'Auto Summary',
+                name: 'autoSummary',
+                type: 'boolean',
+                default: true
+            },
+            {
+                label: 'Auto Summary Template',
+                name: 'autoSummaryTemplate',
+                type: 'string',
+                default: 'This is the summary of the following conversation:\n{summary}',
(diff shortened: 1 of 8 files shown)


wangerzi:
[zep-test Chatflow.zip](https://github.com/FlowiseAI/Flowise/files/11711502/zep-test.Chatflow.zip)
a simple chatflow demo.

HenryHengZJ:
thanks @wangerzi for the Zep addition!
can you help me understand the session id? what do you mean by should be changed manually in every conversion?

wangerzi:
> thanks @wangerzi for the Zep addition! can you help me understand the session id? what do you mean by should be changed manually in every conversion?

@HenryHengZJ Flowise is an interesting project, I'm very happy to make some contributions.😀

The session id in zep just like a unique identity of a conversion, zep server use session id to find memory, find summary for a request.

But when we want to create a new conversion in flowise, we must change the session id (or automatic generate a uuid for new conversion), because we can't **clear** zep chat history in this session by api.
image

Talk is cheap, I'm already prepared a demo to explain it:

Execution Prepare:
- run a zep server locally, refer: https://docs.getzep.com/deployment/quickstart/
- manually install `langchain` and `@getzep/zep-js`

```js
// import { ZepMemory } from 'langchain/memory/zep'
// import { Memory, Message } from '@getzep/zep-js'
const { ZepMemory } = require('langchain/memory/zep')
const { Memory, Message } = require('@getzep/zep-js')

const sessionId = 'test-3'
const zepURL = 'http://localhost:8000'

const memory = new ZepMemory({
    sessionId,
    memoryKey: 'chat_history',
    returnMessages: true,
    baseURL: zepURL
})

async function init() {
    const history = [
        {
            role: 'human',
            content: 'Who was Octavia Butler?'
        },
        {
            role: 'ai',
            content: 'Octavia Estelle Butler (June 22, 1947 – February 24, 2006) was an American' + ' science fiction author.'
        }
    ]

    const messages = history.map(({ role, content }) => new Message({ role, content }))
    const zepMemory = new Memory({ messages })

    await memory.zepClient.addMemory(sessionId, zepMemory)
}

async function testSearch() {
    await init()
    const data = await memory.loadMemoryVariables({
        lastN: 10
    })
    console.log('search data:', data)
}

async function testClear() {
    await memory.clear()
}

async function testCase1() {
    console.log('first search')
    await testSearch()
    await testClear()
    console.log('already cleared')
    await testSearch()
}

testCase1()
```

Execution Result told us if we can't use official api `memory.clear()` to reuse a same session id:
```bash
$ node test-zep.js
first search
search data: {
  chat_history: [
    ChatMessage {
      text: 'Who was Octavia Butler?',
      name: undefined,
      role: 'human'
    },
    ChatMessage {
      text: 'Octavia Estelle Butler (June 22, 1947 – February 24, 2006) was an American science fiction author.',
      name: undefined,
      role: 'ai'
    }
  ]
}
already cleared
D:\phpStudy\WWW\github\Flowise\node_modules\@getzep\zep-js\dist\zep-client.js:143
                    throw new exceptions_1.UnexpectedResponseError(`addMemoryAsync got an Unexpected status code: ${error.response.status}`);
                          ^

UnexpectedResponseError: addMemoryAsync got an Unexpected status code: 400
    at ZepClient.<anonymous> (D:\phpStudy\WWW\github\Flowise\node_modules\@getzep\zep-js\dist\zep-client.js:143:27)
    at Generator.throw (<anonymous>)
    at rejected (D:\phpStudy\WWW\github\Flowise\node_modules\@getzep\zep-js\dist\zep-client.js:29:65)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  responseData: undefined
}

Node.js v18.16.0
```

References:

- Langchain source code: https://github.com/hwchase17/langchainjs/blob/main/langchain/src/memory/zep.ts#L143-L152
- Zep document: https://www.getzep.com/zep-langchainjs-memory/

HenryHengZJ:
@wangerzi oh okay in that case, can we just use timestamp string as session id? that way we can save user from inputting/changing session id everytime

wangerzi:
@HenryHengZJ Sounds like a great idea to solve this problem👏, I'll try to use the conversion start time as zep session after 8 hours  (after wake up🌞)

wangerzi:
@HenryHengZJ According to the current data structure, I think the first chat message id is a better choice for `chat_id`

For generality, I found the first chat message in backend, and pass it by `options` in `INode` definition `init?(nodeData: INodeData, input: string, options?: ICommonObject): Promise<any>`

This is the minimum modification in my mind, but I'm not sure if you have better solution for this.

image

Key Code:
- https://github.com/FlowiseAI/Flowise/pull/284/files#diff-8e6dfbbfb522575fe7c568c2de518bf9de351de83c9dbb48eaa865cd54450eb3R509-R518
- https://github.com/FlowiseAI/Flowise/pull/284/files#diff-473a2e7fb254080e3d1f8e339d5bbf13cb756874dd9aae2b3def6d49a8dfb41dR218

packages/server/src/ChildProcess.ts
```diff
@@ -83,6 +83,7 @@ export class ChildProcess {
                 depthQueue,
                 componentNodes,
                 incomingInput.question,
+                '',
```
HenryHengZJ, line 86: can we also pass in the `chatid` from here - https://github.com/FlowiseAI/Flowise/blob/main/packages/server/src/index.ts#L445 ?

packages/ui/src/views/chatmessage/ChatMessage.js
```diff
@@ -118,7 +118,8 @@ export const ChatMessage = ({ open, chatflowid, isDialog }) => {
 
         setLoading(true)
         setMessages((prevMessages) => [...prevMessages, { message: userInput, type: 'userMessage' }])
-        addChatMessage(userInput, 'userMessage')
+        // waiting for first chatmessage uploaded, the first chatmessage id will be chatId for every components
```
HenryHengZJ, line 121: the code `await addChatMessage(userInput, 'userMessage')` looks the same before and after changes, is the comment redundant?

packages/components/nodes/memory/ZepMemory/ZepMemory.ts
```diff
@@ -0,0 +1,130 @@
...
+                type: 'string',
+                default: 'http://127.0.0.1:8000'
+            },
+            {
+                label: 'Auto Summary',
+                name: 'autoSummary',
+                type: 'boolean',
+                default: true
+            },
+            {
```
HenryHengZJ, line 38: I think we can also still have the session id in case when user want to resume the previous conversation. if the session id field is not filled, then we can use the first `chatid`:

```
{
      label: 'Session Id',
      name: 'sessionId',
      type: 'string',
      optional: true,
      additionalParams: true
},
```

HenryHengZJ:
> @HenryHengZJ According to the current data structure, I think the first chat message id is a better choice for `chat_id`
> 
> For generality, I found the first chat message in backend, and pass it by `options` in `INode` definition `init?(nodeData: INodeData, input: string, options?: ICommonObject): Promise<any>`
> 
> This is the minimum modification in my mind, but I'm not sure if you have better solution for this.
> 
> image
> 
> Key Code:
> 
> * https://github.com/FlowiseAI/Flowise/pull/284/files#diff-8e6dfbbfb522575fe7c568c2de518bf9de351de83c9dbb48eaa865cd54450eb3R509-R518
> * https://github.com/FlowiseAI/Flowise/pull/284/files#diff-473a2e7fb254080e3d1f8e339d5bbf13cb756874dd9aae2b3def6d49a8dfb41dR218

i love the idea of using first chat id, that way when user clear the conversation, the session id will also be refreshed

packages/ui/src/views/chatmessage/ChatMessage.js
wangerzi: Hi, maybe I should say goodmorning to you! Considering that when the chat message is sent for the first time, if the `POST /chatmessage/{id}` request and the `POST /internal-prediction/${id}` request are executed concurrently, the first message id ORM query will fail in the prediction request.

packages/components/nodes/memory/ZepMemory/ZepMemory.ts
wangerzi: This is better than the current, I'll implement it later! 😀

packages/server/src/ChildProcess.ts
wangerzi: It's missing, I have forgot the `chatflow` variable contains a correct chatflow.id we can use, I'll fix it later!👀

packages/ui/src/views/chatmessage/ChatMessage.js
wangerzi: I tried to remove `await` and make some test, as my last comment, it will trigger error from api `/internal-prediction/{id}` sometimes, because chatmessage can't be queried in database
image

wangerzi:
@HenryHengZJ I'm already update the code based on our conversion, and I create a new function `getChatId` at `index.ts`, it's easy to use but I'm not sure if this is the best place.

packages/components/nodes/memory/ZepMemory/ZepMemory.ts
```diff
@@ -0,0 +1,140 @@
...
+                label: 'Session Id',
+                name: 'sessionId',
+                type: 'string',
+                placeholder: 'if empty, chatId will be used automatically',
```
HenryHengZJ, line 42: we can use `description` for the message

packages/server/src/ChildProcess.ts
```diff
@@ -76,13 +77,15 @@ export class ChildProcess {
             const { startingNodeIds, depthQueue } = getStartingNodes(nonDirectedGraph, endingNodeId)
 
             /*** BFS to traverse from Starting Nodes to Ending Node ***/
+            const chatId = await getChatId(chatflow.id)
```
HenryHengZJ, line 80: this will not work because child processes dont have access to `DataSource`. You'll have to get the chatid in `startChildProcess()` and pass it in the `IRunChatflowMessageValue` - https://github.com/FlowiseAI/Flowise/blob/main/packages/server/src/index.ts#L449. Then retrieve it here - https://github.com/FlowiseAI/Flowise/blob/main/packages/server/src/ChildProcess.ts#L26

packages/server/src/index.ts
```diff
@@ -661,6 +665,18 @@ export class App {
     }
 }
```
HenryHengZJ, line 667: can you add 
```
/**
 * Get first chat message id
 * @param {string} chatflowid
 * @returns {string}
*/
```

wangerzi:
@HenryHengZJ Thank you very much for your patience and guidance, I have changed my code by your comment.😀

em, but I just have a little confusion, I don't known how to test `ChildProcess.ts` in the UI

HenryHengZJ:
> @HenryHengZJ Thank you very much for your patience and guidance, I have changed my code by your comment.😀
> 
> em, but I just have a little confusion, I don't known how to test `ChildProcess.ts` in the UI

In the `.env` file in `packages/server`, you can specify the `EXECUTION_MODE`: https://github.com/FlowiseAI/Flowise/blob/main/packages/server/.env.example#L4

wangerzi:
> > @HenryHengZJ Thank you very much for your patience and guidance, I have changed my code by your comment.😀
> > em, but I just have a little confusion, I don't known how to test `ChildProcess.ts` in the UI
> 
> In the `.env` file in `packages/server`, you can specify the `EXECUTION_MODE`: https://github.com/FlowiseAI/Flowise/blob/main/packages/server/.env.example#L4

Thanks for answering, I've updated some code to remove a useless ORM query and tested the `ChildProcess.ts` by modify `.env`

HenryHengZJ:
thanks @wangerzi !! tested and it works like a charm!

really appreciate the ground work of session id using chat id, it makes it easy for other memory implementation thats using session id as well

PR #287, opened 2023-06-10
resolved long link horizontal scrolling problem

vkrm-segta:


packages/ui/src/views/chatmessage/ChatMessage.css
```diff
@@ -75,6 +75,10 @@
 }
 
 .markdownanswer a {
+    display: block;
+    margin-right: 2.5rem;
+    word-wrap: break-word;
+    overflow-wrap: break-word;
```
HenryHengZJ, line 81: dont think u need `overflow-wrap`.

somehow its still showing a little space there: 
image

vkrm-segta: Sure working on it

vkrm-segta: @HenryHengZJ 
While prompting if i get Request failed. I also need to move the horizontal scroll bar to see the full error.

https://github.com/FlowiseAI/Flowise/assets/124426279/ce4d5b4a-7e3f-430b-a643-dbaf7ce278c3

HenryHengZJ: thanks for finding that out too! yea you can disable horizontal scrolling

Issue #225, opened 2023-05-30
⚒️[FEATURE] Figma Loader Integration


HenryHengZJ:
## Describe the feature you'd like
Figma Loader Integration
Docs: https://js.langchain.com/docs/modules/indexes/document_loaders/examples/web_loaders/figma

## ✅Tasks
- [ ] Add Figma Loader node

## 🎯Acceptance Criteria
- [ ] Figma Loader node should be able to load data from Figma

## 👨‍💻How to contribute
Follow this [guide](https://github.com/FlowiseAI/Flowise/blob/main/CONTRIBUTING.md#step-by-step) for step by step on how to fork, commit, and submit PR.

## 👨‍💻Reference code
- [Github Loader](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/documentloaders/Github/Github.ts)

## More info
Like to work on this issue? Let us know, and we will assign it to you 🚀
If you need any help or just want to hangout, reach out to us on our [Discord server](https://discord.gg/jbaHfsRVBW)👾

Pravesh-mansharamani:
@HenryHengZJ Can you assign me this issue

PR #313, opened 2023-06-13
Figma loader integration #225

Pravesh-mansharamani:
Added the Figma Loader. 
Screenshot 2023-06-13 at 8 50 20 PM


packages/components/nodes/textsplitters/LatexTextSplitter/LatexTextSplitter.ts
```diff
@@ -0,0 +1,54 @@
+import { INode, INodeData, INodeParams } from '../../../src/Interface'
```
HenryHengZJ, line 1: LatexTextSplitter shouldnt be part of this PR, if you can remove these 2 files @Pravesh-mansharamani

Issue #311, opened 2023-06-13
⚒️[FEATURE] Puppeteer Integration


HenryHengZJ:
## Describe the feature you'd like
Puppeteer Integration
Docs: https://js.langchain.com/docs/modules/indexes/document_loaders/examples/web_loaders/web_puppeteer

## ✅Tasks
- [ ] Add Puppeteer Loader node

## 🎯Acceptance Criteria
- [ ] Puppeteer Loader node should be able to scrape data from webpages

## 👨‍💻How to contribute
Follow this [guide](https://github.com/FlowiseAI/Flowise/blob/main/CONTRIBUTING.md#step-by-step) for step by step on how to fork, commit, and submit PR.

## 👨‍💻Reference code
- [Cheerio Loader](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/documentloaders/Cheerio/Cheerio.ts)

## More info
Like to work on this issue? Let us know, and we will assign it to you 🚀
If you need any help or just want to hangout, reach out to us on our [Discord server](https://discord.gg/jbaHfsRVBW)👾

AidanTilgner:
Hey, I'm new here but I'd be interested in taking this on. I've worked with puppeteer recently so it's fresh.

PR #314, opened 2023-06-13
feat: added puppeteer loader node

AidanTilgner:
I updated the `components` package to add the puppeteer integration described in [this issue](https://github.com/FlowiseAI/Flowise/issues/311). This is my first PR here, so let me know if I need to change anything/everything.

Screenshot 2023-06-13 at 1 18 00 PM

- [x] Add Puppeteer Loader node
- [x] Puppeteer Loader node should be able to scrape data from webpages
- [x] Build successful 

> **Note:** Puppeteer is updating their default `headless` mode, and therefore a warning is displayed and the option to update the `headless` proeprty to `headless: "new"` is offered. However, it seems that the decision to update to `headless: "new"` is controlled by the `PuppeteerWebBaseLoader` langchain loader. I don't think this is an issue, as the new headless mode will become the default soon, but I figured I'd mention it.


AidanTilgner:
It seems that there are some merge conflicts now since the new emergency changes were made. I will update this tomorrow.

AidanTilgner:
Screenshot 2023-06-14 at 10 52 21 AM
ert

I ran the build and start command according to the [contributing guidelines](https://github.com/FlowiseAI/Flowise/blob/main/CONTRIBUTING.md). I also created the above chain for testing purposes. I confirmed that the "Puppeteer Web Scraper" node was scraping documents successfully. Then I used the "Pinecone Upsert Document" node to confirm that they were being scraped and then pulled into memory by the language model. Everything works the same as the [cherio reference code](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/documentloaders/Cheerio/Cheerio.ts).

packages/components/package.json
```diff
@@ -38,9 +38,9 @@
         "moment": "^2.29.3",
         "node-fetch": "^2.6.11",
         "pdf-parse": "^1.1.1",
+        "puppeteer": "^20.7.1",
         "weaviate-ts-client": "^1.1.0",
-        "ws": "^8.9.0",
-        "html-to-text": "^9.0.5"
```
HenryHengZJ, line 43: accidentally removed `html-to-text`?

AidanTilgner: Yeah I think I accidentally deleted it when resolving the merge conflict

AidanTilgner: Added a fix

HenryHengZJ (approved):
thanks @AidanTilgner! lgtm

Issue #317, opened 2023-06-14
How to hide the node modules while keeping the chat


dnagpal1:
### Discussed in https://github.com/FlowiseAI/Flowise/discussions/164

<div type='discussions-op-text'>

<sup>Originally posted by **Wany917** May 24, 2023</sup>
Hi ! I created a chatbot and would like to have only the chatbot displayed, not the nodes that compose it, so that people can't modify it. Also, I don't know how to code</div>

HenryHengZJ:
As mentioned from the PR, we will implement a standalone chat application that connect to your flow at a later stage

HenryHengZJ:
closing as https://github.com/FlowiseAI/Flowise/pull/416 merged

PR #318, opened 2023-06-14
Feature/hide show dialog

dnagpal1:
**Why?**

Feature Request: #317

**Problem:** Currently users are now able to hide node modules while using the chatbot. This feature allows them to hide the UI while keeping functionality of the chatbot.

**Solution:** Added a toggle button next to node modules creation button to hide/show all node modules. According to React Flow documentation this function loops through all existing flows and hides them. 

**Test Plan:** Manually checked hiding pre-existing marketplace node modules and adhoc created node modules.


HenryHengZJ:
hey @dnagpal1 thanks for the PR! In future we will implement a standalone chat application that connect to your flow. Therefore, I dont think hide/unhide nodes in flow fits the product UI/UX.

Unfortunately I will have to decline the PR for now, thanks again for the effort!

PR #322, opened 2023-06-14
Bugfix/pdf loader add legacy option

HenryHengZJ:


heresandyboy:
Hi Henry, could you also add the docker dependencies to the `./docker/Dockerfile` too, the one that just npm installs flowise? With an update to the dockerhub image. Cheers.

HenryHengZJ:
hey @heresandyboy also updated the file in `docker/Dockerfile`, do you have additional things you want to add?

PR #358, opened 2023-06-17
Feature/logging winston

matthias:
Added winston logger (instead of condole.log) to the server package
- this is for review
- intention is to also add this to components
- as the goal is to have better logging for chat prompts / chains

Features:
- console logging
- middleware logger for all requests as a file logger
- util/config.ts is up to discussion / work in progress => any input how to best configure the app (.env? /  config.json / congig.ts / ...)


HenryHengZJ:
love it! will review asap!

matthias:
Thanks for your reply. Seems I left a ; in the code.
Just let me know if you want anything changed.

packages/server/src/utils/config.ts
```diff
@@ -0,0 +1,19 @@
+import * as path from 'path'
```
HenryHengZJ, line 1: as this `config` file is specific to logger, can we move it to `logger.ts`? or is there any specific reason you put it seperately?

packages/server/src/utils/config.ts
```diff
@@ -0,0 +1,19 @@
...
+import * as fs from 'fs'
+
+// should be config via .env or config file
+const logDir = path.join(__dirname, '../../../..', 'logs')
```
HenryHengZJ, line 5: we can also let user specify their own preferred path to store logs.
For example:
```
const logDir = process.env.LOG_PATH ?? path.join(__dirname, '../../../..', 'logs')
```

packages/server/src/utils/config.ts
```diff
@@ -0,0 +1,19 @@
...
+
+// Create the log directory if it doesn't exist
+if (!fs.existsSync(logDir)) {
+    fs.mkdirSync(logDir)
+}
+
+console.info('Defined logDir:', logDir, ' in ' + __filename)
+
+export const config = {
+    logLevel: 'info', // Set your desired log level
```
HenryHengZJ, line 15: same as well
```
logLevel: process.env.LOG_LEVEL ?? 'info'
```

packages/server/src/utils/config.ts
matthias: Yes. The reason for this incompleteness was that I wasn't sure where the project wants to store it's configuration. Could be on package level. Could be on app level. Could be via .env / ENV. As I'm not that experienced with Node projects I didn't know what's the best practice (especially with multiple packages in the same repo)

But if I may suggest something - I would put the configuration in the root directory of the project. Like babel.config.js -> flowise.config.js => but for now it would exceed this PR - so an intermediate solution I created the config.ts within utils -> could also move it one level up to the package root / or leave it out for now. And just mark it for later.

HenryHengZJ: since the logger only logs the server, I would leave it within `server` folder. As for where should we put `config.ts`, I don't have a strong preference, we can leave it within `utils` folder for now.

as we make repo more configurable, like database/credentials configuration, we can further improve this file

HenryHengZJ:
hey @matthias are you still working on this? would love to get this merge asap and add additional logging at other places

matthias:
Yes. Sorry for the delay. Hope to be done by today - latest on Monday. Almost done - need to do some testing.

I go now with a strategy that could later on be converted to node-config - which seems to be a good option.

matthias:
Changes:
- util/config.ts holds some basic configuration (defaults)
- later on we could migrate / enhance to node-config
 - Added LOG_DIR to .env.example to specify the logging dir (relative to project root)
- I reduced the number of files to:
 - server-requests.log.jsonl (express middleware logger)
 - server.log
 - server-error.log
- changed level for GET requests from 'verbose' to 'http'

Regarding configuration:
- I added a skeleton / default configuration object
- For now only the logging dir can be configured via .env (within server)
- I would suggest to introduce node-config for further configuration

Let me know if you have any suggestion.
I'm also not sure if I need to do anything on github (re-submit the PR ?)

HenryHengZJ:
thanks @matthias for the changes, I can see the PR included the other PR's changes that were merged. You can try pull the latest changes from develop, or if you find it confusing, you can close this PR and open another one with just the logging winston changes

Issue #369, opened 2023-06-19
⚒️[FEATURE] Redis-Backed Chat Memory


HenryHengZJ:
## Describe the feature you'd like
Redis-Backed Chat Memory
Docs: https://js.langchain.com/docs/modules/memory/examples/redis

## ✅Tasks
- [ ] Add Redis Memory node

## 🎯Acceptance Criteria
- [ ] Redis Memory node should be able to remember conversations

## 👨‍💻How to contribute
Follow this [guide](https://github.com/FlowiseAI/Flowise/blob/main/CONTRIBUTING.md#step-by-step) for step by step on how to fork, commit, and submit PR.

## 👨‍💻Reference code
- [Zep Memory](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/memory/ZepMemory/ZepMemory.ts)

## More info
Like to work on this issue? Let us know, and we will assign it to you 🚀
If you need any help or just want to hangout, reach out to us on our [Discord server](https://discord.gg/jbaHfsRVBW)👾

ivalkshfoeif:
Hi @HenryHengZJ, I'd like to work on the issue.

PR #407, opened 2023-06-24
Feature/Redis-Backed Chat Memory

ivalkshfoeif:
## issue

https://github.com/FlowiseAI/Flowise/issues/369

## note

The [Redis-Backed Chat Memory](https://js.langchain.com/docs/modules/memory/examples/redis) is closer to [Buffer Memory](https://js.langchain.com/docs/modules/memory/examples/buffer_memory) than [Zep Memory](https://js.langchain.com/docs/modules/memory/examples/zep_memory) according to the Langchain Docs

I use the same icon with other memory to keep consistent. I'm happy to update it or other memory icons if you think it's not necessary.

Add Redis dependency to flowise-components workspace since it's required in RedisChatMessageHistory

Screen Shot 2023-06-23 at 5 51 29 PM

Screen Shot 2023-06-23 at 5 51 16 PM


packages/components/nodes/memory/RedisBackedChatMemory/RedisBackedChatMemory.ts
```diff
@@ -0,0 +1,85 @@
...
+
+        if (sessionTTL) {
+            obj = {
+                ...obj,
+                sessionTTL
+            }
+        }
+
+        let redisChatMessageHistory = new RedisChatMessageHistory(obj)
+        let redis = new BufferMemory({ memoryKey, chatHistory: redisChatMessageHistory, returnMessages: true })
```
HenryHengZJ, line 79: I tested it, seems like we don't need `memoryKey`, is there any reason why include here or can we remove it?

HenryHengZJ:
think we can replace the icon the Redis icon, purely personal preference 😅 so its not so dull haha

HenryHengZJ:
also putting a note here: you can run redis instance with docker
```
docker run --name my-redis -p 6379:6379 -d redis
```

packages/components/nodes/memory/RedisBackedChatMemory/RedisBackedChatMemory.ts
ivalkshfoeif: Hi Henry, if I delete memory key, it will throw the error: 
Screen Shot 2023-06-27 at 11 35 26 PM

As a result, I use the same implementation as [BufferMemory](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/memory/BufferMemory/BufferMemory.ts)

HenryHengZJ: Oh okay strange I didn't have this error. Cool think we can leave it as it is

HenryHengZJ (approved):
looks good thanks @ivalkshfoeif !

PR #442, opened 2023-06-29
fix: temperature should convert to float

wangerzi:
image

When init `ChatOpenAI`, `temperature` attribution should be float, otherwise we will always set temperature = 0

Before:
image

After fix:
image


HenryHengZJ:
ohhhhh wait whattttttttt all these time i didnt notice that!! would you be able to update all the nodes that have the temperature settings? 

Thank you so much!! @wangerzi

wangerzi:
@HenryHengZJ I updated all the code by search `temperature` variable 😄

PR #450, opened 2023-06-30
modify whitelistURLs

chungyau97:
to bypass header configuration for Zapier


packages/server/src/index.ts
```diff
@@ -93,7 +93,13 @@ export class App {
             const basicAuthMiddleware = basicAuth({
                 users: { [username]: password }
             })
-            const whitelistURLs = ['/api/v1/public-chatflows', '/api/v1/prediction/', '/api/v1/node-icon/', '/api/v1/chatflows-streaming']
+            const whitelistURLs = [
```
HenryHengZJ, line 96: missing `'/api/v1/public-chatflows'`

PR #468, opened 2023-07-03
feat: prevent submission during IME composition

toshilow:
Fixed to prevent submission during converting using IME like Japanese
In Safari, `isComposing` is not as expected, so use `keyCode` to determine

https://developer.mozilla.org/en-US/docs/Web/API/Element/keydown_event

If this fix is acceptable,I will also create a PR for FlowiseAI/FlowiseChatEmbed.


HenryHengZJ (approved):
looks good @toshilow thanks!

PR #469, opened 2023-07-03
Added winston logging

matthias:
- use logger.xxx instead of console.xxx
- added express middleware logging (using jsonl)
- added LOG_PATH as environment variable
- more configs postponed for later iteration


matthias:
Created a new clean merge

matthias (commented):
I understand that you also want to respect the DEBUG setting.

But not sure if the DEBUG flag is the right one to enable / disable request logging at all. Even in a production environment you want HTTP request logging.

I would suggest to move this into the logger.ts / expressRequestLogger -> so it can be configured via config.

HenryHengZJ:
good point @matthias do you reckon we should add another env var like `ENABLE_LOGGING`? This allow user to have flexibiltity to turn on/off loggers if they dont want to burden the system with logs

HenryHengZJ:
You can create another branch base from this branch.

Are you trying to log the output of each component? I was trying to see if we can add the logs output from langchain when the DEBUG is true, as currently they are just printed on the terminal/console

HenryHengZJ:
@matthias I reverted the DEBUG changes and allow logger by default. Let me know if you are still working on this, if not we can merge this and open another PR for the logger on components

HenryHengZJ (approved):

matthias:
Yes you can merge it ... logger on compnents / flows should be another PR imo.

PR #496, opened 2023-07-07
Feature/Nodes Search Quality of Life Updates

AidanTilgner:
If you search for a category, then all of the nodes that match that category will show up. This doesn't break existing functionality, as searching for a specific node still reveals that node as expected. I also added a clear button, which will clear and reset the node filter, that way you don't have to manually clear it.
Screenshot 2023-07-06 at 4 59 31 PM
Screenshot 2023-07-06 at 4 59 47 PM
om/FlowiseAI/Flowise/assets/45406132/ad789c0e-caea-48de-a1ac-e03b5edfffcb">

Here are some examples. This is just an update that I personally wanted, I'm happy to make changes or add other QoL features to search that might be nice to have. As the number of nodes grows, search functionality will need to be able to scale.

### DIFF
--- a/packages/ui/src/views/canvas/AddNodes.js
+++ b/packages/ui/src/views/canvas/AddNodes.js
@@ -34,7 +34,7 @@ import Transitions from 'ui-component/extended/Transitions'
 import { StyledFab } from 'ui-component/button/StyledFab'
 
 // icons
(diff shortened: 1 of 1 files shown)


HenryHengZJ (approved):
tested it and looks really good! thanks @AidanTilgner for the improvement!

PR #504, opened 2023-07-07
fix: docker should install chromium and puppeteer should be no sandbox

AidanTilgner:
A couple of considerations here:

- 1) `--no-sandbox` means that puppeteer and chromium won't execute in a sandbox. This could theoretically have security implications, but since this is in a docker image I doubt it will matter. Maybe something to discuss further.
- 2) I'm not entirely sure the context around `ENV PUPPETEER_SKIP_DOWNLOAD=true` in the Dockerfile, although based on [this pr](https://github.com/FlowiseAI/Flowise/pull/350) it seems that the default install script was causing an error. I'm not seeing this error during build time here, and my thinking is through the Dockerfile there's more control anyway.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,7 +8,12 @@ FROM node:18-alpine
 RUN apk add --update libc6-compat python3 make g++
 # needed for pdfjs-dist
 RUN apk add --no-cache build-base cairo-dev pango-dev
+
+# Install Chromium
+RUN apk add --no-cache chromium
+
 ENV PUPPETEER_SKIP_DOWNLOAD=true
+ENV PUPPETEER_EXECUTABLE_PATH=/usr/bin/chromium-browser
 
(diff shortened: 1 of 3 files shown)


AidanTilgner:
Also for consideration, the `headless: "new"` thing, I added to silence the current warning about the default headless mode switching in the future.

chungyau97:
Results:
image
image
image

chungyau97 (approved):

HenryHengZJ:
@chungyau97 have you tried root Dockerfile, not the one inside `docker` folder

chungyau97:
> @chungyau97 have you tried root Dockerfile, not the one inside `docker` folder

I copy the changes and use it in the root folder 
image

chungyau97:
Results:
**Docker Folder**
Before
image
After

PR #510, opened 2023-07-09
modify options and password api config

chungyau97:
Results:
**Options**
image
image

**Password**
image


packages/server/src/utils/index.ts
```diff
@@ -632,7 +632,7 @@ export const findAvailableConfigs = (reactFlowNodes: IReactFlowNode[]) => {
     for (const flowNode of reactFlowNodes) {
         for (const inputParam of flowNode.data.inputParams) {
             let obj: IOverrideConfig
-            if (inputParam.type === 'password' || inputParam.type === 'options') {
+            if (inputParam.type === 'password') {
```
HenryHengZJ, line 635: i think we can also remove this too

HenryHengZJ: since users are also asking how to put in api key from API

PR #518, opened 2023-07-11
Feature/Log Component Level

HenryHengZJ:
Add `LOG_LEVEL` and more loggers on verbose and debug.

This new changes added logger on every chain/agents, and logger error with stack trace
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,28 @@ FLOWISE_USERNAME=user
 FLOWISE_PASSWORD=1234
 ```
 
+## 🌱 Env Variables
+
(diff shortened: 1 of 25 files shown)


HenryHengZJ:
@matthias here is the PR for adding loggers on components

matthias:
Looks cool. 

I haven't figured out where you configure the logger. From where is options.logger coming? Just can't figure this out.

const loggerHandler = new ConsoleCallbackHandler(options.logger)

And: I did locally install the branch - but couldn't figure out how to enable the chain logging ... (.env ?)

HenryHengZJ:
in `packages/server/src/utils/index.ts` line 233, for chain logging i set it as verbose, so you need to add `LOG_LEVEL=debug` in your .env file

PR #531, opened 2023-07-12
Enabled chain and prediction logging

matthias:
handler.ts
- pass metadata to logger instead of directly writing it to message
- TODO: review of metadate fields

index.ts
- Changed log level for prediction chain logs to 'verbose' (same as in chain)
- Added verbose prediction log message with: question, answer, chatId, chatflowId, ...

logger.ts
- removed config.ts for now and use LOG_* env variables directly
- write metadata to message where appropriate
- _underscore metadata fields are only printed to JSON and if LOG_LEVEL=DEBUG
- removed (unused) defaultMeta
- Colorize Console logs
- Removed timestamp from Console logs
- Print metadata as pretty JSON to Console
- Added server.log.jsonl
- Added prediction.log and prediction.log.jsonl

TODO: we should also remove _underscore values from .jsonl logs


matthias:
Hi @HenryHengZJ - your chain logging is awesome! 
I integrated it better with the logger - so it just is still pretty printed in the console - but we can better control how metadata is printed in files or jsonl.

Todos: (needs review)
- the metadata fields I pass in handler.ts need some review - i couldn't figure out which fields are really available
- I introduced a convention, that _underscore metadata fields are not printed unless LOG_LEVEL=DEUBG -> but it's still printed to .jsonl => this can become VERY large

Next Steps:
- the goal behind the efforts is (as mentioned) to provide a question / answer logging => after all the work (also from your side) I found out that all the info is in index.ts (btw. the file is becoming very large)
- so I added a prediction.log and we are working on a UI for users in the backend - we would need to store the prediction log also in the DB - and then render it in the UI

Hope this PR makes it to main ! Thanks for your great work !

HenryHengZJ:
@matthias I think its good to have more details log, but these logs can grow very huge if we didnt have a job to clear/remove these logs.

From your changes, the logs have become harder to read: 
image

as compared to previous:
image

Also wondering is there a need for `jsonl` file? It seems to just a duplicate of the `.log` files

In term of terminal/console logging, now we we have `verbose` logs on top of the original VERBOSE logs from langchain: 
image
with much more details, but its too much info that I can't easily read whats happening

I find the original VERBOSE log is clear enough for me to see whats going on: 
image

PR #545, opened 2023-07-14
ChatMessage Order Fixed

atishamte:
Currently the chat message order is coming random because of order of UUID.

Added order by as per the createdDate to ASC to get the chat message in order


govind-kumarr (approved):

chungyau97:
Results:
image

PR #547, opened 2023-07-15
Feature/Credential

HenryHengZJ:
Add credentials to be able to store encrypted sensitive info like API keys, and not being exposed to client side

image

Then use in node:

image


goldzulu:
This is awesome stuff and a GODSEND @HenryHengZJ ! I was going through the code and appreciating your handywork!

Architecture wise though, I was wondering though whether there are plans to steer away from file based storage as it will require persistent disk setup. With the other feature waiting merging (non sqlite file based database) #575, it would be good if api keys and credentials do not have to rely on file based storage solely and have options to use alternative storage mechanism.

HenryHengZJ:
what are the other storage mechanism you recommend @goldzulu ?

goldzulu:
@HenryHengZJ there are cloud based like AWS Key Management Service (AWS KMS), Google Cloud Key Management Service (Google Cloud KMS), Azure Key Vault: Microsoft Azure Key Vault, HashiCorp Vault: HashiCorp Vault. There's also https://supabase.com/blog/supabase-vault. Potentially you can also use any database table (isolated) from the one storing chat flows etc. and store the content encrypted (this is probably the fastest option for now)

HenryHengZJ:
@goldzulu yeah with @atishamte changes, you can use postgres and connect to supabase, then encrypt your table. As most folks that use Flowise don't have the technical expertise to do so, thats why settling for in-house file based storage sqlite for ease to use

goldzulu:
@HenryHengZJ Yes sqlite is right choice for quick start. The @atishamte change if I am not wrong only for the sqlite replacement. However there is no current way to store the apikeys and credentials that you have just introduced using the same non sqlite database. Do you think it is a big change to use database (encrypted) for apikeys and credentials?

HenryHengZJ:
it might not be straight-forward. Because chatflow is currently using the id from credential table to retrieve the apikeys, isolating credential table might need additional logic

PR #552, opened 2023-07-16
New Html To Markdown TextSplitter

raffareis:
## Motivation

When using webcrawlers as document loaders to embed and store in a Vector Store, much of what is embedded are useless scripts and styles. This creates a semantic mess for the embeddings.

## Solution

I've introduced a new `HtmlToMarkDownTextSplitter` that inherits from `MarkdownTextSplitter`. This new splitter uses the [`node-html-markdown`](https://www.npmjs.com/package/node-html-markdown) package to convert HTML to Markdown before spliting. Markdown is much more semantically significant for vector embedding.

This approach helps to clean up the embeddings and make them leaner and more meaningful as markdown.

Here's the main code change:

```typescript
splitText(text: string): Promise<string[]> {
    return new Promise((resolve) => {
        const markdown = NodeHtmlMarkdown.translate(text)
        super.splitText(markdown).then((result) => {
            resolve(result)
        })
    })
}
```

## Results

image


HenryHengZJ:
@raffareis thank you so much for this! Indeed the embeddings are much more cleaner!
Here's comparison:

Using recursive text splitter:
image

Using Html Markdown splitter:
image

PR #554, opened 2023-07-16
add -H "Content-Type: application/json" in curl command

Yongtae723:
this is copy from [discord](https://discord.com/channels/1087698854775881778/1096265567058534461/1130069743760588883)

after I deploy flowise on render.com and make some LLM chain, we can not call the LLM via curl and javascript, but python can connect.
after some workaround, I found we have to explicitly specify the encoding of data in JSON, for example adding -H "Content-Type: application/json"in curl command.
python do this specification and thus no error happened.

this is curl version

### DIFF
--- a/packages/ui/src/views/chatflows/APICodeDialog.js
+++ b/packages/ui/src/views/chatflows/APICodeDialog.js
@@ -319,7 +319,7 @@ query(formData).then((response) => {
         } else if (codeLang === 'cURL') {
             return `curl ${baseURL}/api/v1/prediction/${dialogProps.chatflowid} \\
      -X POST \\${getConfigExamplesForCurl(configData, 'formData')} \\
-     -H "Content-Type: application/json"`
+     -H "Content-Type: multipart/form-data"`
         }
         return ''
     }
@@ -366,7 +366,7 @@ query(formData).then((response) => {
         } else if (codeLang === 'cURL') {
             return `curl ${baseURL}/api/v1/prediction/${dialogProps.chatflowid} \\
      -X POST \\${getConfigExamplesForCurl(configData, 'formData')} \\
-     -H "Content-Type: application/json" \\
+     -H "Content-Type: multipart/form-data" \\
      -H "Authorization: Bearer ${selectedApiKey?.apiKey}"`
         }
         return ''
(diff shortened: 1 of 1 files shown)


packages/ui/src/views/chatflows/APICodeDialog.js
```diff
@@ -316,7 +318,8 @@ query(formData).then((response) => {
 `
         } else if (codeLang === 'cURL') {
             return `curl ${baseURL}/api/v1/prediction/${dialogProps.chatflowid} \\
-     -X POST \\${getConfigExamplesForCurl(configData, 'formData')}`
+     -X POST \\${getConfigExamplesForCurl(configData, 'formData')} \\
+     -H "Content-Type: application/json"`
```
HenryHengZJ, line 322: should be `application/x-www-form-urlencoded` when using `formData`

Yongtae723: thank you for pointing out this point!

I searched a bit and found we should use ` multipart/form-data` instead of `application/json` or `application/x-www-form-urlencoded`.
https://stackoverflow.com/questions/4007969/application-x-www-form-urlencoded-or-multipart-form-data

HenryHengZJ: have you tested it? is it not working if we dont specify anything there

Yongtae723: Yes I did!

what I did is 
- using `-H "Content-Type: application/json"` → failure
- using `-H "Content-Type: application/x-www-form-urlencoded"` → failure
- using `-H "Content-Type: multipart/form-data"` → succeed
- without Headers → succeed

current version of cURL use `multipart/form-data` as default when we use `-F` . But the reason I put `-H "Content-Type: multipart/form-data"` is defalut value can be changed by version update of cURL.

But if you think it too much, I will delete `-H "Content-Type: multipart/form-data"`

HenryHengZJ: ah okay thanks for testing different methods! its fine we can use `-H "Content-Type: multipart/form-data"` :)

HenryHengZJ (approved):
thank you @Yongtae723 !

PR #555, opened 2023-07-16
Refactor API request to include Content-Type header

Yongtae723:
this is copy from [discord](https://discord.com/channels/1087698854775881778/1096265567058534461/1130069743760588883)

after I deploy flowise on render.com and make some LLM chain, we can not call the LLM via curl and javascript, but python can connect.
after some workaround, I found we have to explicitly specify the encoding of data in JSON, for example adding -H "Content-Type: application/json"in curl command.
python do this specification and thus no error happened.

this is jacascript version

### DIFF
--- a/packages/ui/src/views/chatflows/APICodeDialog.js
+++ b/packages/ui/src/views/chatflows/APICodeDialog.js
@@ -310,9 +310,9 @@ async function query(formData) {
         {
             method: "POST",
             headers: {
-                "Content-Type": "application/json"
+                "Content-Type": "application/x-www-form-urlencoded"
             },
-            body: JSON.stringify(formData)
+            body: new URLSearchParams(formData)
         }
     );
     const result = await response.json();
@@ -358,10 +358,10 @@ async function query(formData) {
         {
             headers: {
                 Authorization: "Bearer ${selectedApiKey?.apiKey}",
-                "Content-Type": "application/json"
+                "Content-Type": "application/x-www-form-urlencoded"
             },
             method: "POST",
-            body: JSON.stringify(formData)
+            body: new URLSearchParams(formData)
         }
     );
     const result = await response.json();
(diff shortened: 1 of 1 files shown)


packages/ui/src/views/chatflows/APICodeDialog.js
```diff
@@ -303,7 +309,10 @@ async function query(formData) {
         "${baseURL}/api/v1/prediction/${dialogProps.chatflowid}",
         {
             method: "POST",
-            body: formData
+            headers: {
+                "Content-Type": "application/json"
+            },
+            body: JSON.stringify(formData)
```
HenryHengZJ, line 315: Dont think this works. Using `formData` , the header should be `'Content-Type': 'application/x-www-form-urlencoded'` not `application/json`, so I prefer to leave this out.

You also cant `JSON stringify` formData

Yongtae723: oh sorry. you are correct.

thank you for pointing out this point!

packages/ui/src/views/chatflows/APICodeDialog.js
```diff
@@ -347,9 +356,12 @@ async function query(formData) {
...
         "${baseURL}/api/v1/prediction/${dialogProps.chatflowid}",
         {
-            headers: { Authorization: "Bearer ${selectedApiKey?.apiKey}" },
+            headers: {
+                Authorization: "Bearer ${selectedApiKey?.apiKey}",
+                "Content-Type": "application/x-www-form-urlencoded"
+            },
             method: "POST",
-            body: formData
+            body: new URLSearchParams(formData)
```
HenryHengZJ, line 364: do we need `new URLSearchParams` ? 

actually i dont think we should set `"Content-Type": "application/x-www-form-urlencoded"` according to many [answers](https://stackoverflow.com/questions/62644988/using-formdata-inside-fetch-api-is-not-working) online, just let the fetch decide for you

Yongtae723: I see.

then I'll revert to the changes related to the form.

HenryHengZJ (approved):
thank you @Yongtae723 !

PR #567, opened 2023-07-18
fix: bug inputType

comey246:
this the second pull request, the pull request just change the value type. this is my the first pull request( •̀ ω •́ )✧，thank you.


HenryHengZJ (approved):
thanks @comey246 !

Issue #573, opened 2023-07-18
[FEATURE] External flowise database connectivity


atishamte:
**Describe the feature you'd like**
Rather than having flowise sqlite database, while installing, I should be able to define database types like SQLite, MySQL, and Postgresql. So the basic 3 tables can be saved in these databases.

HenryHengZJ:
closing as #575 merged

PR #575, opened 2023-07-18
Feature/flowise multi database support

atishamte:
Flowise support for multiple database sources like sqlite, MySQL or postgres

Also, I added a new environment variable for overriding the database. If set to true, then it will override old db else it will keep as it
refer to #573


HenryHengZJ (approved):
thanks @atishamte for the PR! Looks good 👍

raffareis:
This has a little issue.,In the .md files it states the environment variable: `DATABASE_USERNAME`
image


However on the Datasource.ts code, it expects `DATABASE_USER`
image

@HenryHengZJ 
@atishamte

PR #589, opened 2023-07-21
Fix: server watch

comey246:
add the components change to the server nodemon watch, that if we change the components files, the components will tsc and the server will restart to load the new change components.


HenryHengZJ (approved):
thanks @comey246 !

PR #600, opened 2023-07-22
Feature/AirtableAgent

HenryHengZJ:
image


WizKid1968:
I'm having an issue with getting the airtable agent to identify records in the table. I have a table that has over 800 records, but it only recognizes 32 records. What can be the issue?

PR #612, opened 2023-07-24
Bugfix/HandleEscapeChar

HenryHengZJ:
Handling special chars in input when used with Prompt Template and special chars from Output Prediction:
JSON1

JSON2

json3


chungyau97:
Results:
image
image

chungyau97 (approved):

PR #613, opened 2023-07-25
Add Hyde retriever

Yongtae723:
this is related to [post of discord](https://discord.com/channels/1087698854775881778/1087698855342125118/1130074042540503040)

HyDE retriever is powerful for finding target documents, especially if a question is short.

This is my first js/ts contribution. so please review it carefully.

and I have 2 questions.

1. how can I check HydeRetriever works correctly?
I understand HydeRetriever generates a virtual response and embeds that response.
thus If I can see the virtual response by llm, I think we can check HydeRetriever works correctly.
Somehow it doesn't show up even though DEBUG = true.

2. how can I use a custom prompt.
I deep dive into the implementation of HyDE and found we can set promptKey:str or Custom prompt ([link](https://github.com/hwchase17/langchainjs/blob/504397f4f73529490c3f509bcb5679907750db02/langchain/src/retrievers/hyde.ts#L40))
I would like to know how should I implement into flowise.
Ideally, I would like to make the HyDE component can connect to PromptTemplate node (Optional) and
- If HyDE is connected with PromptTemplate node is prioritized.
or
- If HyDE is connected with PromptTemplate node and promptKey is empty or something like 'use PromptTemplate', PromptTemplate node is used.

I am looking forward review and reply from you!


packages/components/src/Interface.ts
```diff
@@ -166,3 +168,11 @@ export class VectorStoreRetriever {
         this.vectorStore = fields.vectorStore
     }
 }
+
```
HenryHengZJ, line 171: you can have this interface in `HydeRetriever.ts` file

packages/components/nodes/retrievers/HydeRetriever/HydeRetriever.ts
```diff
@@ -0,0 +1,121 @@
...
+                    },
+                    {
+                        label: 'mr-tydi',
+                        name: 'mr-tydi'
+                    }
+                ],
+                default: 'websearch'
+            },
+            {
+                label: 'Custom Prompt',
```
HenryHengZJ, line 77: @Yongtae723 here I added custom prompt:
image

If custom prompt is set, promptKey will be ignored

HenryHengZJ:
We will have to update langchain version in order to see the debug from retriever. I'll do that on other branch. Once updated, I can see the log:
image

HenryHengZJ (approved):
thanks a lot for getting a head start of this @Yongtae723 !

HenryHengZJ:
The final result: 
image

packages/components/nodes/retrievers/HydeRetriever/HydeRetriever.ts
Yongtae723: I confirm your code and this is what I intended!

thanks a lot!

Yongtae723:
> We will have to update langchain version in order to see the debug from retriever. I'll do that on other branch. Once updated, I can see the log:

Alright!
if updated, can I see log without any setting except DEBUG = true?

Yongtae723:
>thanks a lot for getting a head start of this @Yongtae723 !

Sure!
I learned a lot during this implementation!
So finally I can code it by myself when there is something new that has not in Flowise!

Also, I am really honored to contribute such a good project!

PR #617, opened 2023-07-25
bugfix/OpenAITopP

chungyau97:
**Before**: 
image
**After**:
image

**Caution**:
We generally recommend altering `temperature` or `top_p` but not both
**Reference**:
[OpenAI Temperature and TopP](https://platform.openai.com/docs/api-reference/completions/create#completions/create-temperature)

### DIFF
--- a/packages/components/nodes/vectorstores/Chroma_Existing/Chroma_Existing.ts
+++ b/packages/components/nodes/vectorstores/Chroma_Existing/Chroma_Existing.ts
@@ -69,7 +69,7 @@ class Chroma_Existing_VectorStores implements INode {
         const chromaURL = nodeData.inputs?.chromaURL as string
         const output = nodeData.outputs?.output as string
         const topK = nodeData.inputs?.topK as string
-        const k = topK ? parseInt(topK, 10) : 4
+        const k = topK ? parseFloat(topK) : 4
 
         const obj: {
             collectionName: string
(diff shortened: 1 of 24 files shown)


chungyau97:
> **Before**: image **After**: image
> 
> **Caution**: We generally recommend altering temperature or top_p but not both **Reference**: [OpenAI Temperature and TopP](https://platform.openai.com/docs/api-reference/completions/create#completions/create-temperature)

packages/components/nodes/vectorstores/Chroma_Existing/Chroma_Existing.ts
```diff
@@ -69,7 +69,7 @@ class Chroma_Existing_VectorStores implements INode {
         const chromaURL = nodeData.inputs?.chromaURL as string
         const output = nodeData.outputs?.output as string
         const topK = nodeData.inputs?.topK as string
-        const k = topK ? parseInt(topK, 10) : 4
+        const k = topK ? parseFloat(topK) : 4
```
HenryHengZJ, line 72: these are `parseInt`. they are different from the `topK` used in models, although they have same name.

check with langchainjs code and API docs from models to see which one is decimal values and which one is integer

HenryHengZJ: oh nvm my bad, `parseFloat` also works

HenryHengZJ (approved):

PR #627, opened 2023-07-27
Translate the Readme file into Simplified Chinese

xiexin12138:
My native language is Simplified Chinese, so the i18n process is divided into stages. Currently, the Readme file is being translated into Chinese, and later on, i18next and react-i18next will be used for multilingual support.


HenryHengZJ (approved):
thank you @xiexin12138 !

PR #646, opened 2023-07-29
Feature/google custom search

denchi:
Added Google Custom Search.

Here is an example of a flowgraph using it.
Screenshot 2023-07-29 at 11 29 17


packages/components/credentials/GoogleSearchApi.credential.ts
```diff
@@ -0,0 +1,29 @@
...
+    label: string
+    name: string
+    version: number
+    description: string
+    inputs: INodeParams[]
+
+    constructor() {
+        this.label = 'Google Custom Search API'
+        this.name = 'googleCustomSearchApi'
+        this.version = 1.0
```
HenryHengZJ, line 13: can we add a description like [this](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/credentials/ZepMemoryApi.credential.ts#L14) to let user know how and where to get these credentials?

denchi: Sure. Just added the description.

HenryHengZJ (approved):
thank you @denchi !

Issue #731, opened 2023-08-10
[FEATURE] parameterize the Folder with Files Document Loader to allow custom file extensions


rkeshwani:
As a user I'd like to add the ability to add additional file extensions to the folder with files loader so that I can load documents that aren't just text documents or the pre-coded file extensions like code for example.

One suggestion might be to allow a user to add a file extension, and then in a dropdown select from a list of pre-built loaders.
This list would include: JSONLoader, CSVLoader, TextLoader, JSONLinesLoader, PDFLoader, DOCXLoader, EPUBLoader, and UnstructuredLoader

rkeshwani:
Note: The code needs to be tested, I was unable to get it to build locally even without my changes.

rkeshwani:
This has a potential issue in that it indexes everything including node_modules, nuget packages, ruby gems, etc..
So might have to open a new ticket for reading the .gitignore file while iterating through the files or adding that as a additional parameter to respect the gitignore file.

PR #733, opened 2023-08-11
Issue #731: Add additional optional input parameter for adding additional file loaders

rkeshwani:
For now added additional input parameters to allow addition of more file loaders. Not sure if this will work well. I couldn't get the main brand to build successfully locally so I couldn't run any of the tests. I highly suggest testing this locally before approving the pull request. Original issue: Issue #731.

### DIFF
--- a/packages/components/nodes/documentloaders/Folder/Folder.ts
+++ b/packages/components/nodes/documentloaders/Folder/Folder.ts
@@ -46,7 +46,7 @@ class Folder_DocumentLoaders implements INode {
                 type: 'json',
                 optional: true,
                 additionalParams: true
-            },
+            }
         ]
     }
 
@@ -84,9 +84,9 @@ class Folder_DocumentLoaders implements INode {
             '.rs': (path) => new TextLoader(path), // Rust
             '.scala': (path) => new TextLoader(path), // Scala
             '.sc': (path) => new TextLoader(path), // Scala
-            '.scss': (path) => new TextLoader(path),// Sass
+            '.scss': (path) => new TextLoader(path), // Sass
             '.sol': (path) => new TextLoader(path), // Solidity
-            '.sql': (path) => new TextLoader(path),//SQL
+            '.sql': (path) => new TextLoader(path), //SQL
             '.swift': (path) => new TextLoader(path), // Swift
             '.markdown': (path) => new TextLoader(path), // Markdown
             '.md': (path) => new TextLoader(path), // Markdown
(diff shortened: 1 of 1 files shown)


packages/components/nodes/documentloaders/Folder/Folder.ts
```diff
@@ -54,14 +61,17 @@ class Folder_DocumentLoaders implements INode {
...
 
         const loader = new DirectoryLoader(folderPath, {
             '.json': (path) => new JSONLoader(path),
             '.txt': (path) => new TextLoader(path),
             '.csv': (path) => new CSVLoader(path),
             '.docx': (path) => new DocxLoader(path),
             // @ts-ignore
-            '.pdf': (path) => new PDFLoader(path, { pdfjs: () => import('pdf-parse/lib/pdf.js/v1.10.100/build/pdf.js') })
+            '.pdf': (path) => new PDFLoader(path, { pdfjs: () => import('pdf-parse/lib/pdf.js/v1.10.100/build/pdf.js') }),
+            ...parsedLoaders
```
HenryHengZJ, line 74: I don't think it will work. Because you need specific loader for specific file types, otherwise if let's say you have a file with extension `.ppt`, there is no loader to process the file

rkeshwani: Well this is the tricky part because you have to leave the loader selection in the hands of the user right? 

This might require an addition of a new type of user control to flowise-ui that allows a user to type in a file type and select a loader from the list. Then in this code you could create a new function that returns a instance of the loader depending on the selected loader. 

There are hundreds of file types that don't fit in this list that could use just the TextLoader. Alternatively do you suggest that I add some of the basic code extensions here? like .cs, .php, .js, .xml, .cpp, etc?

HenryHengZJ: I think for now we can just add the other extensions here. For example: 

`'.cs': (path) => new TextLoader(path),`

You can refer to all available code extensions [here](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/textsplitters/CodeTextSplitter/CodeTextSplitter.ts#L33)

HenryHengZJ (approved):

HenryHengZJ:
@rkeshwani a little lint issue, if you can run `yarn lint-fix` that'll do

rkeshwani:
@HenryHengZJ done. Spaces missing after comments and commas apparently were the issue.

HenryHengZJ:
thanks @rkeshwani !

PR #757, opened 2023-08-13
Add Vectara templates to marketplace

3eif:
This PR adds a Vectara templates to the Flowise marketplace, a template for uploading docs and a template for loading from existing documents. The template looks like this:

Screenshot 2023-08-11 at 10 33 10 PM

Screenshot 2023-08-11 at 10 32 40 PM


chungyau97:
IMO, I feel this not required as this is very similar towards `Conversational Retrieval QA Chain`.
Previously you added the [Vectara documentation](https://github.com/FlowiseAI/FlowiseDocs/blob/main/vector-stores/vectara.md) is enough to guide them on setting up.

HenryHengZJ (commented):
thanks @3eif for the templates! I think we can use one of the templates here, maybe the upsert one. since there are already multiple templates in marketplace that has similar flows

3eif:
> thanks @3eif for the templates! I think we can use one of the templates here, maybe the upsert one. since there are already multiple templates in marketplace that has similar flows

Done!

HenryHengZJ (approved):

Issue #668, opened 2023-08-01
[FEATURE] ADD pgvector Vector Store


mahmoudsorf:
**Describe the feature you'd like**
It will be good to add pgvector from postgresql to the list of supported vector stores

rkeshwani:
Note: Langchain has support for the Prisma client which supports querying postgres vectors stores utilizing the pgvector extension.
See: https://js.langchain.com/docs/modules/data_connection/vectorstores/integrations/prisma
Could this be integated into Flowise?.

chungyau97:
Hi @rkeshwani,

You can reference some [vector store source code](https://github.com/FlowiseAI/Flowise/tree/main/packages/components/nodes/vectorstores) to implement it.

rkeshwani:
Build will not work because of the following issue:
There seems to be an issue with langchain itself and prisma.
LangchainJS's type definitions may need to be updated to match new Prisma's specifications in the following commit: https://github.com/prisma/prisma/commit/44bc6576b8c5b5dbfe5533e1b0ef2e34f8025d48.

This commit was pushed 3 weeks ago but the Prisma namespace definitions in the langchainJS repo were last pushed over a month ago:
https://github.com/hwchase17/langchainjs/blob/main/langchain/src/vectorstores/prisma.ts
https://github.com/hwchase17/langchainjs/commit/460e110885fb349bc5d946fbd678879a8d8eedf0

Once this is resolved, this code should work correctly. So there is a dependency to update the Prisma Namespace type definitions to match the prisma client object exported by the @prisma/client library.

I will open a issue with langchainjs.

HenryHengZJ:
closing as #782 merged

PR #759, opened 2023-08-13
Issue #668 Feature/prisma vector store

rkeshwani:
Initial attempt at adding prisma functionality. 
Build will not work because of the following issue: 
There seems to be an issue with langchain itself and prisma. 
LangchainJS's type definitions may need to be updated to match new Prisma's specifications in the following commit: https://github.com/prisma/prisma/commit/44bc6576b8c5b5dbfe5533e1b0ef2e34f8025d48.

This commit was pushed 3 weeks ago but the Prisma namespace definitions in the langchainJS repo were last pushed over a month ago:
https://github.com/hwchase17/langchainjs/blob/main/langchain/src/vectorstores/prisma.ts
https://github.com/hwchase17/langchainjs/commit/460e110885fb349bc5d946fbd678879a8d8eedf0

Once this is resolved, this code should work correctly. So there is a dependency to update the Prisma Namespace type definitions to match the prisma client object exported by the @prisma/client library.
Issue Reference: #668


rkeshwani:
@HenryHengZJ I have an issue open with them but I have not opened a PR.
See: https://github.com/hwchase17/langchainjs/issues/2265

rkeshwani:
Note: this might not be needed if https://github.com/FlowiseAI/Flowise/pull/782 works well. As it works directly with postgres using a different library specifically for postgress.

HenryHengZJ:
@rkeshwani #782 is merged, let me know if that works for you, if so we can close this

PR #762, opened 2023-08-14
Fix Vectara filter

3eif:
This PR fixes the Vectara metadata filter type

### DIFF
--- a/packages/components/nodes/vectorstores/Vectara_Existing/Vectara_Existing.ts
+++ b/packages/components/nodes/vectorstores/Vectara_Existing/Vectara_Existing.ts
@@ -34,7 +34,7 @@ class VectaraExisting_VectorStores implements INode {
             {
                 label: 'Vectara Metadata Filter',
                 name: 'filter',
-                type: 'json',
+                type: 'string',
                 additionalParams: true,
                 optional: true
             },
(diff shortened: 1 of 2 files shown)


packages/components/nodes/vectorstores/Vectara_Existing/Vectara_Existing.ts
```diff
@@ -34,7 +34,7 @@ class VectaraExisting_VectorStores implements INode {
             {
                 label: 'Vectara Metadata Filter',
                 name: 'filter',
-                type: 'json',
+                type: 'string',
                 additionalParams: true,
```
HenryHengZJ, line 38: can we add a `description` with how to use the filter, or the link to the docs page? Something like:

```
Refer to <a target="_blank" href="https://docs.vectara.com">official guide</a> on how to use filter
```

3eif: This has been added in https://github.com/FlowiseAI/FlowiseDocs/pull/22

HenryHengZJ: thanks for the docs PR! would be good to add a description property as well

`description: 'Refer to....'`

3eif: Ah I didn't know there was a description property. Will add that as well!

3eif: @HenryHengZJ Added the description property. Let me know if there is anything else I should change.

HenryHengZJ (approved):
looks good! thanks @3eif

Josephrp:
thank you for this ! 🚀

PR #782, opened 2023-08-16
Add Postgres vector store using pgvector

fabiovalencio:
Add a component to vectorstore to allow saving vectors in a Postgres database with pgvector


HenryHengZJ:
you might have accidentally removed the `.env.example` files

fabiovalencio:
> you might have accidentally removed the `.env.example` files

Sorry about that, I just renamed them to use, putting back.

HenryHengZJ:
@fabiovalencio were you abel to get this running? I have a docker compose file running like this:
```
services:
  db:
    image: ankane/pgvector
    ports:
      - 5432:5432
    volumes:
      - ./data:/var/lib/postgresql/data
    environment:
      - POSTGRES_PASSWORD=mypassword
      - POSTGRES_USER=myuser
      - POSTGRES_DB=api
  ```

And I created postgres credential in Flowise with the above user and password. However, when running the chatflow got this error: `password authentication failed for user "myuser"`

**Comment by fabiovalencio** [comment]
Hi @HenryHengZJ this probably happened because I hadn't set the default table name, I've just fixed it.

**Comment by HenryHengZJ** [comment]
> Hi @HenryHengZJ this probably happened because I hadn't set the default table name, I've just fixed it.

still the same: 
image

**Comment by fabiovalencio** [comment]
I redid it from the beginning and didn't find any errors.

I started by creating Postgres with Docker
Screenshot 2023-08-18 at 10 17 30

Then I tried accessing the database
Screenshot 2023-08-18 at 10 35 04

I didn't use Flowise from Docker because it doesn't have the Postgres vector store yet. So I started another application with Flowise and I saved the Postgres access credentials.
Screenshot 2023-08-18 at 10 48 40

I tried to create a chat just like yours
Screenshot 2023-08-18 at 10 55 19

Screenshot 2023-08-18 at 11 13 34


Screenshot 2023-08-18 at 10 55 52

I haven't really had any problems. Just wondering, can you access the database via psql?

**Comment by HenryHengZJ** [comment]
hmm pretty sure its my connection to postgres..

In the docker compose file I have set
```
- POSTGRES_USER=myuser
- POSTGRES_PASSWORD=mypassword
- POSTGRES_DB=api
```

Then using cli:
```
psql postgresql://abc:123@localhost:5432/api
```
I was still able to get into psql, even though the user and password are clearly not the one in docker compose

File: packages/components/nodes/vectorstores/Postgres_Upsert/Postgres_Upsert.ts [hunk 1]
```diff
@@ -0,0 +1,212 @@
...
+        this.type = 'Postgres'
+        this.icon = 'postgres.svg'
+        this.category = 'Vector Stores'
+        this.description = 'Upsert documents to Postgres using pgvector'
+        this.baseClasses = [this.type, 'VectorStoreRetriever', 'BaseRetriever']
+        this.credential = {
+            label: 'Connect Credential',
+            name: 'credential',
+            type: 'credential',
+            description: 'Needed when using Postgres cloud hosted',
```
HenryHengZJ, line 36: is the credential optional here?

fabiovalencio: Sure not! I don't know why the upsert was different from the existing one

HenryHengZJ:
thank you @fabiovalencio !

HenryHengZJ (approved):

rkeshwani:
This is returning a Error: Error: Credentials could not be decrypted on a windows machine running docker version of the database.
Note, I'm starting this based on the basic instructions without any configurations.

rkeshwani:
Getting a different error now:
TypeError: this.embeddings.embedQuery is not a function
    at TypeORMVectorStore.similaritySearch(
flowise\node_modules\langchain\dist\vectorstores\base.cjs:95:90)

This is while using a Conversational Retrieval QA Chain with ChatGPT and pgvector existing index and buffer memory. But based on the error it looks like the issue is specific to the pgvector. Perhaps a missing method?

raymatos:
would it be possible to have flowise create the database and enable the vector extension when adding a database name that doesnt exist

PR #784, opened 2023-08-16
Puppeteer & Playwright additional config options

atishamte:
Puppeteer & Playwright takes additional configs to load the data from webpages.


HenryHengZJ (approved):
thanks @atishamte !!

PR #824, opened 2023-08-24
add skip extra credential on google credential

Yongtae723:
If you use GCP service like cloud RUN, you don't have to set extra credential.
Not to registering extra credential is also good for security point of view.

Thus I implement Skip button on google credential.
If you ON this bottom, Flowise use Vertex AI without credential setting.

if you feel good, please merge it after #823 

BTW : since I added button on credential, I add the same code for validate input of credential for a couple of place. but if this validate can do with credential.ts, I think architecture would be clear and not to lead mis-coding
like I proposed on [Discord](https://discord.com/channels/1087698854775881778/1087698855342125118/1143432037798256691)


HenryHengZJ:
hey I'm confused, do you mean its optional to use credential for Vertex? Do you have any link to docs, or step by step guide etc? If its optional, we don't have to include this switch on/off button, instead you can just use `optional` like for [Zep node](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/memory/ZepMemory/ZepMemory.ts#L32)

Yongtae723:
@HenryHengZJ 
>do you mean its optional to use credential for Vertex? 

Yes. If we use GCP service like cloud run or if we install default credential on  local machine, credential is Optional
I think [this](https://cloud.google.com/docs/authentication/provide-credentials-adc#local-dev) is one of the document which descrive credential

>If its optional, we don't have to include this switch on/off button, instead you can just use optional like for [Zep node](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/memory/ZepMemory/ZepMemory.ts#L32)

oh sorry simply I didn't know credential can be optional.
as your propose, Optional credential make much more sense to me.
I will modify the code like your propose.

Yongtae723:
@HenryHengZJ I updated!

so could you review it?
thanks!

HenryHengZJ:
thanks for the change! I think we can close the PR on docs (https://github.com/FlowiseAI/FlowiseDocs/pull/26) ?

HenryHengZJ (approved):

Yongtae723:
Yes!

I will close it!
thank you!!

Yongtae723:
Hi @HenryHengZJ 

if there are no problem, could you merge this PR?

Thanks in advance!

HenryHengZJ:
sorry for the delay!

Yongtae723:
Perfect!

thank you!

PR #838, opened 2023-08-27
Update docker-compose.yml

shalalalaw:
Helps avoid the error docker users have when trying to implement Langsmith via their environment variables.


chungyau97:
Hi @shalalalaw, 

I feel this is better to be written in [Flowise Docs](https://github.com/flowiseai/flowisedocs)

Reference:
[Setup Chroma and Flowise in Docker](https://docs.flowiseai.com/vector-stores/chroma#additional)

HenryHengZJ:
thanks @shalalalaw , we are working on a UI way that allow users to set langsmith credentials, so will close this for now

Issue #839, opened 2023-08-27
[BUG] Error: QueryFailedError: invalid input syntax for type uuid: ""


isikhi:
**Describe the bug**
When credentials not settled properly the ui app throws  it
```
Error: QueryFailedError: invalid input syntax for type uuid: ""
```

**To Reproduce**

1. Go to any kind of schema
2. Start chat without credentials

**Expected behavior**
A clear and humanizied error to understand what is wrong with in

**Screenshots**
![image](https://github.com/FlowiseAI/Flowise/assets/41375111/98e7e334-7ef4-455b-935d-1b88e441820e)

**Flow**
If applicable, add exported flow in order to help replicating the problem.

**Setup**
-   Installation [e.g. docker, `npx flowise start`, `yarn start`]
-   Flowise Version [e.g. 1.2.11]
-   OS: [e.g. macOS, Windows, Linux]
-   Browser [e.g. chrome, safari]

**Additional context**
I will send a pr/mr for it

PR #840, opened 2023-08-27
fix: unknown error throwing on missing credentials. error handling improves on ui section

isikhi:
Closes/Related: (https://github.com/FlowiseAI/Flowise/issues/839)

### DIFF
--- a/packages/components/src/utils.ts
+++ b/packages/components/src/utils.ts
@@ -432,6 +432,10 @@ export const getCredentialData = async (selectedCredentialId: string, options: I
     const databaseEntities = options.databaseEntities as IDatabaseEntity
 
     try {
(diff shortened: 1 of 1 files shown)


HenryHengZJ (approved):

isikhi:
I found it while exploring different places within the app. Maybe a different state in the front-end is throwing an instant error. Since I am away from computer right now, I can't say exactly in which state it throwing. But I don't think this code will have any side effects. If someone didn't post the credentials correctly, we can also qualify it as an extra protection code that gives the same answer as wrong credentails. So it can be merged in. When I go to the computer and review the project again, I will reproduce the error again and inform you. 👋

HenryHengZJ:
perfect, yeah I agree

PR #873, opened 2023-09-04
feature/Migration

chungyau97:
### Results:
**Previous user existing data is not affected**
image

image

image

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,8 @@
         "lint": "eslint \"**/*.{js,jsx,ts,tsx,json,md}\"",
         "lint-fix": "yarn lint --fix",
         "quick": "pretty-quick --staged",
-        "postinstall": "husky install"
+        "postinstall": "husky install",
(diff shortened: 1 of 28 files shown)


AidanTilgner (commented):
This looks pretty good to me. I don't see any major problems and after testing locally it seems that sqlite at least is working well. I didn't do too much testing outside of local stuff, and I'll be honest I'm not super familiar with postgres. However, everything seems pretty standard and it looks good to me.

chungyau97:
> This looks pretty good to me. I don't see any major problems and after testing locally it seems that sqlite at least is working well. I didn't do too much testing outside of local stuff, and I'll be honest I'm not super familiar with postgres. However, everything seems pretty standard and it looks good to me.

Great to hear. Thank you for your time and effort in reviewing! 😄
I have tested this PR on Local [SQLite, MySQL, PostgreSQL] and Render [SQLite], which shows that the user's previous data still exists and has not disappeared.

PR #882, opened 2023-09-05
PaLM API LLM component

mdda:
This is a first draft of the the LLM component for the PaLM API (ref: #862)

Screenshot_2023-09-06_16-24-33

Could you have a look at the overall style? 

Once this part is satisfactory, I can do PaLM-Chat and PaLM-embeddings to match...

For the update to package.json : I wasn't sure how to make sure this installed when building overall, since it's required by langchain, but didn't seem to have installed itself.

I noticed that the 'json' field in the optional parameters don't seem to be properly editable - either on this component, or on the existing OpenAI one - so maybe that needs an additional fix.

PR #891, opened 2023-09-07
PaLM API Chat and Embedding components

mdda:
This is a follow-on from the recently accepted PaLM API LLM component (#882).

Working Chat Flow

Working Embeddings Flow

### DIFF
--- a/packages/components/nodes/chatmodels/ChatGooglePaLM/ChatGooglePaLM.ts
+++ b/packages/components/nodes/chatmodels/ChatGooglePaLM/ChatGooglePaLM.ts
@@ -1,0 +1,140 @@
+import { ICommonObject, INode, INodeData, INodeParams } from '../../../src/Interface'
+import { getBaseClasses, getCredentialData, getCredentialParam } from '../../../src/utils'
+import { ChatGooglePaLM, GooglePaLMChatInput } from 'langchain/chat_models/googlepalm'
+
+class ChatGooglePaLM_ChatModels implements INode {
+    label: string
+    name: string
+    version: number
+    type: string
+    icon: string
+    category: string
+    description: string
+    baseClasses: string[]
+    credential: INodeParams
+    inputs: INodeParams[]
+
+    constructor() {
+        this.label = 'ChatGooglePaLM'
+        this.name = 'chatGooglePaLM'
+        this.version = 1.0
+        this.type = 'ChatGooglePaLM'
+        this.icon = 'Google_PaLM_Logo.svg'
+        this.category = 'Chat Models'
+        this.description = 'Wrapper around Google MakerSuite PaLM large language models using the Chat endpoint'
+        this.baseClasses = [this.type, ...getBaseClasses(ChatGooglePaLM)]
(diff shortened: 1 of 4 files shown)


packages/components/nodes/chatmodels/ChatGooglePaLM/ChatGooglePaLM.ts
```diff
@@ -0,0 +1,140 @@
...
+                additionalParams: true,
+                description:
+                    'Top-k changes how the model selects tokens for output.\n' +
+                    'A top-k of 1 means the selected token is the most probable among ' +
+                    'all tokens in the model vocabulary (also called greedy decoding), ' +
+                    'while a top-k of 3 means that the next token is selected from ' +
+                    'among the 3 most probable tokens (using temperature).'
+            }
+            /*
+            {
```
HenryHengZJ, line 90: can we remove commented unused code?

packages/components/nodes/chatmodels/ChatGooglePaLM/ChatGooglePaLM.ts
```diff
@@ -0,0 +1,140 @@
...
+            modelName: modelName,
+            temperature: parseFloat(temperature),
+            apiKey: googleMakerSuiteKey
+        }
+
+        if (topP) obj.topP = parseFloat(topP)
+        if (topK) obj.topK = parseFloat(topK)
+
+        /*
+        let parsedExamples: any | undefined = undefined
```
HenryHengZJ, line 124: same here

packages/components/nodes/chatmodels/ChatGooglePaLM/ChatGooglePaLM.ts
mdda: Yes it is unused at the moment, but, OTOH, it is available for use in langchain.js (https://github.com/hwchase17/langchainjs/blob/main/langchain/src/chat_models/googlepalm.ts#L132) : It's just that Flowise doesn't make use of that available feature (also available in the VertexAI models, but unused).

I'm happy to remove the code - but put it in as a placeholder in case it inspired someone...  My guess is that langchain itself won't make use of this prior examples[] data until it appears in the OpenAI API.

mdda: In the interests of getting this into 'main', I've removed the commented out code as first suggested.  The extra fields (and the associate UI) might be worth having if we're convinced that langchain would make good use of them (i.e. more efficiently implement few-shot prompts on the PaLM models), but my guess is the default option for the backend is to just add as previous context.  So, really, there's currently little to be gained from added these fields at the moment.

HenryHengZJ: that make sense, I'm good with that. we can always add in that if there is a need in the future

HenryHengZJ (approved):

HenryHengZJ:
thank you for the effort @mdda !

PR #913, opened 2023-09-12
Add Vectara upload file component

3eif:
This component utilizes Vectara's upload endpoint which supports uploading and indexing of any document, not just text. 

This allows users to upload file types such as docx, html, epub, and more without needing document loaders. Those files are then processed to extract and chunk the text.

image

This PR also adds support for querying from multiple corpora at a time.

I had to update the langchainjs package and some class paths to get the component working, I hope this is okay.


packages/components/nodes/vectorstores/Vectara/Vectara_Upload.ts
```diff
@@ -0,0 +1,177 @@
...
+            const splitDataURI = file.split(',')
+            splitDataURI.pop()
+            const bf = Buffer.from(splitDataURI.pop() || '', 'base64')
+            const blob = new Blob([bf])
+            vectaraFiles.push({ blob: blob, fileName: getFileName(file) })
+        }
+
+        const vectorStore = new VectaraStore(vectaraArgs)
+        await vectorStore.addFiles(vectaraFiles)
+        files = []
```
HenryHengZJ, line 147: this seems redundant?

3eif: You're right. I've removed it

HenryHengZJ (approved):

HenryHengZJ:
@3eif thanks for this! There is a merge conflict if you can resolve - I think its the langchainjs version, you can use the latest one here.

Another thing, I think it would be good to update the marketplace Vectara template to use this upload node, instead of the upsert node, mainly because this is slightly different than the upsert node and its good to let others know what they can do differently using Vectara

3eif:
Thanks. I've resolved the merge conflicts and updated the marketplace template! @HenryHengZJ

HenryHengZJ:
awesome thanks!

PR #948, opened 2023-09-19
Added CSS selector to Cheerio

matthias:
Added selector parameter to Cheerio web loader

As described here: https://js.langchain.com/docs/modules/data_connection/document_loaders/integrations/web_loaders/web_cheerio

Improves the quality of the scraped content A LOT


HenryHengZJ:
awesome! @matthias what selector do you use that you can see improvement?

matthias:
We ingest from this Site: (example page)
https://www.waff.at/beruf-weiterbildung/karenz-und-wiedereinstieg/

The standard selector (body) will also ingest all menus, sidebars, footers, etc ... which add a lot of useless information. 

So I can add "body > main > article" as a selector in this case.

HenryHengZJ (approved):

matthias:
🙏

PR #954, opened 2023-09-20
Supabase_Upsert: Prevent duplicates with the same meta->source

matthias:
This PR solves the issue of duplicate embeddings for the same "source".

As all Loaders (as far as I'm aware) provide a metadata.source -> we use this field to:

1. Collect all source strings from incoming Documents
2. Delete all existing embeddings with a matching metadata.source
3. Go on with upserting (creating) the embeddings

### DIFF
--- a/packages/components/nodes/vectorstores/Supabase_Upsert/Supabase_Upsert.ts
+++ b/packages/components/nodes/vectorstores/Supabase_Upsert/Supabase_Upsert.ts
@@ -22,7 +22,7 @@ class SupabaseUpsert_VectorStores implements INode {
     constructor() {
         this.label = 'Supabase Upsert Document'
         this.name = 'supabaseUpsert'
-        this.version = 1.0
+        this.version = 1.1
         this.type = 'Supabase'
         this.icon = 'supabase.svg'
         this.category = 'Vector Stores'
@@ -94,6 +94,7 @@ class SupabaseUpsert_VectorStores implements INode {
         const output = nodeData.outputs?.output as string
         const topK = nodeData.inputs?.topK as string
         const k = topK ? parseFloat(topK) : 4
+        const logger = options.logger
 
         const credentialData = await getCredentialData(nodeData.credential ?? '', options)
         const supabaseApiKey = getCredentialParam('supabaseApiKey', credentialData, nodeData)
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
this only works for web scraper nodes? because other document loaders have different type or source:

PDF:
```
{
      "pageContent": "abc",
      "metadata": {
        "source": "blob",
        "blobType": "",
        "pdf": {
          "version": "1.10.100",
          "info": {
            "PDFFormatVersion": "1.5",
            "IsAcroFormPresent": false,
            "IsXFAPresent": false,
            "Title": "",
            "Author": "",
            "Subject": "",
            "Keywords": "",
            "Creator": "LaTeX with hyperref",
            "Producer": "pdfTeX-1.40.21",
            "CreationDate": "D:20230414003603Z",
            "ModDate": "D:20230414003603Z",
            "Trapped": {
              "name": "False"
            }
          },
          "metadata": null,
          "totalPages": 17
        },
        "loc": {
          "pageNumber": 8,
          "lines": {
            "from": 301,
            "to": 317
          }
        }
      }
    }
```
Web: 
```
{
        "pageContent": "Main article: Views of Elon Musk",
        "metadata": {
          "source": "https://en.wikipedia.org/wiki/Elon_Musk",
          "loc": {
            "lines": {
              "from": 2409,
              "to": 2409
            }
          }
        }
    }
```
Github:
```
 {
      "pageContent": "abc",
      "metadata": {
        "source": "README.md",
        "loc": {
          "lines": {
            "from": 176,
            "to": 184
          }
        }
      }
    },
```
CSV:
```
{
      "pageContent": "First Name: John\nLast Name: Doe\nAddress: 120 jefferson st.\nStates: Riverside\nCode: NJ\nPostal: 8075",
      "metadata": {
        "source": "blob",
        "blobType": "",
        "line": 1,
        "loc": {
          "lines": {
            "from": 1,
            "to": 6
          }
        }
      }
    },
```

As you can see the source is always the same for other doc loaders

matthias:
I tested Folder with Files - where you also get the path.
But you are right on a lot of other cases - especially on direct uploads.

So for now - not a general solution for Langchain / Flowise...

If you want to close this PR - fine for me. Will investigate further.

HenryHengZJ:
think the right solution is to use cache - https://blog.langchain.dev/syncing-data-sources-to-vector-stores, will close this for now

matthias:
Cool. Thanks for looking into it. And thx for the link.

matthias:
So the indexing API from langchain is in developemnt - currently only in the Python API

Ultimatly they also use source="", while you are able to specify the field via `source_id_key` when calling the API functions.

`index(
    changed_doggy_docs,
    record_manager,
    vectorstore,
    cleanup="incremental",
    source_id_key="source",
)`

Based on this it (will) provide further functions for a better upserting experience. 
This needs to be implemented (by Langchain?) for every VectoreStore.



The documentation also states:

> The metadata attribute contains a filed called source. **This source should be pointing at the ultimate provenance associated with the given document**.
>
> For example, if these documents are representing chunks of some parent document, the source for both documents should be the same and reference the parent document.
>
> In general, source should always be specified. Only use a None, if you never intend to use incremental mode, and for some reason can't specify the source field correctly.

And:

> Attention: The loader must set source keys correctly.

So if the PDF File is uploaded via PDFLoader we would need to pass a source (unique source identifier) as a metadata to PDFLoader ... as PDFLoader itself has no knowledge about the original path of the document (it just gets a blob)

https://github.com/langchain-ai/langchainjs/blob/6e1a112/langchain/src/document_loaders/fs/pdf.ts#L67

---
### Conclusion

In my opinion we should check the Flowise loaders which don't provide a proper 'source' ID (like PDFLoader), so they pass a source, based on the original upload path / document - or/and - provide a text input field so the user can enter a source (unique source identifier) 

And then wait, until Langchain provides JS API function which we can use with '*Upsert' nodes

PR #981, opened 2023-09-28
Update README.md to bring Docs higher up

pooriaarab:


HenryHengZJ:
these are the icons for README, adding the docs doesnt fit nicely.
image

PR #989, opened 2023-09-29
Feature/llm cache 

vinodkiran:
Support LLM Cache 
 - Add Cache as Optional Input for LLM Models - DONE
 - Cache Memory Implementations
    - In Memory Cache - OUT OF SCOPE FOR THIS PR
    - Momento Cache - DONE
    - Redis - DONE
    - Uptash Redis - DONE
 - Add Cache as Optional Input for Chat Models - DONE


lucasbuges:
Thanks @HenryHengZJ !!!!

packages/components/src/handler.ts
```diff
@@ -175,13 +177,31 @@ export class CustomChainHandler extends BaseCallbackHandler {
     }
 
     handleLLMEnd() {
-        this.socketIO.to(this.socketIOClientId).emit('end')
+        /* send the end event from handleChainEnd */
+        // this.socketIO.to(this.socketIOClientId).emit('end')
```
HenryHengZJ, line 181: why is this commented out?

packages/components/package.json
```diff
@@ -44,7 +45,7 @@
         "graphql": "^16.6.0",
         "html-to-text": "^9.0.5",
         "ioredis": "^5.3.2",
-        "langchain": "^0.0.154",
+        "langchain": "^0.0.157",
```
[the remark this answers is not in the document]
HenryHengZJ: yeah the latest main is already using 0.0.157

HenryHengZJ:
@vinodkiran thank you so much for all the changes!

Few thoughts: Can we extend the cache support to Chat model as well? Looking at the [docs](https://js.langchain.com/docs/modules/model_io/models/chat/how_to/llm_caching), the cache seems to also compatible with chat models. If it is doable, can we rename the category to just `cache` without the `llm`?

packages/components/nodes/llmcache/LLMMomentoCache/LLMMomentoCache.ts
```diff
@@ -0,0 +1,75 @@
...
+                baseClasses: [this.type, ...getBaseClasses(MomentoCache)]
+            }
+        ]
+    }
+
+    async init(nodeData: INodeData, _: string, options: ICommonObject): Promise<any> {
+        const credentialData = await getCredentialData(nodeData.credential ?? '', options)
+        const apiKey = getCredentialParam('momentoApiKey', credentialData, nodeData)
+        const cacheName = getCredentialParam('momentoCache', credentialData, nodeData)
+        const endPoint = getCredentialParam('momentoEndpoint', credentialData, nodeData)
```
HenryHengZJ, line 56: is this being used?

HenryHengZJ:
hmm something is not working correctly.. the cache is working fine, but if the question is not the same, the same cache answer is being used:

image

packages/components/nodes/llmcache/LLMInMemoryCache/LLMInMemoryCache.ts
```diff
@@ -0,0 +1,36 @@
+import { INode, INodeParams } from '../../../src/Interface'
```
HenryHengZJ, line 1: Tested InMemoryCache, it won't work in Flowise. Reason because the class is being reinitalized everytime, hence making the cache `Map` to be reset everytime.

For now, we can get rid of this. I have a solution in place, will open a PR after this get merged

packages/components/src/handler.ts
vinodkiran: Commented as the 'end' has to be sent on Chain End, Line 204. Sending on LLM End will mean that the end is NOT sent for cached responses, hence the need to send in Chain End.

packages/components/nodes/llmcache/LLMMomentoCache/LLMMomentoCache.ts
vinodkiran: miss at my end will clean up the unsed vars.

packages/components/nodes/llmcache/LLMInMemoryCache/LLMInMemoryCache.ts
vinodkiran: OK.

vinodkiran:
> @vinodkiran thank you so much for all the changes!
> 
> Few thoughts: Can we extend the cache support to Chat model as well? Looking at the [docs](https://js.langchain.com/docs/modules/model_io/models/chat/how_to/llm_caching), the cache seems to also compatible with chat models. If it is doable, can we rename the category to just `cache` without the `llm`?

Yes, will add the feature, it was on the original 'todo' list anyway.

vinodkiran:
> @vinodkiran thank you so much for all the changes!
> 
> Few thoughts: Can we extend the cache support to Chat model as well? Looking at the [docs](https://js.langchain.com/docs/modules/model_io/models/chat/how_to/llm_caching), the cache seems to also compatible with chat models. If it is doable, can we rename the category to just `cache` without the `llm`?

Makes sense! Will update the category name.

packages/components/nodes/llmcache/LLMInMemoryCache/LLMInMemoryCache.ts
HenryHengZJ: In this PR, we won't be able to let users select simple cache. Will open another PR to allow users to do that. We'll have to create a global cache pool in flowise for that, similar to `ChatflowPool` in `server`

packages/components/src/handler.ts
HenryHengZJ: ah okay make sense

packages/components/nodes/chatmodels/AWSBedrock/AWSChatBedrock.ts
```diff
@@ -39,6 +41,12 @@ class AWSChatBedrock_ChatModels implements INode {
             optional: true
         }
         this.inputs = [
+            {
+                label: 'Cache',
+                name: 'cache',
+                type: 'LLMCache',
```
HenryHengZJ, line 47: Can we find and replace all `LLMCache` with `BaseCache`

vinodkiran: Sorry, that was sloppy on my end. :-) I have changed all and pushed the new commit.

packages/components/nodes/llmcache/LLMInMemoryCache/LLMInMemoryCache.ts
vinodkiran: Done! Removed InMemoryCache for now, we will do a new feature to enable the local cache.

packages/components/nodes/chatmodels/AWSBedrock/AWSChatBedrock.ts
```diff
@@ -138,7 +138,7 @@ class AWSChatBedrock_ChatModels implements INode {
         const iModel = nodeData.inputs?.model as string
         const iTemperature = nodeData.inputs?.temperature as string
         const iMax_tokens_to_sample = nodeData.inputs?.max_tokens_to_sample as string
-        const cache = nodeData.inputs?.llmCache as BaseCache
+        const cache = nodeData.inputs?.cache as BaseCache
```
vinodkiran, line 141: One might need to delete the nodes from saved flows and recreate them for this change to be seen.

packages/components/nodes/llmcache/LLMMomentoCache/LLMMomentoCache.ts
vinodkiran: Done! Removed the unused vars.

packages/components/nodes/chatmodels/AWSBedrock/AWSChatBedrock.ts
HenryHengZJ: yes exactly

packages/components/src/handler.ts
```diff
@@ -178,9 +180,30 @@ export class CustomChainHandler extends BaseCallbackHandler {
...
-    handleChainEnd(outputs: ChainValues): void | Promise<void> {
-        if (this.returnSourceDocuments) {
-            this.socketIO.to(this.socketIOClientId).emit('sourceDocuments', outputs?.sourceDocuments)
+    handleChainEnd(outputs: ChainValues, _: string, parentRunId?: string): void | Promise<void> {
+        /*
+            Langchain does not call handleLLMStart, handleLLMEnd, handleLLMNewToken when the chain is cached.
+            Callback Order is "Chain Start -> LLM Start --> LLM Token --> LLM End -> Chain End" for normal responses.
+            Callback Order is "Chain Start -> Chain End" for cached responses.
+         */
+        if (this.cachedResponse && parentRunId === undefined) {
```
HenryHengZJ, line 189: found out that if used in ConversationRetrievalQAChain where there are multiple chains, this get triggered multiple times. Adding check for `parentRunId` to keep track what is the last chain, and only emit token for that

HenryHengZJ (approved):
huge thanks to @vinodkiran !!

lucasbuges:
@HenryHengZJ  @vinodkiran  Thank you very much for implementing LLM cache. I managed to test it here and it works perfectly. But I have a question, what should I do when the AI caches a wrong answer?

PR #995, opened 2023-10-02
Duplicate icon when market is empty

vincent-pli:
See the snapshot:
image


chungyau97:
Hi @vincent-pli, thank you very much! We appreciate your fix a lot.👍

PR #1017, opened 2023-10-09
credential should be-rebuild eveytimes

vincent-pli:
When update the `credential` in `Credentials` page,  the `chatflow` will not rebuild, that's make the new `credential` not effected unless people click `save chatflow` button to make it `unsync`.

I think the chatflow with start node has `credential` should be rebuild every time.


HenryHengZJ:
but when you edit the credential, you have to save your flow in order to proceed right? because having rebuild every flow with credential as start node might cause more latency

vincent-pli:
@HenryHengZJ yes, right, but if I shared the flow to others, and I update the creational in Flowise, is that a case?

PR #1029, opened 2023-10-10
Text output for Text File and Plain Text components

jan-stanek:
Added possibility to select "Text output" instead of "Document" for components "Text File" and "Plain Text". It allows to use the output directly as a value of a variable in prompt.

Screenshot_20231010_164409

### DIFF
--- a/packages/components/nodes/documentloaders/PlainText/PlainText.ts
+++ b/packages/components/nodes/documentloaders/PlainText/PlainText.ts
@@ -18,7 +18,7 @@ class PlainText_DocumentLoaders implements INode {
     constructor() {
         this.label = 'Plain Text'
         this.name = 'plainText'
-        this.version = 1.0
+        this.version = 2.0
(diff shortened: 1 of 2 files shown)


packages/components/nodes/documentloaders/PlainText/PlainText.ts
```diff
@@ -12,6 +13,7 @@ class PlainText_DocumentLoaders implements INode {
     category: string
     baseClasses: string[]
     inputs: INodeParams[]
+    outputs: INodeOutputsValue[]
 
     constructor() {
         this.label = 'Plain Text'
```
HenryHengZJ, line 19: can we upgrade the `this.version=2.0` for both nodes?

jan-stanek: done

HenryHengZJ:
thanks @jan-stanek !

PR #1042, opened 2023-10-12
Remove unnecessary Langfuse inputs

marcklingen:
The optional inputs `flushAt`, `flushInterval` and `requestTimeout` only need to be changed in edge cases. I suggest to remove them from Flowise to simplify getting started with Langfuse.


HenryHengZJ (approved):
thanks @marcklingen !

Issue #911, opened 2023-09-12
Docker Image not running. Error Command failed with exit code 127


Zulelee:
### Discussed in https://github.com/FlowiseAI/Flowise/discussions/900

<div type='discussions-op-text'>

<sup>Originally posted by **Zulelee** September  9, 2023</sup>
I have been trying to create a docker image of Flowise. I am stuck on this task since 1 week and I can not find the solution. I have searched everything I can regarding it. 

The image was created successfully. But when I run the container with the image, I get the following error:

```
2023-09-09 21:05:08 yarn run v1.22.19
2023-09-09 21:05:08 $ run-script-os
2023-09-09 21:05:08 $ cd packages/server/bin && ./run start
2023-09-09 21:05:09 info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
': No such file or directory't execute 'node
2023-09-09 21:05:09 error Command failed with exit code 127.
2023-09-09 21:05:09 error Command failed with exit code 127.
2023-09-09 21:05:09 info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
``` 
Can anyone please guide me how to solve it?

</div>

automaton82:
I have the same issue. The problem appears to be in the `/server/bin/run` and `dev` commands. They have `^M` endings if you edit them in `vi` which I believe is Windows CRLF endings. If I remove them manually inside of the docker image and run the command manually, it appears to fix it. Let me attempt to fix the file and I could propose a pull request to only have LF in these files to solve this.

PR #1054, opened 2023-10-13
Fix for #911, CRLF line endings on Linux scripts.

automaton82:
As per https://github.com/FlowiseAI/Flowise/issues/911, the line endings on the Linux scripts are CRLF which do not work properly if a Docker build is created on Windows. To solve this, the line endings are converted to LF in source.


HenryHengZJ:
awesome thanks @automaton82 for the fix!

HenryHengZJ:
do you mind pulling latest changes from main branch to get the Node CI pass?

automaton82:
Sure I did a sync just now. Hopefully that solves it.

automaton82:
I think they passed now, so it seems to have worked.

HenryHengZJ (approved):

HenryHengZJ:
Render deployment failed: 
image

Looks like we need to revert this

automaton82:
Looks like a permissions issue on the file when in docker. But the permissions don't come from the host machine, so I'm not sure how it could be related. That's odd.

HenryHengZJ:
Yeah, merged the revert PR and it fixed the issue. Not sure what was the cause of it

automaton82:
Ok. I'll try to replicate. The way it is now won't docker run in Windows, so there must be a way to fix both.

automaton82:
@HenryHengZJ I think I can fix this by following https://stackoverflow.com/questions/30195151/can-i-commit-specific-files-in-git-without-affecting-the-original-line-ending and adding a `.gitattributes` file into that folder targeting only `run` and `dev` files to have `LF` line endings. This would mean no changes to the actual files.

I have this working locally in Windows, but I'm now going to test it on Ubuntu WSL2 Linux. If it passes both of these, are you okay if I submit another PR with this file added to solve this instead of the failed way?

automaton82:
Well I submitted a new PR #1077 which I tested to ensure it would run on Linux and Windows. It's a 1 file addition for git to deal with the line endings, and no changes to the actual files to avoid any permissions issues again.

Hopefully this is ok. It does fix it, as long as the file is touched by git (or re-cloned).

PR #1070, opened 2023-10-16
fix: ChatPromptTemplate function name

yavisht:
@deprecated — Renamed to .fromMessages

Screen Shot 2023-10-16 at 6 48 57 pm


HenryHengZJ (approved):
thanks @yavisht !

PR #1071, opened 2023-10-16
Add missing env variables for database to work in container

KeLeKaPe:
The docker/.env file database credentials and more where not included in the docker container image, and this would make sure the env variables where present in the container image. It now works for me, so wanted to share.


docker/docker-compose.yml
```diff
@@ -10,6 +10,12 @@ services:
             - FLOWISE_PASSWORD=${FLOWISE_PASSWORD}
             - DEBUG=${DEBUG}
             - DATABASE_PATH=${DATABASE_PATH}
+            - DATABASE_TYPE=${DATABASE_TYPE}
```
HenryHengZJ, line 13: the reason we leave this out from `docker-compose.yml` is because you can either use `DATABASE_PATH` or `DATABASE_PORT` and the rest of the env vars.

When using `DATABASE_PATH`, sqlite will be used. If you dont want to use sqlite, you can then use other env vars to configure your backend database like Postgres

KeLeKaPe: The issue I had, is that even if I have an .env file present in the docker/ folder, when the image is started with `docker-compose up -d` and I go into the container and run `printenv`, the DATAHUB_* variables are not there, so when trying to set up the MySQL credentials, it does nothing.
That is why I had to do this to get it to work.
(Am I doing something wrong then?)

HenryHengZJ: ah I see. You're doing right, we were just under the assumption that devs can add those env variables to `docker-compose.yml`. But no harm adding these in so

HenryHengZJ (approved):

PR #1076, opened 2023-10-16
Add Memory Node Input to Multi Prompt Chain

neoplastic:
Expose the memory node input inside the Multi Prompt Chain. 

I didn't know how to hide the * as I assume this is label that signifies this is required? But I've done a test to see if its required and it doesn't seem to be.


neoplastic:
Yeah... Looking through the Langchain documentation the llmChainOpts: LLMChainInput does have a memory option, however I had this working before (or seemed to) but its not now?

PR #1107, opened 2023-10-20
Feature/Postgres VS additional config

HenryHengZJ:
image


dkindlund:
So there was a problem with this fix, @HenryHengZJ.

Essentially, the `similaritySearchVectorWithScore` function wasn't getting passed the ssl parameters in order to actually connect to the Postgres DB using SSL.

Here's the fix that I think solves this problem:
https://github.com/FlowiseAI/Flowise/commit/07eb4cf8815cba8d814c2dedb6f11b8bd4b12ee5

Let me know if this makes sense to you, and I'll submit a PR for it accordingly.  Thanks!

Issue #1123, opened 2023-10-24
[BUG]TypeError: Cannot read properties of undefined (reading 'replace')


Ares0101:
2023-10-24 15:01:50 [ERROR]: Cannot read properties of undefined (reading 'replace')
TypeError: Cannot read properties of undefined (reading 'replace')
    at /home/huiyu/Flowise/node_modules/langchain/dist/embeddings/openai.cjs:142:92
    at Array.map (<anonymous>)
    at OpenAIEmbeddings.embedDocuments (/home/huiyu/Flowise/node_modules/langchain/dist/embeddings/openai.cjs:142:79)
    at MemoryVectorStore.addDocuments (/home/huiyu/Flowise/node_modules/langchain/dist/vectorstores/memory.cjs:41:54)
    at MemoryVectorStore.fromDocuments (/home/huiyu/Flowise/node_modules/langchain/dist/vectorstores/memory.cjs:129:24)
    at InMemoryVectorStore_VectorStores.init (/home/huiyu/Flowise/packages/components/dist/nodes/vectorstores/InMemory/InMemoryVectorStore.js:62:62)
    at buildLangchain (/home/huiyu/Flowise/packages/server/dist/utils/index.js:223:72)
    at async App.processPrediction (/home/huiyu/Flowise/packages/server/dist/index.js:822:40)
    at async /home/huiyu/Flowise/packages/server/dist/index.js:609:13
2023-10-24 15:01:50 [ERROR]: [server]: Error: TypeError: Cannot read properties of undefined (reading 'replace')
Error: TypeError: Cannot read properties of undefined (reading 'replace')
    at buildLangchain (/home/huiyu/Flowise/packages/server/dist/utils/index.js:235:19)
    at async App.processPrediction (/home/huiyu/Flowise/packages/server/dist/index.js:822:40)
    at async /home/huiyu/Flowise/packages/server/dist/index.js:609:13

chz8494:
I’ve seen this error before, actually very easy fix, it’s actually because of undefined pagecontent in document, only need one line, such as if doc.pagecontent !== undefined

Issue #1106, opened 2023-10-20
[BUG]TypeError: Cannot read properties of undefined (reading 'replace')


LiKe-rm:
**Describe the bug**
I select Local QnA in the marketplace, fill out the basepath and model name for ChatLocalAI and LocalAIembeeding, upload the txt file, save it and talk to it, and it reports an error TypeError: Cannot read properties of undefined ( read 'replace')

**To Reproduce**

1. Use Local QnA
2. fill out the basepath and model name for ChatLocalAI and LocalAIembeeding
3. upload the txt file
4. chat and you will see the erro

**Expected behavior**
normal conversation.

**Screenshots**
![image](https://github.com/FlowiseAI/Flowise/assets/72687314/0859c626-13a7-4f78-a9f8-d9493526c8b5)

**Additional context**
Couldn't you use flowise to load a local model?

jpcorso:
I'm having the same trouble here wile making a chat with a file

LordOfTheRats:
Same here

HenryHengZJ:
is this only happening for LocalAI?

LiKe-rm:
It seems that "Base Path" + "Module Name" should locate the model bin file, but what should I do if there are multiple bin files? For example "pytorch_model-00001-of-00002.bin", "pytorch_model-00002-of-00002.bin".

LiKe-rm:
> is this only happening for LocalAI?

It looks like this, when I replace ChatLocalAI, LocalAI Embedding with ChatOpenAI, OpenAI Embedding, it works fine

gboxxx:
I'm getting this error using OpenAI embeddings & ChatOpenAI - using the Conversational QnA Marketplace template (with Text File updated to version 2).

pinyhagt:
I'm having the same issue. Running Flowise on my local machine with OpenAI and Pinecone.

![image](https://github.com/FlowiseAI/Flowise/assets/147110829/3f78eacb-fa1a-44a8-9e03-3bd4210dbd67)

HenryHengZJ:
https://github.com/FlowiseAI/Flowise/pull/1127 this should fixed it! feel free to reopen if issue still persists

mario6464:
[...]

PR #1127, opened 2023-10-24
[Bug] Fix Cannot read properties of undefined (reading 'replace')

chz8494:
undefined document pagecontent is causing reading issue with replace function.
fix for:
#1106  #1123

### DIFF
--- a/packages/components/nodes/vectorstores/InMemory/InMemoryVectorStore.ts
+++ b/packages/components/nodes/vectorstores/InMemory/InMemoryVectorStore.ts
@@ -71,7 +71,7 @@ class InMemoryVectorStore_VectorStores implements INode {
         const flattenDocs = docs && docs.length ? flatten(docs) : []
         const finalDocs = []
         for (let i = 0; i < flattenDocs.length; i += 1) {
-            if (flattenDocs[i] !== undefined) {
+            if (flattenDocs[i] !== undefined && flattenDocs[i].pageContent !== undefined) {
                 finalDocs.push(new Document(flattenDocs[i]))
             }
         }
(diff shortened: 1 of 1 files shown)


chz8494:
@HenryHengZJ

HenryHengZJ:
thanks for the suggested fix! Can you let me know how to replicate the issue? I've been trying to use the Conversational Retrieval QA Chain template, but mine works everytime

chz8494:
For example, if you have an empty confluence page, the document results will end up with undefined pagecontent and some metadata. Also if you drag a confluence loader as document input then replace it with other loader, it also end up with undefined pagecontent with a confluence element in the document array.

HenryHengZJ:
@chz8494 do you have any other examples? I was not able to have Confluence setup correctly

packages/components/nodes/vectorstores/InMemory/InMemoryVectorStore.ts
```diff
@@ -71,7 +71,9 @@ class InMemoryVectorStore_VectorStores implements INode {
         const flattenDocs = docs && docs.length ? flatten(docs) : []
         const finalDocs = []
         for (let i = 0; i < flattenDocs.length; i += 1) {
-            finalDocs.push(new Document(flattenDocs[i]))
+            if (flattenDocs[i] !== undefined) {
+                finalDocs.push(new Document(flattenDocs[i]))
```
HenryHengZJ, line 75: yeah you are right. The error occurs when you have undefined `pageContent`. However this won't fix it because `flattenDocs[i]` is an object:
```
{
            pageContent: 'example',
            metadata: {
                url: 'pageUrl'
            }
        }
```

I think the correct fix is to check if the `pageContent` is undefined. To make it extra secure we can check `if (flattenDocs[i] !== undefined && flattenDocs[i].pageContent !== undefined)`

chz8494:
I’ll change that in my PR. My actual fix was supposed to be just "if (flattenDocs[i].pageContent !== undefined)" but had put it only flattenDocs[i] accidentally.

HenryHengZJ (approved):
thanks @chz8494 !

PR #1136, opened 2023-10-26
Build error fixed (TS)

siddiqss:
Error: "Property 'INDEX_NOT_EXIST' does not exist on type 'typeof ErrorCode'. Did you mean 'IndexNotExist'?ts(2551)
error.d.ts(39, 5): 'IndexNotExist' is declared here."
in "packages/components/nodes/vectorstores/Milvus/Milvus_Upsert.ts".

Fixed: Replaced INDEX_NOT_EXIST with IndexNotExist

issue described here: #1135

PR #1149, opened 2023-10-28
Fix  "'INDEX_NOT_EXIST' does not exist" error in Milvus_Upsert.ts

DevOtts:
When I tried to run `yarn build` I was getting this error. 
image

nodes/vectorstores/Milvus/Milvus_Upsert.ts(255,59): error TS2551: Property 'INDEX_NOT_EXIST' does not exist on type 'typeof ErrorCode'. Did you mean 'IndexNotExist'?

Probably just something that someone forgot while doing a refactor in the code.

After that change, build success:
image


HenryHengZJ:
this was fixed - https://github.com/FlowiseAI/Flowise/pull/1136

PR #1158, opened 2023-10-30
Add OpenAIEmbeddingCustom

tirongi:
Similarly to the OpenAIChatModelCustom node, 
this feature allow to select a customised modelName or the default one.


HenryHengZJ:
Can we add an extra field for Model Name on the existing `OpenAIEmbedding` node instead of creating a new custom node 
 Reason `ChatOpenAICustom` was created separately because the original `ChatOpenAI` node has different pre-coded dropdown options for different models, whereas `OpenAIEmbedding` has only 1 model, so there is no need for dropdown.

We can have the extra Model Name field added in, and when provided, it will be used: `if (modelName) obj.modelName= modelName`

tirongi:
> Can we add an extra field for Model Name on the existing `OpenAIEmbedding` node instead of creating a new custom node? Reason `ChatOpenAICustom` was created separately because the original `ChatOpenAI` node has different pre-coded dropdown options for different models, whereas `OpenAIEmbedding` has only 1 model, so there is no need for dropdown.
> 
> We can have the extra Model Name field added in, and when provided, it will be used: `if (modelName) obj.modelName= modelName`

Hi, thanks for your comment.

Hi did what you said in the beginning, but then I realised that would be better to use the same approach used in OpenAI chat model components for better extensibility, if eventually OpenAI in the future may release some other embedding model and we want to pre-code them in the "not-custom" model.

HenryHengZJ:
> > Can we add an extra field for Model Name on the existing `OpenAIEmbedding` node instead of creating a new custom node? Reason `ChatOpenAICustom` was created separately because the original `ChatOpenAI` node has different pre-coded dropdown options for different models, whereas `OpenAIEmbedding` has only 1 model, so there is no need for dropdown.
> > We can have the extra Model Name field added in, and when provided, it will be used: `if (modelName) obj.modelName= modelName`
> 
> Hi, thanks for your comment.
> 
> In the beginning I did what you said, but then I realised that would be better to use the same approach used in OpenAI chat model components for better extensibility beyond the reason of the analogy between chat and embedding components. I thought the case eventually OpenAI in the future may release some other embedding model and we want to pre-code them in the "not-custom" model, in this way we will fall in the same case as in the chat model.

yeah got it, make sense. Dont have a strong preference though

PR #1175, opened 2023-11-02
implementation for allowing prompts to be loaded from the langchain hub

vinodkiran:


vinodkiran:
image

Select your credential
enter a valid prompt in the text field, and click on lookup.
the prompt templates should be populated in the corresponding fields

PR #1176, opened 2023-11-02
Feature/Add query to vec2doc

HenryHengZJ:
Add ability to define a query to retrieve docs from vector store:
image


kientv:
Look like that the document is the same at every time when the query empty (I mean query is user question)

PR #1182, opened 2023-11-06
fix S3 loader, added region

muratlutfigoncu:
S3 File loader was failing due to "Missing region". Added a region key to s3 config.


HenryHengZJ:
Thanks for the suggested fix. But if you try to console log the `s3Config`, region is being passed correctly no?

PR #1191, opened 2023-11-08
Bug fix -- all remote packages should be referenced as local

automaton82:
Currently if you accidentally install the packages from npm, such as `npm install flowise-components`, you will be stuck as both `yarn` and `npm` will constantly use the public version instead of the local version.

This issue cost me several hours to diagnose when my build would successfully work, but no changes would be seen. It's especially dangerous during version changes of packages which is when you'll notice it happen the most.

To solve this, `package.json` supports [local dependencies](https://www.stefanjudis.com/today-i-learned/npm-install-supports-local-packages/) via the `file:` notation. Using this instead of `*` ensures the package comes from the local one, not any global one that happens to be around.


HenryHengZJ:
will have to test this against different installation method. Currently FW can be installed and run via:

- `npm install -g flowise` -> `npx flowise start`
- docker compose
- docker local image
- yarn

automaton82:
Ah yea that's fair, I didn't try that.

hrayr-artunyan:
@automaton82 I ran into the same issue today, here's how I resolved it:

1. Run `yarn link` in `packages/ui`
2. Run `yarn link` in `packages/components`
3. Run `yarn link flowise-ui flowise-components` in `packages/server`,

automaton82:
Thank you for sharing that information! I like this permanent fix more 😄 but I get there's a risk.

automaton82:
I'm closing this PR, it's not correct. As of today it fails for me because it copies both `flowise-components` and `flowise-ui` into `$/packages/server/node_modules` and those do not have their icons (among other problems).

If `*` is used instead this doesn't happen (they aren't copied), and if the problem arises then the above suggestion to use `yarn link` should suffice.

PR #1210, opened 2023-11-10
Adding ability to hide node param from UI

automaton82:
This feature adds the ability for node params to mark themselves as `hidden` from the UI.

This is required for nodes which don't have static API credentials, but rather have session context tokens or otherwise that are dependent and change when the chatbot is embedded, and should not be set in the UI.

These params can still be targeted and passed a value through `chatflowConfig` options. In addition, the `i` info button continues to show these parameters inside of `NodeInfoDialog` so that developers know they exist.

### DIFF
--- a/packages/components/src/Interface.ts
+++ b/packages/components/src/Interface.ts
@@ -72,6 +72,7 @@ export interface INodeParams {
     fileType?: string
     additionalParams?: boolean
     loadMethod?: string
+    hidden?: boolean
 }
 
 export interface INodeExecutionData {
--- a/packages/ui/src/views/canvas/CanvasNode.js
+++ b/packages/ui/src/views/canvas/CanvasNode.js
@@ -207,9 +207,11 @@ const CanvasNode = ({ data }) => {
                         {data.inputAnchors.map((inputAnchor, index) => (
                             <NodeInputHandler key={index} inputAnchor={inputAnchor} data={data} />
                         ))}
-                        {data.inputParams.map((inputParam, index) => (
-                            <NodeInputHandler key={index} inputParam={inputParam} data={data} />
-                        ))}
+                        {data.inputParams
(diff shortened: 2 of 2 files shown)


HenryHengZJ (approved):
thanks @automaton82 !

automaton82:
There is a node in development and unfortunately not publicly available at this time. However, I will share what I can to give you an idea of why it's required and how it works.

We have an Enterprise application that is able to provide data in multiple ways, and would like a node to help with this. This application has a complicated sign-on infrastructure including single sign-on and federated authentication for N users, however once a user has authenticated they're given a unique session (or token) ID. This ID has a lifetime of the session, but is never fixed.

All of the data provided by the node is always *in the context of the session*. If user A signs on and uses the chat, they get different data than user B despite using the exact same chatflow. This is because the session ID is different between them, and injected at embed-time.

This is accomplished in multiple ways. First, the parameter itself in the node is marked as hidden like this:

```
{
    label: 'Application Session ID',
    name: 'ourApplicationSessionId',
    type: 'string',
    description: 'The ID of the session that should be used for accessing data from the application.',
    placeholder: 'Session ID',
    optional: true,
    hidden: true
},
```

The user never sees this part in the UI, nor do they need to care about it. For the embedded case it's obvious: the example embed code in `EmbedChat` has been augmented with something like:

```
chatflowConfig: {
    ourApplicationSessionId: window.getApplicationSessionId()
}
```

This addresses the embedding. However, for the case where they're using Flowise directly it requires something like:

`incomingInput.overrideConfig.ourApplicationSessionId = ourSessionId`

The next phase could be to extend Flowise with N users instead of just 1, but that's a much bigger feature.

HenryHengZJ:
thanks for the detail explanation! another thing I can think of is ssomething like OAuth2 scopes, where you want to get the specified scopes from the node on UI but dont want to display it out on UI.

PR #1217, opened 2023-11-11
Removal of older models from OpenAI

Zochory:
https://platform.openai.com/docs/models/


HenryHengZJ:
think these models are marked as legacy, but still can be used. we can remove them once its completely removed from OpenAI

HenryHengZJ:
closing for now, will revisit once they are completely out, as some are still using these models

PR #1222, opened 2023-11-13
ResponsibleAI - Input Moderation - …

vinodkiran:
Ability to verify inputs and prevent potentially harmful content generation


vinodkiran:
image

bilalnawaz072 (approved):

packages/components/nodes/responsibleAI/InputModeration/SimplePromptModeration.ts
```diff
@@ -0,0 +1,56 @@
...
+        this.version = 1.0
+        this.type = 'Moderation'
+        this.icon = 'simple_moderation.png'
+        this.category = 'Responsible AI'
+        this.description = 'Detecting and mitigating prompt attacks'
+        this.baseClasses = [this.type, ...getBaseClasses(Moderation)]
+        this.inputs = [
+            {
+                label: 'Deny List',
+                name: 'denyList',
```
HenryHengZJ, line 29: is it better to be separated by comma? there might be cases where a string literal contains `\n`

packages/components/nodes/responsibleAI/ResponsibleAI.ts
```diff
@@ -0,0 +1,30 @@
...
+export const checkInputs = async (inputModerations: Moderation[], llm: BaseLanguageModel, input: string): Promise<string> => {
+    for (const moderation of inputModerations) {
+        input = await moderation.checkForViolations(llm, input)
+    }
+    return input
+}
+
+// is this the correct location for this function?
+// should we have a utils files that all node components can use?
+export const streamResponse = (isStreaming: any, response: string, socketIO: Server, socketIOClientId: string) => {
```
HenryHengZJ, line 19: for now we can keep this here as its still specific to this moderation implementation. we can move this out whenever we encounter the needs of this streaming in future

packages/components/nodes/chains/LLMChain/LLMChain.ts
```diff
@@ -144,14 +153,25 @@ const runPrediction = async (
...
      * TO: { "value": "hello i am ben\n\n\thow are you?" }
      */
     const promptValues = handleEscapeCharacters(promptValuesRaw, true)
 
+    if (moderations && moderations.length > 0) {
+        try {
+            // Use the output of the moderation chain as input for the LLM chain
+            input = await checkInputs(moderations, chain.llm, input)
+        } catch (e) {
+            await new Promise((resolve) => setTimeout(resolve, 500))
```
HenryHengZJ, line 169: adding slight delay because when the response return too fast, the order of the chat message returned became opposite:
Screenshot 2023-11-22 165835

packages/components/nodes/responsibleAI/ResponsibleAI.ts
```diff
@@ -0,0 +1,30 @@
+import { BaseLanguageModel } from 'langchain/base_language'
+import { Server } from 'socket.io'
+
+export abstract class ResponsibleAI {}
```
HenryHengZJ, line 4: @vinodkiran should we remove `ResponsibleAI` and rename to `Moderation` ? The original thinking was that we will be adding input & output moderation, but after discussion we will just add input moderation, so `ResponsibleAI` might not make sense anymore

vinodkiran: Agreed! I have updated the file and directory names to reflect better alignment to the functionality.

HenryHengZJ (approved):

PR #1243, opened 2023-11-16
Feature/add vector upsert ability

HenryHengZJ:
image

Currently Vector Store has Upsert and Load Existing Index nodes.

- Problem - when user use an upsert node, and changed some configs on the flow, and start asking question in chat, it causes the upsert to be done again.
 
- Solution - combine upsert and load existing into a node, and have a UPSERT button on UI to allow user to manually upsert, hence the node on flow will always load from existing index.


tirongi:
In any case upserting nodes should be kept, as they are useful in case you need to create a flow (and hence an API) for upserting documents from a specific folder.
Another solution could be that of having a flag in the node configuration allowing the user to specify if the node should upsert documents or use those already upserted.

HenryHengZJ:
@tirongi in this new update, you have a separate API for upsert: 
image

reason we will be deprecating these upserts nodes is because when there is a slight change on the flow, everything will be re-upsert again, and we want to avoid that, leaving user with the options to manually kick off the upsert

HenryHengZJ:
@vinodkiran everything should be addressed. Kindly take a look again

vinodkiran (approved):

vijaykammili:
> @tirongi in this new update, you have a separate API for upsert: image
> 
> reason we will be deprecating these upserts nodes is because when there is a slight change on the flow, everything will be re-upsert again, and we want to avoid that, leaving user with the options to manually kick off the upsert

@HenryHengZJ  - What is the process to get to this screen? I don't see this screen in the latest build. I see deprecating Vector stores and NEW Vector stores but can't see the screen where you have the upsert button.

Issue #1206, opened 2023-11-10
[BUG] Conversational Agent Unable To Determine What Tool To Use ?


DebajitKumarPhukan:
**Describe the bug**
I tried a Local LLM (LLama2) as the LLM backing the Conversational Agent. The Conversational Agent has 2 tools attached: Retrieval QA Chain Tool and a Serper. I asked for today's date to the agent for which it failed miserably and returned some old date. When I specifically asked to use Server then it managed to search the internet to find the date but failed in json parsing.

The above behavior is reproducible in any local LLM or hugging face LLM.

**To Reproduce**

1. Load the attached Chatflow.
2. Run `any local LLM` and use the API as base URL.
3. Set Up your free Serper account and configure the API key in the ChatFlow.
4. Ask `What is today's date ?`
5. See error

**Expected behavior**
1. The conversational agent shouldn't respond back with any "parse error".
2. There should be a provision to describe about the "Serper" tool so that we can provide the description to the conversational agent to better understand when to use.

**Screenshots**
![image](https://github.com/FlowiseAI/Flowise/assets/71944582/5194cb12-61b8-41d3-82ea-ba39549813c5)

**Flow**
If applicable, add exported flow in order to help replicating the problem.
[Local LLM Chatflow - Issues.json](https://github.com/FlowiseAI/Flowise/files/13318122/Local.LLM.Chatflow.-.Issues.json)

**Setup**
-   Installation [`yarn start`]
-   Flowise Version [1.4.1]
-   OS: [Windows]
-   Browser [chrome]

**Additional context**
Please find attached logs for your reference which will help you to find the areas when the error is thrown.
[Local LLM Error Logs.log](https://github.com/FlowiseAI/Flowise/files/13318167/Local.LLM.Error.Logs.log)

HenryHengZJ:
should be resolved after langchain version upgrade - https://github.com/FlowiseAI/Flowise/pull/1281

PR #1255, opened 2023-11-20
Resolves Langchain issues appearing in Flowise

DebajitKumarPhukan:
Anchoring initial Langchain version to the latest version as default. This commit will fix below issue(s).

https://github.com/FlowiseAI/Flowise/issues/1206


HenryHengZJ:
thanks, will do a regression test on the latest ver of langchain

PR #1257, opened 2023-11-20
Connect to elastic custom server using basic auth

tirongi:
I add a new Elastic Credential type that allows the Elastic connectors to connect to a custom Elastic instance using the basic auth.

### DIFF
--- a/packages/components/credentials/ElasticsearchURLUserPassword.credential.ts
+++ b/packages/components/credentials/ElasticsearchURLUserPassword.credential.ts
@@ -1,0 +1,36 @@
+import { INodeParams, INodeCredential } from '../src/Interface'
+
+class ElasticSearchURLUserPassword implements INodeCredential {
+    label: string
+    name: string
+    version: number
+    description: string
+    inputs: INodeParams[]
+
+    constructor() {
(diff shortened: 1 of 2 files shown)


packages/components/credentials/ElasticsearchURLUserPassword.credential.ts
```diff
@@ -0,0 +1,35 @@
...
+    inputs: INodeParams[]
+
+    constructor() {
+        this.label = 'ElasticSearch Custom URL User Password'
+        this.name = 'elasticSearchURLUserPassword'
+        this.version = 1.0
+        this.description = 'Allow to connect to a custom Elastic index'
+        this.inputs = [
+            {
+                label: 'Custom URL',
```
HenryHengZJ, line 17: I think we can just add this `customURL` to the existing `ElectricsearchUserPassword.credential.ts` as `type: optional` field

tirongi: In that case we need to set the `cloudId` as optional as well, since the user need to set only one of the two parameters and enter in the right statement in the `ElasticSearchBase.ts`.

PR #1260, opened 2023-11-21
UX Changes to the API Keys Screens

vinodkiran:
This PR updates the API Keys Screen to show the number of chatflows associated with each key and allows the user to view the flow details. The warning message is also updated to ensure that the user does not accidentally delete the key.

image


HenryHengZJ:
small update on ui: 
image

HenryHengZJ (approved):

PR #1262, opened 2023-11-21
Enable inserting custom URL  using basic auth

tirongi:
In the credential nodeId can be inserted a custom server URL and the basic auth will be used.

### DIFF
--- a/packages/components/credentials/ElectricsearchUserPassword.credential.ts
+++ b/packages/components/credentials/ElectricsearchUserPassword.credential.ts
@@ -11,8 +11,7 @@ class ElasticSearchUserPassword implements INodeCredential {
         this.label = 'ElasticSearch User Password'
         this.name = 'elasticSearchUserPassword'
         this.version = 1.0
-        this.description =
-        `Use Cloud ID field to enter your Elastic Cloud ID or the URL of the Elastic server instance.
+        this.description = `Use Cloud ID field to enter your Elastic Cloud ID or the URL of the Elastic server instance.
         Refer to <a target="_blank" href="https://www.elastic.co/guide/en/elasticsearch/reference/current/setting-up-authentication.html">official guide</a> on how to get User Password from ElasticSearch.`
         this.inputs = [
             {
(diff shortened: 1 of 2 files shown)


packages/components/credentials/ElectricsearchUserPassword.credential.ts
```diff
@@ -15,7 +15,7 @@ class ElasticSearchUserPassword implements INodeCredential {
             'Refer to <a target="_blank" href="https://www.elastic.co/guide/en/kibana/current/tutorial-secure-access-to-kibana.html">official guide</a> on how to get User Password from ElasticSearch'
         this.inputs = [
             {
-                label: 'Cloud ID',
+                label: 'Cloud ID or custom server URL',
                 name: 'cloudId',
```
HenryHengZJ, line 19: can we add a `description` to explain how to get Cloud ID OR what is custom server URL? Better if it has link to docs like [here](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/credentials/AWSCredential.credential.ts#L16)

packages/components/nodes/vectorstores/Elasticsearch/ElasticSearchBase.ts
```diff
@@ -144,15 +144,30 @@ export abstract class ElasticSearchBase {
...
             let password = getCredentialParam('password', credentialData, nodeData)
-            elasticSearchClientOptions = {
-                cloud: {
-                    id: cloudId
-                },
-                auth: {
-                    username: username,
-                    password: password
+            if (cloudId.startsWith('http')) {
+                let username = getCredentialParam('username', credentialData, nodeData)
```
HenryHengZJ, line 148: these 2 lines are redundant as its already declared above

tirongi:
Thanks for your suggestions.
I just added the explanation, you can use an Elastic Cloud ID or put the URL address to the Elastic server instance in which the index is stored.
Based on the given string the client will be configured with the right auth settings. 
See: [Elastic guide](https://www.elastic.co/guide/en/elasticsearch/client/javascript-api/current/client-connecting.html)

HenryHengZJ (approved):
thanks @tirongi !

PR #1272, opened 2023-11-23
Maintenance/pnpm vite jsx (Enhancements to the Project Infrastructure)

use-the-fork:
I have made several significant updates to the project's foundation, aimed at improving efficiency and standardization. Here's a breakdown of the changes:

1. **Switch from Yarn to PNPM:** This change brings multiple advantages:
    - **Efficient Package Management:** With PNPM, `node_modules` are symlinked and shared, leading to the reuse of 2,860 packages.
    image
    - **Elimination of Turbo:** Turbo is removed as a dependency because PNPM inherently supports parallel execution.

2. **Committing a Lock File:** A lock file has been committed to ensure consistent package versions across different setups.

3. **Introduction of `kill-port` in Root `package.json`:** This utility ensures that ports are terminated before running the `dev` and `start` scripts, preventing port conflicts.

4. **Standardization of File Extensions in UI Folder:** Renamed `.js` files to `.jsx` to adhere to standard practices, enhancing code readability and maintainability.

5. **Replacement of Craco with Vite:** This upgrade introduces several improvements:
    - **Enhanced Hot Module Replacement (HMR):** Vite provides faster HMR, making development more efficient.
    - **Improved Packaging:** Vite offers better packaging strategies, enhancing overall performance.

6. **Adding `@/` Alias for UI Directory:** This alias simplifies referencing within the UI directory, improving code readability and organization.

PR #1274, opened 2023-11-23
Bugfix/docker server

knighttower:
**Fixes**  
- Issues with docker server
- Removes duplicate "Dockerfiles"
- Addresses performance issues during dev (by using the new docker container, the app runs a lot faster and with less issues)
- Simplify the use with docker

**New Features**  
- quick scripts to use docker via npm scripts
- auto docker install
- avoids manual docker installation or weird setups
- auto yarn install into container
- scripts to start, stop and cli (enters the docker via bash interactive)
- optimizes the performance by using dockers features when using node modules and shared folders
- It auto copies the required files (env and docker-compose)
- Added quick shortcuts to the scripts so that:
  - `npm run docker:start `
    - installs, starts and runs the yarn stuff all in one
    - If everything is already installed, it will just start the docker and the flowise
  - `npm run docker:stop`  
    - Stops the correct container
  - `npm run docker:cli`  
    - enters the correct container in bash mode

**Usage and instructions**  
- Go to [Instructions](https://github.com/FlowiseAI/Flowise/blob/main/docker/README.md)


HenryHengZJ:
thanks for the PR @knighttower !

Few question: 
1. Is it still possible to use the docker-compose way to install Flowise via npm?
2. Is it still possible to use the root level Dockerfile to build & install Flowise locally?
3. Does the new method works across different deployments like Render, AWS, DigitalOcean, Azure, GCP - https://docs.flowiseai.com/deployment

knighttower:
Hello @HenryHengZJ ,

****Q-** "Is it still possible to use the docker-compose way to install Flowise via npm?"**
> **A-** Short answer, yes. Long answer, I added a new 'script' command to automate all that and avoid confusion and errors. This has been documented in the readme file (https://github.com/knighttower/Flowise/tree/bugfix/docker-server/docker). The new method is simple and fast, devs only have to run `npm run docker:start` and it will:

1. Copy the ENV if there is none. (it will skip if there is one present).
2. Run the docker-compose container.
3. Install Flowise (it will use the cache via Yarn if it has already been installed and skip the install if so).
4. Start the Flowiser server.

If anyone still wants or need to install extra stuff, they can use the `npm run docker:cli` or use standard docker method `docker exec -it bash` but using the scripts is preferred for simplicity and accuracy. 

****Q-** "Is it still possible to use the root level Dockerfile to build & install Flowise locally?"**
**A-** The container build is always locally to favor offline builds, customization and mostly easy maintenance of the "Dockerfile". I noticed a lot of discrepancies with the previous files and remote Docker image that it is very hard to know which one was the current one.
 

**Q- "Does the new method works across different deployments like Render, AWS, DigitalOcean, Azure, GCP - https://docs.flowiseai.com/deployment"**
**A-** Yes, in any service that allows the use of Docker, it will allow running the Dockerfile. With a Few exceptions where the Dockerfile is not used, like Render, Railway, Replit because they use their own Docker setup and not the Flowise image.

Note that once this PR is merged the docs there need to Updated. They seem to be outdated anyway, so this will be a good push to make updates happen at that end.

For reference: I updated the Readme files of this repo to reflect the new changes:
https://github.com/knighttower/Flowise/blob/bugfix/docker-server/README.md#-docker
https://github.com/knighttower/Flowise/tree/bugfix/docker-server/docker

Let me know if you have any other questions.

Thanks!

HenryHengZJ:
Reason we have 2 Dockerfiles, 1 for building image locally, and the other 1 for building via docker-compose using image from dockerhub, is because some cloud services dont support docker-compose like you mentioned Render. Will have to test thoroughly across different deployments to verify if they are still working the same. Plus, we will be merging in this [PR](https://github.com/FlowiseAI/Flowise/pull/1272) which introduces new infrastructure change, so might hold off on this PR for a while

PR #1302, opened 2023-11-28
Feature/add disable file download and regex to remove citation

HenryHengZJ:
From OpenAI Assistant Message, it contains redundant citations:
image

- Added regex to replace those lenticular brackets:
image

- Added Disable File Download to ignore file annotations


refractstud:
Thanks for the fix, I pulled the new image from docker compose and tested how ever I still see the issue come up sometimes, also file path of uploaded knowledge file shows up along with a download knowledge file button ? can this be fixed?
Screenshot 2023-11-29 at 7 50 28 pm

HenryHengZJ:
Just pushed the latest `flowise-components` release. If you try pull the latest image again, it should work

PR #1319, opened 2023-12-01
Feature/Prompt Fetch from Langsmith Hub

vinodkiran:


packages/server/src/index.ts
```diff
@@ -1036,6 +1039,55 @@ export class App {
...
+                if (!credential) return res.status(404).json({ error: `Credential ${req.body.credential} not found` })
+                // Decrypt credentialData
+                const decryptedCredentialData = await decryptCredentialData(credential.encryptedData, credential.credentialName, undefined)
+
+                const headers = {}
+                // @ts-ignore
+                headers['x-api-key'] = decryptedCredentialData.langsmithApiKey
+
+                const tags = req.body.tags ? `tags=${req.body.tags}` : ''
+                const url = `https://web.hub.langchain.com/repos/?${tags}offset=0&limit=20&has_commits=true&sort_field=num_likes&sort_direction=desc&is_archived=false`
```
HenryHengZJ, line 1080: do we need offset and limit here? on the UI we dont have pagination right?

packages/ui/src/ui-component/dialog/PromptLangsmithHubDialog.js
```diff
@@ -0,0 +1,475 @@
...
+            const createResp = await promptApi.getPrompt({
+                credential: credentialId,
+                promptName: selectedPrompt.full_name
+            })
+            if (createResp.data) {
+                prompt.detailed = createResp.data.templates
+            }
+        }
+        setSelectedPrompt(prompt)
+        await new Promise((resolve) => setTimeout(resolve, 500))
```
HenryHengZJ, line 163: why do we need a setTimeout here

HenryHengZJ:
Updated UI: 
image

image

packages/ui/src/ui-component/dialog/PromptLangsmithHubDialog.js
vinodkiran: Not needed, fixed and pushed changes

packages/server/src/index.ts
vinodkiran: Not needed for now, fixed and pushed changes

PR #1325, opened 2023-12-02
Fix for issue#1323 - delete duplicate chatflow

### LINKED ISSUE #1323: [BUG] Deleting a duplicated chatflow does not redirect after
**Describe the bug**
If you duplicate any chatflow, give it a name and save it. Then click "Settings" and "Delete", it will delete but you won't be redirected anymore. This works for chatflows if you open them directly (but not opened from duplicate)

**To Reproduce**

1. Open any chatflow
2. Click duplicate
3. In the duplicated chatflow, save it and give it a name
4. Click settings
5. Click delete

**Expected behavior**
Chatflow deleted and page redirected to chatflows like normal

automaton82:
This commit fixes #1323 that I found.

Previously after deleting it was calling `navigate('-1')`, but if it's duplicate it's in a new window with no history.

Here I changed it to `navigate('/')` which will have almost identical behaviour (in almost all cases they come from chatflows) and fixes the defect.

### DIFF
--- a/packages/ui/src/views/canvas/index.js
+++ b/packages/ui/src/views/canvas/index.js
@@ -170,7 +170,7 @@ const Canvas = () => {
             try {
                 await chatflowsApi.deleteChatflow(chatflow.id)
                 localStorage.removeItem(`${chatflow.id}_INTERNAL`)
-                navigate(-1)
+                navigate('/')
             } catch (error) {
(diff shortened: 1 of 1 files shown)


HenryHengZJ (approved):

HenryHengZJ:
thanks @automaton82 !

PR #1327, opened 2023-12-02
Bugfix/xss

vinodkiran:


packages/server/src/index.ts
```diff
@@ -210,9 +224,14 @@ export class App {
         })
 
         // Returns specific component node icon via name
-        this.app.get('/api/v1/node-icon/:name', (req: Request, res: Response) => {
-            if (Object.prototype.hasOwnProperty.call(this.nodesPool.componentNodes, req.params.name)) {
-                const nodeInstance = this.nodesPool.componentNodes[req.params.name]
+        this.app.get('/api/v1/node-icon/:name', param('name').notEmpty().escape(), (req: Request, res: Response) => {
+            const name = req.params.name
```
HenryHengZJ, line 228: for every validation, do we have to create another `const name`? can we just use the `req.params.name` so that when we remove this `notEmpty().escape()` check in future we have less work to do

packages/server/src/index.ts
```diff
@@ -402,58 +466,84 @@ export class App {
...
-                    const chatTypeFilterArray = JSON.parse(chatTypeFilter)
-                    if (chatTypeFilterArray.includes(chatType.EXTERNAL) && chatTypeFilterArray.includes(chatType.INTERNAL)) {
-                        chatTypeFilter = undefined
-                    } else if (chatTypeFilterArray.includes(chatType.EXTERNAL)) {
-                        chatTypeFilter = chatType.EXTERNAL
-                    } else if (chatTypeFilterArray.includes(chatType.INTERNAL)) {
-                        chatTypeFilter = chatType.INTERNAL
+        this.app.get(
+            '/api/v1/chatmessage/:id',
+            query('chatId').notEmpty().escape(),
```
HenryHengZJ, line 471: all these query are optional. by declaring this as `notEmpty()`, a basic API call to get chat messages without these query will fail

packages/server/src/index.ts
```diff
@@ -293,14 +326,19 @@ export class App {
                     .orderBy('cf.name', 'ASC')
                     .getMany()
                 if (chatflows.length >= 1) return res.status(200).send(chatflows)
-                return res.status(404).send('Chatflow not found')
+                return res.status(404).send('APIKey not found')
```
HenryHengZJ, line 329: Should be `Chatflow not found` right? we are getting chatflows from apikey

packages/server/src/index.ts
```diff
@@ -1032,7 +1181,11 @@ export class App {
             }
         )
 
-        this.app.post('/api/v1/vector/internal-upsert/:id', async (req: Request, res: Response) => {
+        this.app.post('/api/v1/vector/internal-upsert/:id', param('id').notEmpty().escape(), async (req: Request, res: Response) => {
```
HenryHengZJ, line 1184: The API above this `'/api/v1/vector/upsert/:id` also need to be updated

packages/server/src/index.ts
vinodkiran: I have changed this and removed validation at each function level.

packages/server/src/index.ts
vinodkiran: I have changed the entire implementation and removed validation for each param

HenryHengZJ:
I love this newer approach, much cleaner, but the library `strip-js` is deprecated, any other alternatives?

HenryHengZJ:
> I love this newer approach, much cleaner, but the library `strip-js` is deprecated, any other alternatives?

this looks like a better library and its up to date - https://github.com/apostrophecms/sanitize-html

HenryHengZJ (approved):

Issue #965, opened 2023-09-25
[BUG] Store username/password in localStorage not secure


kientv:
**Describe the bug**
Seem that username/password not secure when storing in localStorage. Because someone can inspect and copy it from browser. I think that secure cookie is better solution.

PR #1328, opened 2023-12-02
moved all user credentials to secure cookies from localStorage

vinodkiran:
fix for [Issue 965](https://github.com/FlowiseAI/Flowise/issues/965)

Moved all user credentials from localStorage to Secure Cookies

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -65,7 +65,7 @@ export class App {
     chatflowPool: ChatflowPool
     cachePool: CachePool
     AppDataSource = getDataSource()
-
+    authenticated = false
     constructor() {
         this.app = express()
     }
@@ -122,7 +122,11 @@ export class App {
             const username = process.env.FLOWISE_USERNAME
             const password = process.env.FLOWISE_PASSWORD
             const basicAuthMiddleware = basicAuth({
-                users: { [username]: password }
+                users: { [username]: password },
+                authorizer: (username: string, password: string) => {
+                    this.authenticated = username === process.env.FLOWISE_USERNAME && password === process.env.FLOWISE_PASSWORD
+                    return this.authenticated
+                }
             })
             const whitelistURLs = [
                 '/api/v1/verify/apikey/',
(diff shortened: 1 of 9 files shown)


vinodkiran:
@automaton82 See [my comment](https://github.com/FlowiseAI/Flowise/pull/1327#issuecomment-1837842194) from the other PR. The idea of this specific PR is to reduce the usage of localstorage from multiple code locations to a single function/piece of code and remove the password in clear text stored locally. Yes, it is still possible to decode the cookie, and we will prevent this going forward.

Subsequent PR will remove this functionality completely in favor Secure Http Only Cookies. I am breaking the overall implementation into a series of small steps to ensure we don't have any intended consequences (bugs).

packages/ui/src/api/cookies.js
```diff
@@ -0,0 +1,45 @@
...
+
+export function getUsername() {
+    if (!localStorage.getItem('loggedIn')) {
+        return undefined
+    }
+    return getCookieValue('flowise-user', document.cookie)
+}
+
+export function removeCookies() {
+    document.cookie = 'cookiename= ; expires = Thu, 01 Jan 1970 00:00:00 GMT'
```
HenryHengZJ, line 43: why do we nedd this?

HenryHengZJ:
when I change the `FLOWISE_PASSWORD` or `FLOWISE_USERNAME` and restart the app, then login with the new credential, at first it loads fine. But when refresh page, its having 401

packages/server/src/index.ts
```diff
@@ -122,7 +122,11 @@ export class App {
             const username = process.env.FLOWISE_USERNAME
             const password = process.env.FLOWISE_PASSWORD
             const basicAuthMiddleware = basicAuth({
-                users: { [username]: password }
+                users: { [username]: password },
+                authorizer: (username: string, password: string) => {
+                    this.authenticated = username === process.env.FLOWISE_USERNAME && password === process.env.FLOWISE_PASSWORD
```
HenryHengZJ, line 127: not recommend to do `===` comparsion, use `safeCompare` - https://www.npmjs.com/package/express-basic-auth#custom-authorization

HenryHengZJ:
> when I change the `FLOWISE_PASSWORD` or `FLOWISE_USERNAME` and restart the app, then login with the new credential, at first it loads fine. But when refresh page, its having 401

as we will be moving to JWT approach, is it better to just start with the proper JWT authentication instead of storing basic auth to cookies, then having to migrate to JWT in future.

vinodkiran:
@HenryHengZJ  On second thoughts, I would recommend that we drop this PR and go straight to JWT at a later time. This implementation is too flaky and there are too many edge cases that can cause issues.

Issue #615, opened 2023-07-25
[BUG] OpenAI token exceeded error


atishhamte:
I am using the flowise configuration like the image below
![Screenshot 2023-07-25 at 8 37 24 PM](https://github.com/FlowiseAI/Flowise/assets/8375139/d8de8cd3-4ae6-4638-9057-2535db8d76b1)

Even after sending 1 line message like "Hello, how are you?" it is giving error like,

[ERROR]: [server]: Error: Request failed with status code 400 and body {"error":{"message":"This model's maximum context length is 4097 tokens. However, your messages resulted in 5029 tokens. Please reduce the length of the messages.","type":"invalid_request_error","param":"messages","code":"context_length_exceeded"}}

I think the conversation chain Is doing some things weird.

HenryHengZJ:
Try delete the flow, and use the template "Simple Conversationa Chain" from marketplace.
It works fine:
![image](https://github.com/FlowiseAI/Flowise/assets/26460777/201e46c6-6212-409b-9868-c8e3cafb759b)

matheusfterra:
Hi, same error in here. I Try get a "Simple Conversationa Chain" from marketplace too, but not working for me.

`Request failed with status code 400 and body {"error":{"message":"This model's maximum context length is 4097 tokens. However, your messages resulted in 9663 tokens. Please reduce the length of the messages.","type":"invalid_request_error","param":"messages","code":"context_length_exceeded"}}

`

Javokhir-1999:
Hi, I the same problem, it was working for some time after deployment but know I am getting this error.

atishamte:
@matheusfterra @Javokhir-1999 one alternate hack I used here to mitigate this situation is using a random value for memorykey parameter of buffer memory.
this actually solved my issue but this is a HACK

pellicceama:
I still get this issue both on Buffer Memory and on Dynamo DB Memory. 

Hi @atishamte , can you share how you passed in the custom memory key parameter?

HenryHengZJ:
or we can use a higher context model like gpt3.5-16k

atishamte:
@pellicceama I am using API to call these as we are not using flowise UI for the operation. So everytime I call the API, we are using UUID as a random key

pellicceama:
Me too! I tried this without much success. 

overrideConfig: {..., inputKey: `${randomUUID}`}} without 

How do you currently pass the UUID as a random key?

atishamte:
[...]

PR #1379, opened 2023-12-13
Add windowSize input to avoid "token exceeded error"

kzhang-dsg:
See: https://github.com/FlowiseAI/Flowise/issues/615

This is resolve the "token exceeded error" for redis backed memory.


HenryHengZJ (approved):
thanks @kzhang-dsg !

PR #1381, opened 2023-12-13
[BUGFIX/REGRESSION] Fix wrong elastic client setup when use custom URL

tirongi:
In the new Elastic node, there is the same connection error that was solved in the deprecated nodes whenever connecting to a custom server URL. 
The bug was previously solved in #1262


HenryHengZJ:
ah my mistake, forgot to move to the new node. thanks @tirongi !

HenryHengZJ (approved):

PR #1383, opened 2023-12-13
FEATURE/Environment Variables

vinodkiran:


packages/server/src/utils/index.ts
```diff
@@ -952,3 +954,29 @@ export const getAllValuesFromJson = (obj: any): any[] => {
     extractValues(obj)
     return values
 }
+
+export const replaceEnvVariables = async (question: string, appDataSource: DataSource): Promise<string> => {
```
HenryHengZJ, line 958: I think we can delete this right? It was a misunderstanding to have the env variables enabled in question

HenryHengZJ: actually hold on, we can reuse this

HenryHengZJ:
I think the runtime and static are kinda confusing for me. What if we get rid of the types and just:
UI:
- ability to add a new env variable with a value

API:
- if `envVars` is provided in body, then override it

Unless there are some cases where you need runtime and static 2 different types to achieve?

HenryHengZJ:
FYI, I've added a pair of setter/getter variable node in this [PR](https://github.com/FlowiseAI/Flowise/pull/1397/files#diff-0511388857b5d74918a4e7e9d4b83344352428c41ee6602063dfd86b195f6dd6)

packages/components/nodes/tools/CustomTool/CustomTool.ts
```diff
@@ -80,7 +80,32 @@ class CustomTool_Tools implements INode {
                 code: tool.func
             }
             if (customToolFunc) obj.code = customToolFunc
-            return new DynamicStructuredTool(obj)
+
+            const variables = await appDataSource.getRepository(databaseEntities['Variable']).find()
+
+            // override variables defined in overrideConfig
+            // nodeData.inputs.variables is an Object, check each property and override the variable
+            if (nodeData?.inputs?.envVars) {
```
HenryHengZJ, line 88: where do we pass the `envVars` ?

HenryHengZJ: got it. from the `overrideConfig`:

```
{
    "question": "hello",
    "overrideConfig": {
        "envVars": {
            "var": "123"
        }
    }
}
```

HenryHengZJ:
PR moved to [here](https://github.com/FlowiseAI/Flowise/pull/1422)

PR #1421, opened 2023-12-21
UI/Updated logos for Chatmodels, Embeddings and LLMs

abhishekshankr:
Updating icons for chat models, embeddings and LLM categories to keep them consistent with other icons in the UI. 

image


HenryHengZJ:
many thanks for the PR! but im just concern that in future when we add more integrations, every icons then will have to follow this pattern, with some that dont have the skill to do that (including me lol), it might add another layer of dependency to that.

abhishekshankr:
I plan on supporting this project well into the future! But I can make a guide for the icon set and how it was built, so if someone feels like they would like to edit the set, they can. It is quite close to the style of the Phosphor icon set as a fallback https://github.com/phosphor-icons/homepage

PR #1426, opened 2023-12-22
FEATURE Addition : MMR and Document Compressors

vinodkiran:
# 1) Support for MMR Search
Maximal marginal relevance search, which takes a combination of documents that are most similar to the inputs, then reranks and optimizes for diversity.

Enabled for Vector Store Nodes _(in the additional parameters dialog)_
- Pinecone
- Zep
- Weaviate
- Supabase

image
image

> [!NOTE]
> The default search is 'Similarity Search' which is the current behavior of these nodes

---------
# 2) Document Compressors as Retrievers
Challenge with retrieval is that usually we don't know the specific queries the document storage system will face when we ingest data into the system. This means that the information most relevant to a query may be buried in a document with a lot of irrelevant text. Passing that full document through the application can lead to more expensive LLM calls and poorer responses.

## Introducing 4 new Nodes under the Retriever Category
image

### a) Cohere Rerank Retriever
Leverages the Cohere rerank endpoint to rerank the returned results. Will add a relevance score and reindex the documents from most to least semantically relevant to the query.

### b) LLM Filter Retriever
This retriever will iterate over the initially returned documents and extract, from each, only the content that is relevant to the query.

### c) Embeddings Filter Retriever
Making an extra LLM call over each retrieved document is expensive and slow. The EmbeddingsFilter provides a cheaper and faster option by embedding the documents and query and only returning those documents which have sufficiently similar embeddings to the query.

### d) Reciprocal Rank Fusion Retriever
Reciprocal Rank Fusion (RRF) is a technique for combining the ranks of multiple search result lists to produce a single, unified ranking. 
-Based on a user question, multiple sub queries are generated
-For each of those sub queries, fetch documents from vector store
-Merge the retrieved documents together using Reciprocal Rank Fusion


HenryHengZJ:
@vinodkiran let me know once its ready for review!

vinodkiran:
> @vinodkiran let me know once its ready for review!

YES! ready for review

packages/components/nodes/retrievers/RRFRetriever/ReciprocalRankFusion.ts
```diff
@@ -0,0 +1,96 @@
+import { BaseDocumentCompressor } from 'langchain/retrievers/document_compressors'
```
HenryHengZJ, line 1: is this own implementation? or is there any docs link to this?

packages/components/nodes/retrievers/CohereRerankRetriever/CohereRerank.ts
```diff
@@ -0,0 +1,53 @@
...
+            headers: {
+                Authorization: `Bearer ${this.cohereAPIKey}`,
+                'Content-Type': 'application/json',
+                Accept: 'application/json'
+            }
+        }
+        const data = {
+            model: this.model,
+            topN: this.k,
+            max_chunks_per_doc: 10,
```
HenryHengZJ, line 35: can we make this as optional field?

vinodkiran: Done! Pushed as a new commit.

HenryHengZJ:
Updated retrievers to have multiple outputs and option to take in custom input: 
image

HenryHengZJ (approved):

PR #1439, opened 2023-12-28
Feature/LLM Moderation

HenryHengZJ:
added LLM to moderation to detect similar deny sentences

### DIFF
--- a/packages/components/nodes/moderation/SimplePromptModeration/SimplePromptModerationRunner.ts
+++ b/packages/components/nodes/moderation/SimplePromptModeration/SimplePromptModerationRunner.ts
@@ -19,6 +19,7 @@ export class SimplePromptModerationRunner implements Moderation {
         if (this.model) {
             const denyArray = this.denyList.split('\n')
             for (const denyStr of denyArray) {
+                if (!denyStr || denyStr === '') continue
                 const res = await this.model.invoke(
                     `Are these two sentences similar to each other? Only return Yes or No.\nFirst sentence: ${input}\nSecond sentence: ${denyStr}`
                 )
--- a/packages/components/nodes/moderation/SimplePromptModeration/SimplePromptModerationRunner.ts
+++ b/packages/components/nodes/moderation/SimplePromptModeration/SimplePromptModerationRunner.ts
@@ -1,23 +1,38 @@
 import { Moderation } from '../Moderation'
+import { BaseChatModel } from 'langchain/chat_models/base'
 
 export class SimplePromptModerationRunner implements Moderation {
     private readonly denyList: string = ''
     private readonly moderationErrorMessage: string = ''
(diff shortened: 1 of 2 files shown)


HenryHengZJ:
image

packages/components/nodes/moderation/SimplePromptModeration/SimplePromptModerationRunner.ts
```diff
@@ -1,23 +1,38 @@
...
     }
 
     async checkForViolations(input: string): Promise<string> {
-        this.denyList.split('\n').forEach((denyListItem) => {
-            if (denyListItem && denyListItem !== '' && input.toLowerCase().includes(denyListItem.toLowerCase())) {
-                throw Error(this.moderationErrorMessage)
+        if (this.model) {
+            const denyArray = this.denyList.split('\n')
+            for (const denyStr of denyArray) {
+                const res = await this.model.invoke(
```
vinodkiran, line 22: denyString can be empty string, if there are trailing \n

packages/components/nodes/moderation/SimplePromptModeration/SimplePromptModerationRunner.ts
```diff
@@ -1,23 +1,38 @@
...
+                    `Are these two sentences similar to each other? Only return Yes or No.\nFirst sentence: ${input}\nSecond sentence: ${denyStr}`
```
vinodkiran, line 23: can we not send the entire denyList and make one LLM call with an adjusted prompt?

packages/components/nodes/moderation/SimplePromptModeration/SimplePromptModerationRunner.ts
HenryHengZJ: updated check

packages/components/nodes/moderation/SimplePromptModeration/SimplePromptModerationRunner.ts
HenryHengZJ: we can add an option to let user select if they want to make one LLM call per deny string, or all together at once

HenryHengZJ: been trying to make a call like this: 
`const res = await this.model.invoke(
                    `Is this sentence: "${input}" similar to any of the sentences below? Only return Yes or No.\nSentences: ${this.denyList}`
                )`
However it always return false..

HenryHengZJ: will let it run one LLM call per deny for now, as i cant get consistent performance when using the whole list

HenryHengZJ:
@vinodkiran if you can take a final review again

PR #1462, opened 2024-01-01
added support for MMR

ofermend:
@HenryHengZJ 
Please see updates to the Vectara component with support for MMR reranking
Plus some minor updates


HenryHengZJ:
@ofermend if you can pull latest changes from main, or enable access to the branch so I can fix any merge conflict

ofermend:
Latest from main merged into the branch/PR. Can you try now?

packages/components/nodes/chains/VectaraChain/VectaraChain.ts
```diff
@@ -247,17 +248,30 @@ class VectaraChain_Chains implements INode {
             lexicalInterpolationConfig: { lambda: vectaraFilter?.lambda ?? 0.025 }
         }))
 
+        const mmrRerankerId = 272725718 // Vectara reranker ID for MMR
```
HenryHengZJ, line 251: can we add a comment above to the docs link - https://docs.vectara.com/docs/api-reference/search-apis/reranking, as in why the hardcoded value

packages/components/nodes/vectorstores/Vectara/Vectara.ts
```diff
@@ -82,16 +82,36 @@ class Vectara_VectorStores implements INode {
...
+                description: 'Number of top results to fetch for MMR. Defaults to 50',
+                placeholder: '50',
+                type: 'number',
+                additionalParams: true,
+                optional: true
+            },
+            {
+                label: 'MMR diversity bias',
+                name: 'mmrDiversityBias',
+                description: 'The diversity bias to use for MMR. Defaults to 0 (MMR disabled)',
```
HenryHengZJ, line 113: maybe we can add more details like:
```
The diversity bias to use for MMR, ranging between 0.0 and 1.0. Values closer to 1.0 optimize for the most diverse results. Defaults to 0 (MMR disabled)
```
and specify `step: 0.1`

ofermend: Added. 
note however there is no "step" - this value can be anything b/w 0.0 and 1.0

HenryHengZJ: `step` is allow the UI to increment by how many step, by default its 1. So when clicking +, it will increment by 1. By specifying `0.1` we allow smaller increment:
image

ofermend: Oh I see - that makes sense. Added.

HenryHengZJ:
looks good! @ofermend you might want to update the marketplace template

HenryHengZJ:
if you can do `yarn lint-fix` that'd be great

HenryHengZJ (approved):

PR #1480, opened 2024-01-04
Feature/qdrant port fix

tuxBurner:
Hello :)

We have our qdrant running behind an ingress which is doing https and listens on port 443.

The problem is that the qdrant-client.js pareses the qdrant server url with new Url(url).

As mentioned in https://stackoverflow.com/questions/59104197/nodejs-new-url-urlhttps-myurl-com80-lists-the-port-as-empty giving 443 or 80 as port results in an empty port in the url object. This leads the qdrant-client to default back to 6663.

So for example using a qdrant server url like: https://mysuperserver:443 results into https://mysuperserver:6663.

### DIFF
--- a/packages/components/nodes/vectorstores/Qdrant/Qdrant.ts
+++ b/packages/components/nodes/vectorstores/Qdrant/Qdrant.ts
@@ -268,7 +268,6 @@ class Qdrant_VectorStores implements INode {
             port = 80
         }
 
-
         return port
     }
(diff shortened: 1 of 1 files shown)


HenryHengZJ (approved):
thanks for the fix @tuxBurner !

tuxBurner:
Sure thing @HenryHengZJ

Issue #1099, opened 2023-10-19
[FEATURE] Localai Authorization key


Dbone29:
**Describe the feature you'd like**
You can set one or more API_KEY= in the .env in localai. If this is set, the localai server requires an Authorisation: Bearer Token and thus no longer allows requests that do not have this header. It would be nice to be able to specify this token in flowise.

**Additional context**
This function exists since localai v1.24.0 and must be entered manually in the .env.

avin-kavish:
@HenryHengZJ  I can try this. Would it work if I implement it as a input to the nodes or should I have it as an `.env` setting?

kkacsh321:
I worked this out locally, and put up a PR here: https://github.com/FlowiseAI/Flowise/pull/1495 that should satisfy this with any luck.

PR #1495, opened 2024-01-07
FEATURE: LocalAI Credential, Optional LocalAI Auth, Model list env var

kkacsh321:
This PR enables an ability to store a LocalAI based API Key as a credential, then OPTIONALLY use that credential in the LocalAI Chat Model.

This is for those of us hosting LocalAI but have authorization turned on with api keys.

This mimics the code already in place for OpenAI API.

I believe this should address https://github.com/FlowiseAI/Flowise/issues/1099 in full also 😉 

Screenshot 2024-01-06 at 5 25 12 PM
Screenshot 2024-01-06 at 5 23 23 PM

No api key set: (LocalAI server set to not use api_key)
Screenshot 2024-01-08 at 5 49 11 PM

Credential api key set:  (LocalAI server set to use api_key)
Screenshot 2024-01-08 at 5 44 34 PM

### DIFF
--- a/packages/components/nodes/chatmodels/ChatLocalAI/ChatLocalAI.ts
+++ b/packages/components/nodes/chatmodels/ChatLocalAI/ChatLocalAI.ts
@@ -1,5 +1,5 @@
-import { INode, INodeData, INodeParams } from '../../../src/Interface'
-import { getBaseClasses } from '../../../src/utils'
+import { ICommonObject, INode, INodeData, INodeParams } from '../../../src/Interface'
+import { getBaseClasses, getCredentialData, getCredentialParam } from '../../../src/utils'
 import { OpenAIChat } from 'langchain/llms/openai'
 import { OpenAIChatInput } from 'langchain/chat_models/openai'
 import { BaseCache } from 'langchain/schema'
@@ -14,6 +14,7 @@ class ChatLocalAI_ChatModels implements INode {
     category: string
     description: string
     baseClasses: string[]
+    credential: INodeParams
     inputs: INodeParams[]
 
     constructor() {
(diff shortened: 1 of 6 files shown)


packages/components/nodes/chatmodels/ChatLocalAI/ChatLocalAI.ts
```diff
@@ -41,8 +52,10 @@ class ChatLocalAI_ChatModels implements INode {
             {
                 label: 'Model Name',
                 name: 'modelName',
-                type: 'string',
-                placeholder: 'gpt4all-lora-quantized.bin'
+                type: 'options',
```
HenryHengZJ, line 55: Thanks for the suggestion! We didn't use `options` because we didn't want to limit the selections, hence added as `string` so that user can just specify the model they have in LocalAI.

Other users might face confusion as in why there is no selection available unless they read the repo carefully and realize they have to specify `LOCALAI_CHAT_MODELS` env. For now, I'd prefer to switch back to `string`. In future, we will have community nodes which allow users to create their custom component - https://github.com/FlowiseAI/Flowise/pull/1066

kkacsh321: @HenryHengZJ Understood completely 👍🏻 , If I remove that and put it back to string are you ok with the way I implemented the auth with a credential? I'll fix up and remove that env var/option after work today when I get a chance.

HenryHengZJ: yep everything else looks good!

HenryHengZJ: and also you can fix the linting issues `yarn lint-fix`

kkacsh321: doh 🤦🏻 -  @HenryHengZJ ok I reverted what was discussed, and in my re-testing also figured out I needed to take a couple steps back and add back defaults for people not using keys. I think this should work with minimal risk/change to current users unless they want to turn on api auth on their LocalAI. 🤞🏻 Let me know if I need to change anything else or haven't followed your processes (apologies).

HenryHengZJ (approved):

packages/components/nodes/chatmodels/ChatLocalAI/ChatLocalAI.ts
HenryHengZJ: awesome that all looks good! thanks!

kkacsh321:
@HenryHengZJ  fixed some typos, and handling for embedding auth with LocalAI also. Sorry was working on it when you approved.

HenryHengZJ:
> @HenryHengZJ fixed some typos, and handling for embedding auth with LocalAI also. Sorry was working on it when you approved.

awesome! thanks for the embedding part

PR #1497, opened 2024-01-08
Bugfix/Concurrent Chat Session

HenryHengZJ:
Bug: when concurrent users are chatting with a flow, memory is mixed up:

Steps to reproduce:

1.) Have a simple conversation chain and memory with sessionId:

image

2.) Open the chat, and introduce your name as “Evan”

image

3.) Open the Share Chatbot (to imitate another session), and introduce yourself as “Henry”

image

4.) Now, here’s the importance step. Prepare the question in the chatbox first:

**Henry’s session**
image

**Evan’s  session**
image

Then from the Share Chatbot (Henry’s session) page, quickly press enter to send

Switch to UI tab (Evan’s session), enter again to send

We are trying to simulate concurrent request

5.) Go to Redis Insight, observe the session:

**Henry’s session**
image

**Evan’s session:**
image

as you can see, Henry’s session messages got saved into Evan’s. Why? Because UI (Evan’s) is where I last entered “whats my name”


HenryHengZJ:
Fix: Change agent/chain to use RunnableSequence, this allow us to retrieve memory during runtime when `.run()` is called

PR #1505, opened 2024-01-09
Avoid accidentally deleting components by pressing the Backspace key …

mokeyish:
Avoid accidentally deleting components by pressing the Backspace key when text input loses focus.

### DIFF
--- a/packages/ui/src/views/canvas/index.js
+++ b/packages/ui/src/views/canvas/index.js
@@ -525,7 +525,7 @@ const Canvas = () => {
                                 onConnect={onConnect}
                                 onInit={setReactFlowInstance}
                                 fitView
-                                deleteKeyCode={canvas.canvasDialogShow ? null : ['Backspace', 'Delete']}
+                                deleteKeyCode={canvas.canvasDialogShow ? null : ['Delete']}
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
thanks for the suggested changes! But I have doubt that mac users gonna like it, cus there isnt delete button on mac, and most are relying on the backspace. A better solution is to do a better check when text input loses focus

TheArthurAbbott:
I like that there is a shortcut for deleting elements, but it is way too unintentional no matter what key is used (unless it is a weird one like the home button or weird combo like alt+;). The unfocusing while deleting is killer to productivity, and I can't imagine if you added up all the time lost not having the shortcut would be more than having it. So is there a way to fix the unfocusing so people aren't trying to remove shortcuts as a quick fix?

automaton82:
Somewhat unrelated, but using Ctrl+A to select "all the text" in an input will often select all the text on the entire page, with no good way to unselect it. It's rather frustrating.

But for this particular topic I have definitely accidentally deleted nodes with backspace quite a few times.

HenryHengZJ:
moving to here - https://github.com/FlowiseAI/Flowise/pull/1983

PR #1508, opened 2024-01-10
Add milvusTextField configuration for Milvus

mokeyish:
langchain python use `text` 
https://github.com/langchain-ai/langchain/blob/master/libs/community/langchain_community/vectorstores/milvus.py#L119

while langchian js use `langchain_text` 
https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-community/src/vectorstores/milvus.ts#L61

so it is necessary to add milvusTextField configuration for Milvus.


HenryHengZJ (approved):
Thanks @mokeyish !

PR #1509, opened 2024-01-10
Fix OpenAIFunctionAgent that function not return string result

mokeyish:
refer to https://github.com/langchain-ai/langchain/blob/master/libs/langchain/langchain/agents/format_scratchpad/openai_functions.py#L29

and fix the role of systemMessage from `ai` to `system`.


HenryHengZJ (approved):
thanks @mokeyish !

PR #1513, opened 2024-01-11
FEATURE: Sticky Note node

0xi4o:
- Adds a "Sticky Note" node under "Utilities". This node can be used to add notes to a Chatflow to explain what a node or a group of nodes does, explain input and output parameters, etc.

**Update**: Made the node look like a sticky note to visually distinguish it from other nodes. I think this will reduce visual clutter when you have a lot of nodes and help avoid confusion.

Here's how it looks:
Flowise-Sticky Notes


packages/ui/src/views/canvas/LightTooltip.js
```diff
@@ -0,0 +1,12 @@
+import { styled } from '@mui/material/styles'
```
HenryHengZJ, line 1: Do u think it make sense to refractor this into `ui-component/tooltip` and maybe rename as smtg that is specific like `NodeTooltip` as its only used for `CanvasNode` right?

HenryHengZJ:
We need to also update files under `ui/views/marketplaces`. With sticky note, it appears like this:
image

packages/ui/src/views/canvas/LightTooltip.js
0xi4o: I've moved the component to `ui-component/tooltip` and renamed it to `NodeTooltip`.

packages/ui/src/views/canvas/CardWrapper.js
```diff
@@ -0,0 +1,21 @@
+// material-ui
```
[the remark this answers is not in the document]
0xi4o: I've moved it to `ui-component/cards` and renamed it to `NodeCardWrapper`.

0xi4o:
> We need to also update files under `ui/views/marketplaces`. With sticky note, it appears like this: image

Updated the files under `marketplaces` so it renders sticky notes properly :+1:

HenryHengZJ (approved):
awesome thanks @0xi4o !

maximometascript:
I'm seeing this:
image

Kinda wishing I didn't delete my old information flow in favor of these. I didn't know I was going to brick them by closing the flow.

0xi4o:
@maximometascript can you open an issue with more details? I have the latest changes and I can't seem to replicate this.

PR #1539, opened 2024-01-16
Fix CustomFunction receiving excaped inputs

mokeyish:
When the input contains FLOWISE_DOUBLE_QUOTE, it will not be parsed using JSON.parse.


HenryHengZJ:
thanks @mokeyish !

PR #1566, opened 2024-01-19
FEATURE: Select which links should be used in web scraper nodes (cheerio, puppeteer, and playwright)

0xi4o:
- Adds a `Manage Links` button to Web Scraper nodes like Cheerio, Puppeteer, and Playwright. Until now, users can't choose which links to scrap. The node just picks the first `n` links (configurable in `Additional Parameters`) and scraps their content. Now users can pick which links they want to scrap. 
- Clicking on the `Manage Links` button will open a dialog where you can fetch all the links from the given URL, and remove the ones you don't need. During upsert, only the selected links will be scraped. Selecting links is optional. The nodes will fallback to the old behavior when no links are selected.
- If `Relative Links Method` is set in `Additional Parameters` before selecting links, the manage links dialog will use the selected method to fetch the links.
- If `Relative Links Limit` is set and is smaller than the number of selected links, the limit will apply to the selected links and the first `n` links are chosen. If it's larger, then all the selected links are scraped.

Flowise-webscraper-manage-links

Flowise-webscraper-manage-links-dialog


packages/components/nodes/documentloaders/Cheerio/Cheerio.ts
```diff
@@ -127,13 +128,22 @@ class Cheerio_DocumentLoaders implements INode {
             if (!limit) limit = '10'
             else if (parseInt(limit) < 0) throw new Error('Limit cannot be less than 0')
             const pages: string[] =
-                relativeLinksMethod === 'webCrawl' ? await webCrawl(url, parseInt(limit)) : await xmlScrape(url, parseInt(limit))
+                selectedLinks && selectedLinks.length > 0
+                    ? selectedLinks.slice(0, parseInt(limit))
+                    : relativeLinksMethod === 'webCrawl'
+                    ? await webCrawl(url, parseInt(limit))
+                    : await xmlScrape(url, parseInt(limit))
             if (process.env.DEBUG === 'true') console.info(`pages: ${JSON.stringify(pages)}, length: ${pages.length}`)
```
vinodkiran, line 136: Can we use the logger instead of console for DEBUG ?

packages/components/nodes/documentloaders/Cheerio/Cheerio.ts
```diff
@@ -127,13 +128,22 @@ class Cheerio_DocumentLoaders implements INode {
             if (!limit) limit = '10'
             else if (parseInt(limit) < 0) throw new Error('Limit cannot be less than 0')
```
vinodkiran, line 129: Minor Readability:: avoid multiple calls to parseInt

packages/ui/src/ui-component/dialog/ManageScrapedLinksDialog.js
```diff
@@ -0,0 +1,184 @@
...
+                    <>
+                        {loading && <BackdropLoader open={loading} />}
+                        <div style={{ display: 'flex', alignItems: 'center', justifyContent: 'center' }}>
+                            <Typography sx={{ my: 2 }}>Links scraped from the URL will appear here</Typography>
+                        </div>
+                    </>
+                )}
+            </DialogContent>
+            <DialogActions>
+                <Button onClick={onCancel}>Cancel</Button>
```
vinodkiran, line 166: dialogProps.cancelButtonName ?

packages/ui/src/ui-component/dialog/ManageScrapedLinksDialog.js
```diff
@@ -0,0 +1,184 @@
...
+                <StyledButton variant='contained' onClick={handleSaveLinks}>
+                    Save
```
vinodkiran, line 168: dialogProps.saveButtonName ?

packages/components/nodes/documentloaders/Cheerio/Cheerio.ts
```diff
@@ -90,6 +90,7 @@ class Cheerio_DocumentLoaders implements INode {
         const textSplitter = nodeData.inputs?.textSplitter as TextSplitter
         const metadata = nodeData.inputs?.metadata
         const relativeLinksMethod = nodeData.inputs?.relativeLinksMethod as string
+        const selectedLinks = nodeData.inputs?.selectedLinks as string[]
         let limit = nodeData.inputs?.limit as string
```
vinodkiran, line 94: can we set the default value in the inputs to 10 ?

vinodkiran (changes requested):

packages/ui/src/ui-component/dialog/ManageScrapedLinksDialog.js
0xi4o: I think it's fine for it to be hardcoded here. We don't really gain anything by making it a prop since the component has a specific use and not a generalized one like `Dialog` or `Button`.

packages/ui/src/ui-component/dialog/ManageScrapedLinksDialog.js
0xi4o: Same as the cancel button

packages/components/nodes/documentloaders/Cheerio/Cheerio.ts
0xi4o: You're right. Updated the code so `parseInt` is only used once.

packages/components/nodes/documentloaders/Cheerio/Cheerio.ts
0xi4o: Set the default value for the limit field to `10`.

packages/components/nodes/documentloaders/Cheerio/Cheerio.ts
0xi4o: Updated the code to use `logger` instead of `console`

HenryHengZJ:
Bug:

1.) Enter `https:flowiseai.com` into the Cheerio node:
image

2.) Open Manage Links and modified to `https://docs.flowiseai.com`:
image

3.) Fetch links

4.) Click save

5.) On the canvas, the Cheerio Node URL is still `https:flowiseai.com`, same as when you click manage links:
image

0xi4o:
@HenryHengZJ Fixed the bug. The URL field in the node will be updated if we change it in the manage links dialog.

0xi4o:
@HenryHengZJ Sure! I'll update it :+1:

0xi4o:
@HenryHengZJ There's already an `isUpsert` flag in the `buildLangchain()` in `server/src/index.ts`. It's set to `true` only during upsert operations. When calling the `prediction` endpoints, it's not set so it won't do upsert. This means the scraping will also only happen during upsert and not during prediction.

HenryHengZJ:
> @HenryHengZJ There's already an `isUpsert` flag in the `buildLangchain()` in `server/src/index.ts`. It's set to `true` only during upsert operations. When calling the `prediction` endpoints, it's not set so it won't do upsert. This means the scraping will also only happen during upsert and not during prediction.

it wont upsert, but it will still go through the `init` of Cheerio node, scraping all the links, which might take significant portion of the time

0xi4o:
@HenryHengZJ Added a condition to skip initializing web scraper nodes during prediction.

packages/server/src/utils/index.ts
```diff
@@ -328,6 +328,14 @@ export const buildLangchain = async (
                 logger.debug(`[server]: Finished upserting ${reactFlowNode.data.label} (${reactFlowNode.data.id})`)
                 break
             } else {
+                // skip initializing web scraper nodes during prediction since they would have already run as a part of upsert
```
HenryHengZJ, line 331: @0xi4o actually ignore me, my bad i was testing on diff branch. The addition of this make it not working. I think its fine as it is. If you can just remove this

HenryHengZJ (approved):

packages/server/src/utils/index.ts
HenryHengZJ: also if u can pull latest changes from main

0xi4o: Reverted the changes.

0xi4o: Pulled from main as well.

PR #1570, opened 2024-01-19
Feature/Add Verbose to Chains

HenryHengZJ:
- add verbose (console callback manager) to chains
- updated conversation chain to have ability to use chatprompt template

image


maximometascript:
Is the document loader input in the additional parameters?

PR #1586, opened 2024-01-21
Feature/Update Return Response

HenryHengZJ:
Some users would like to have more info from API response:
image


giorgosera:
Wow thanks! That was quick :) 

Will this be released in the new version?

PR #1591, opened 2024-01-23
Allows CustomFunction to be set as an Ending Node.

### LINKED ISSUE #1590: [FEATURE] Allows any node to be set as the end node.
**Describe the feature you'd like**

Sometimes the output results of the model are used to call third-party tools to get the results, which are already the final results. At this time, adding chain or agent will not achieve the ideal output because of the randomness of the model.

Wish there was a feature that would allow any node to be set as the end node.

Let me know what you think @HenryHengZJ

mokeyish:
closes #1590 

With this PR, we can set CustomFunction as an Ending Node.

图片

### DIFF
--- a/packages/components/nodes/utilities/CustomFunction/CustomFunction.ts
+++ b/packages/components/nodes/utilities/CustomFunction/CustomFunction.ts
@@ -52,11 +52,19 @@ class CustomFunction_Utilities implements INode {
                 label: 'Output',
                 name: 'output',
                 baseClasses: ['string', 'number', 'boolean', 'json', 'array']
+            },
+            {
+                label: 'Ending Node',
+                name: 'EndingNode',
+                baseClasses: [this.type]
             }
         ]
     }
 
     async init(nodeData: INodeData, input: string, options: ICommonObject): Promise<any> {
+        const isEndingNode = nodeData?.outputs?.output === 'EndingNode'
+        if (isEndingNode && !options.isRun) return // prevent running both init and run twice
+
         const javascriptFunction = nodeData.inputs?.javascriptFunction as string
         const functionInputVariablesRaw = nodeData.inputs?.functionInputVariables
         const appDataSource = options.appDataSource as DataSource
@@ -123,13 +131,18 @@ class CustomFunction_Utilities implements INode {
         const vm = new NodeVM(nodeVMOptions)
         try {
             const response = await vm.run(`module.exports = async function() {${javascriptFunction}}()`, __dirname)
-            if (typeof response === 'string') {
+
+            if (typeof response === 'string' && !isEndingNode) {
                 return handleEscapeCharacters(response, false)
             }
             return response
         } catch (e) {
             throw new Error(e)
         }
+    }
+
+    async run(nodeData: INodeData, input: string, options: ICommonObject): Promise<string> {
+        return await this.init(nodeData, input, { ...options, isRun: true })
     }
 }
 
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -465,8 +465,12 @@ export class App {
                 const endingNodeData = endingNode.data
                 if (!endingNodeData) return res.status(500).send(`Ending node ${endingNode.id} data not found`)
 
-                if (endingNodeData && endingNodeData.category !== 'Chains' && endingNodeData.category !== 'Agents') {
-                    return res.status(500).send(`Ending node must be either a Chain or Agent`)
+                const isEndingNode = endingNodeData?.outputs?.output === 'EndingNode'
+
+                if (!isEndingNode) {
+                    if (endingNodeData && endingNodeData.category !== 'Chains' && endingNodeData.category !== 'Agents') {
+                        return res.status(500).send(`Ending node must be either a Chain or Agent`)
+                    }
                 }
 
                 isStreaming = isFlowValidForStream(nodes, endingNodeData)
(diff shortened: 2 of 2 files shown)


mokeyish:
@HenryHengZJ This is the case. My chatflow uses llm to generate SQL, input it to customFunction, then send it to the database, and finally construct the final JSON result from this customFunction.

Jaredude:
This is interesting. It allows Flowise to return a custom API output! I can see a number of useful benefits from consumption by other APIs or custom chat handling. Something I can think for a current project I'm working on would be to return an additional attribute in the payload so that perform additional logic on the client based upon output beyond the "text" value returned to the chat bot.

```
{
  text: "I just found you a great product! I'm sending you there now!",
  additionalResource: "https://example.com/cool/product",
  savedSearchId: "aie9-nnie4-87ugoi-817"
}
```

I feel like it might make sense to include an additional API entry point distinct from prediction. Similar to internal-prediction, an additional variable could be added to buildChatflow: isCustomOutput. By doing this, we should be able to simplify the logic in buildChatflow as well; reducing the changes to packages/server/src/index.ts and allowing for additional validation logic to be added in the future (e.g. define a flow as custom output and only allow it to be called from the new entry point and not from chat by mistake <<< not something needed to launch a feature like this, but someone certainly could make the mistake of trying to chat with a custom output).

Perhaps something along the lines of /api/v1/custom-prediction/:id  would work?

@mokeyish Minor ts adjustment on the change to packages/components/nodes/utilities/CustomFunction/CustomFunction.ts. The value of isEndingNode on line 131 ends up being iffy (boolean | 0 | undefined) instead of just boolean. While 0 and undefined are both falsey, typecasting to boolean and using ternary logic makes the boolean clear. We can also leave off the check of Object.keys().length since we're already checking that nodeData.outputs exists and it can only be undefined or an Object. An object with no attributes will still return an array to Object.values().

`const isEndingNode: boolean = nodeData.outputs && Object.values(nodeData.outputs).includes('EndingNode') ? true : false`

Also, I'm happy to help out on this PR if needed.

mokeyish:
@Jaredude Thanks, I've updated it. But this is more concise and can also infer the boolean type.

图片

mokeyish:
> @HenryHengZJ and @mokeyish , could this PR also solve this additional use case?: #1489


It seems like it should be possible. Can you pull the code from this branch and test it?

HenryHengZJ:
@dkindlund I think for your use case, we cant do that yet, because you cant take the output from agent to custom function, only llmchain at the moment

@mokeyish exact same thought. thats why we are working on a workflow feature which allow more sequential flow. that way you have a seperate endpoint `api/v1/workflow/` and you can import chatflow inside workflow to do additional stuff. We want to keep chatflow to chat-based interaction, and workflow for non-chat-based pipeline and processing.

having said that, as a near term solution, we can use custom function as ending node

HenryHengZJ (approved):

Jaredude:
While going through debug, I found one issue: The code runs twice.

run is just a proxy for init, so the end result is that all of the Custom JS code is running twice for every request.

Screenshot 2024-01-25 at 11 21 27 AM

This is only happening with the CustomJS that's an end node. Any Custom JS that's not an end node will not execute twice.

HenryHengZJ:
@Jaredude good spot, cus we are hitting both `init` and `run` functions

HenryHengZJ (changes requested):
in the init function, add a check to see if the output is `Ending Node`, if yes, dont execute sandbox stuff, this way we can prevent running both `init` and `run` twice

Jaredude:
This should be an easy workaround for the double run issue...
Place the following at the top of init:
`        if (nodeData?.outputs?.output === 'EndingNode' && !options.isRun) return
`
and for run, call init like this:
`        return await this.init(nodeData, input, { ...options, isRun: true })
`

Example: In the first CustomJS, which is an output, init is called directly without options.isRun. On the second CustomJS, init is called but returns immediately since it's an ending node. The call to run (since it's an ending node) then adds options.isRun, and the ending node code runs.
Screenshot 2024-01-25 at 11 47 54 AM

I have only done basic testing on this with Lorem Ipsum data so far. I'll run this using some real world flows and see if there are any issue that surface.

Jaredude:
@mokeyish better yet, I think we can dual purpose the logic at the top...
```
  const isEndingNode = nodeData?.outputs?.output === 'EndingNode'
  if (isEndingNode && !options.isRun) return
```

Then the logic on line 1659 is already taken care of. I'll add comments on the PR code.

packages/server/src/index.ts
```diff
@@ -1652,20 +1656,20 @@ export class App {
                     const endingNodeData = endingNode.data
                     if (!endingNodeData) return res.status(500).send(`Ending node ${endingNode.id} data not found`)
 
-                    if (endingNodeData && endingNodeData.category !== 'Chains' && endingNodeData.category !== 'Agents') {
-                        return res.status(500).send(`Ending node must be either a Chain or Agent`)
-                    }
+                    const isEndingNode = Object.values(endingNodeData.outputs ?? {}).includes('EndingNode')
```
Jaredude, line 1659: To solve the double run issue, we can move the isEndingNode logic to the top of the init function and then reuse the variable down here:
```
  const isEndingNode = nodeData?.outputs?.output === 'EndingNode'
  if (isEndingNode && !options.isRun) return

```

HenryHengZJ: so there are 2 ways to solve the issue right, one is we do it at the `CustomFunction.ts` like my comment above, or the way u suggest where we do it at the `server/index.ts`

Jaredude: Oops... I put this in the wrong place... I'll move this comment to CustomFunction.ts

packages/components/nodes/utilities/CustomFunction/CustomFunction.ts
```diff
@@ -123,14 +128,20 @@ class CustomFunction_Utilities implements INode {
         const vm = new NodeVM(nodeVMOptions)
         try {
             const response = await vm.run(`module.exports = async function() {${javascriptFunction}}()`, __dirname)
-            if (typeof response === 'string') {
+            const isEndingNode = Object.values(nodeData.outputs ?? {}).includes('EndingNode')
```
Jaredude, line 131: To solve the double run issue, we can move the isEndingNode logic to the top of the init function and then reuse the variable down here:
```
  const isEndingNode = nodeData?.outputs?.output === 'EndingNode'
  if (isEndingNode && !options.isRun) return

```

mokeyish:
@HenryHengZJ @Jaredude please review again.

Jaredude (commented):
I ran this through two separate LLM Chain flows and received the expected outcome. Here's a screenshot. 
Screenshot 2024-01-25 at 11 22 03 PM

I also tested validations such as no end node, multiple JS functions, JS code that errors, and also existing node validation (e.g. LLM Chain with an Output Prediction and no end node). Everything is working as expected!

HenryHengZJ (approved):

HenryHengZJ:
awesome, thanks guys!

PR #1599, opened 2024-01-23
Added support for gpt-4 and gpt-4-32k models

dkindlund:
Hey @HenryHengZJ , I've updated the AzureOpenAI LLM node to support gpt-4 and gpt-4-32k models, since those were missing from the dropdown selectors.  Let me know if you have any questions. Thanks!

### DIFF
--- a/packages/components/nodes/llms/Azure OpenAI/AzureOpenAI.ts
+++ b/packages/components/nodes/llms/Azure OpenAI/AzureOpenAI.ts
@@ -18,7 +18,7 @@ class AzureOpenAI_LLMs implements INode {
     constructor() {
         this.label = 'Azure OpenAI'
         this.name = 'azureOpenAI'
-        this.version = 2.0
+        this.version = 2.1
         this.type = 'AzureOpenAI'
         this.icon = 'Azure.svg'
         this.category = 'LLMs'
@@ -89,6 +89,14 @@ class AzureOpenAI_LLMs implements INode {
                     {
                         label: 'gpt-35-turbo',
                         name: 'gpt-35-turbo'
+                    },
+                    {
+                        label: 'gpt-4',
+                        name: 'gpt-4'
+                    },
+                    {
+                        label: 'gpt-4-32k',
+                        name: 'gpt-4-32k'
                     }
                 ],
                 default: 'text-davinci-003',


packages/components/nodes/llms/Azure OpenAI/AzureOpenAI.ts
```diff
@@ -89,6 +89,14 @@ class AzureOpenAI_LLMs implements INode {
                     {
                         label: 'gpt-35-turbo',
                         name: 'gpt-35-turbo'
+                    },
```
HenryHengZJ, line 92: these are already available on `Azure ChatOpenAI` no?

HenryHengZJ: image

HenryHengZJ: @dkindlund is gpt4 and gpt3.5 supported in the Completion endpoint of Azure? `Azure OpenAI`in Flowise is using Completion endpoint whereas `Azure ChatOpenAI` is using Chat Completion endpoint.

what is the difference between using gpt-4 in `Azure ChatOpenAI` versus `Azure OpenAI`?

dkindlund: Hey @HenryHengZJ , you bring up a good point.  I had originally thought that the `llms/Azure OpenAI/AzureOpenAI.ts` was using the legacy `completions` endpoint while the `chatmodels/AzureChatOpenAI/AzureChatOpenAI.ts` uses the new `chat.completions` endpoint -- is that correct?

Regardless, I was able to successfully get the `llms/Azure OpenAI/AzureOpenAI.ts` node to correctly work using both the `gpt-4` and `gpt-4-32k` models in Azure for testing -- so I figured others who use Azure might want similar functionality.

That said, if you have plans to retire/deprecate the `llms/Azure OpenAI/AzureOpenAI.ts` node, I understand if you don't want to accept this PR.

HenryHengZJ: yep you are right `llms/Azure OpenAI/AzureOpenAI.ts` was using the legacy completions endpoint while the `chatmodels/AzureChatOpenAI/AzureChatOpenAI.ts` uses the new chat.completions endpoint.

I'm just wondering if gpt-4 and others are supported on the completions endpoint, trying to avoid duplication. But since you have tested and it works, thats good!

HenryHengZJ (approved):

PR #1602, opened 2024-01-24
feature : include chat message id in result

tsiougkosn:
Return the chat message id in the api response.

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -1814,7 +1814,8 @@ export class App {
             if (result?.sourceDocuments) apiMessage.sourceDocuments = JSON.stringify(result.sourceDocuments)
             if (result?.usedTools) apiMessage.usedTools = JSON.stringify(result.usedTools)
             if (result?.fileAnnotations) apiMessage.fileAnnotations = JSON.stringify(result.fileAnnotations)
-            await this.addChatMessage(apiMessage)
+            const chatMessage = await this.addChatMessage(apiMessage)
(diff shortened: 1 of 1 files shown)


packages/server/src/index.ts
```diff
@@ -1814,7 +1814,8 @@ export class App {
             if (result?.sourceDocuments) apiMessage.sourceDocuments = JSON.stringify(result.sourceDocuments)
             if (result?.usedTools) apiMessage.usedTools = JSON.stringify(result.usedTools)
             if (result?.fileAnnotations) apiMessage.fileAnnotations = JSON.stringify(result.fileAnnotations)
-            await this.addChatMessage(apiMessage)
+            const chatMessage = await this.addChatMessage(apiMessage)
+            result.chatMessageId = chatMessage.id
```
HenryHengZJ, line 1818: do you need the `chatId` (similar to sessionid) or the id of the message?
FYI - the latest changes return chatId along with other data - https://github.com/FlowiseAI/Flowise/pull/1586

tsiougkosn: We need the message id, so we can keep feedback on the specific response. Else we need to have a second api call to /chatmessage in order DESC to get the last message, but this is not efficient

tsiougkosn: Our end goal is to have map the message with the trace in Langfuse

HenryHengZJ: got it!

HenryHengZJ (approved):

PR #1608, opened 2024-01-25
Implementing CORS and CSP headers from config

automaton82:
This PR implements both [Cross Origin Resource Sharing (CORS)](https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS) and [Content Security Policy (CSP)](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/frame-ancestors) frame ancestors headers.

Both are very important from a security standpoint, with CSP specifically stopping XSS attacks (iframe embedding).

These new options use two new _optional_ env variables:
```
# CORS_ORIGINS="*"
# EMBEDDING_ORIGINS="*"
```

By default both are not set and backwards compatible. If they are not set, we do not emit CSP headers (as before), and CORS sends the origin domain (also as before).

Here's an example that allows CORS requests and embedding from a specific domain:
```
CORS_ORIGINS="https://www.testme.com"
EMBEDDING_ORIGINS="https://www.testme.com"
```

Here's an example that allows CORS, but not iframe embedding except by itself:
```
CORS_ORIGINS="https://www.testme.com"
EMBEDDING_ORIGINS="self"
```

Here's an example that allows multiple CORS domains, but no embedding at all:
```
CORS_ORIGINS="https://www.testme.com, https://www.testmeagain.com"
EMBEDDING_ORIGINS="none"
```

Here's an example that allows multiple CORS domains and embedding domains:
```
CORS_ORIGINS="https://www.testme.com, https://www.testmeagain.com"
EMBEDDING_ORIGINS="https://www.testme.com, https://www.testmeagain.com"
```
This applies the same CORS headers to both websockets and node backend.

### DIFF
--- a/packages/server/.env.example
+++ b/packages/server/.env.example
@@ -1,4 +1,6 @@
 PORT=3000
+# CORS_ORIGINS="*"
+# EMBEDDING_ORIGINS="*"
 # DATABASE_PATH=/your_database_path/.flowise
 # APIKEY_PATH=/your_api_key_path/.flowise
 # SECRETKEY_PATH=/your_api_key_path/.flowise
--- a/packages/server/.env.example
+++ b/packages/server/.env.example
@@ -1,6 +1,6 @@
 PORT=3000
 # CORS_ORIGINS="*"
-# EMBEDDING_ORIGINS="*"
+# IFRAME_ORIGINS="*"
 # DATABASE_PATH=/your_database_path/.flowise
 # APIKEY_PATH=/your_api_key_path/.flowise
 # SECRETKEY_PATH=/your_api_key_path/.flowise
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -62,7 +62,7 @@ import { CachePool } from './CachePool'
 import { ICommonObject, IMessage, INodeOptionsValue, handleEscapeCharacters } from 'flowise-components'
 import { createRateLimiter, getRateLimiter, initializeRateLimiter } from './utils/rateLimit'
 import { addAPIKey, compareKeys, deleteAPIKey, getApiKey, getAPIKeys, updateAPIKey } from './utils/apiKey'
-import { sanitizeMiddleware, getAllowedCorsOrigins, getAllowedEmbeddingOrigins } from './utils/XSS'
+import { sanitizeMiddleware, getCorsOptions, getAllowedIframeOrigins } from './utils/XSS'
 import axios from 'axios'
 import { Client } from 'langchainhub'
 import { parsePrompt } from './utils/hub'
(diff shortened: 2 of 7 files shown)


automaton82:
Ping @HenryHengZJ this is the CORS and CSP security PR we spoke about in December.

HenryHengZJ:
> I guess the term 'embedding' is a bit confusing here since it's for iframe embedding, not AI embeddings.

yeah I agree, maybe rename as `IFRAME_ORIGINS`?

packages/server/.env.example
```diff
@@ -1,4 +1,6 @@
 PORT=3000
+# CORS_ORIGINS="*"
+# EMBEDDING_ORIGINS="*"
```
HenryHengZJ, line 3: if you search for `DATABASE_PATH`, you might find other places that are using it. We might need to add these new vars to the missing places.

packages/server/src/index.ts
```diff
@@ -1863,7 +1885,14 @@ export async function start(): Promise<void> {
 
     const io = new Server(server, {
         cors: {
-            origin: '*'
+            origin: function (origin: string | undefined, callback: (err: Error | null, allow?: boolean) => void) {
```
HenryHengZJ, line 1888: the function looks the same as the one above, is it possible to refractor into a single function?

packages/server/.env.example
automaton82: Found them. Addressed in 657dace.

packages/server/src/index.ts
automaton82: Agreed. Addressed in 657dace.

automaton82:
> > I guess the term 'embedding' is a bit confusing here since it's for iframe embedding, not AI embeddings.
> 
> yeah I agree, maybe rename as `IFRAME_ORIGINS`?

Yup. Changed to this in 657dace.

HenryHengZJ (approved):
looks good to me! let's see if @0xi4o has any more stuff to add

automaton82:
Hey @0xi4o did you get a chance to review?

0xi4o:
Yep. Reviewed it :+1: We can go ahead and merge this one @HenryHengZJ

ahhany:
i am having the same issue (i think :) ), is there a step-by-step guide to how setup the chatbot on a hosted website.. the flows work from the console, but when I put the <script> tag in the website it only shows the 3 dots rolling.

toi500:
@automaton82 I have been checking this CORS implementation and was able to bypass it without any problem using a simple API route. I am unsure if this is a limitation of the original implementation or if the Flowwise codebase has changed in the last few months, causing the CORS protection to malfunction.

In either case, even with the CORS settings, all embedded chatbots or unprotected Flowwise implementations can be bypassed easily at present.

Issue #1598, opened 2024-01-23
[BUG] Unable to connect to PostgreSQL with self-signed SSL


efthymiosh:
**Describe the bug**

We have been setting the env variable `PGSSLMODE` to `no-verify` for connecting to our self-signed PostgreSQL database.
After upgrading from 1.4.1 to 1.4.10 we observed the following error:

```bash
2024-01-23 19:34:32 [ERROR]: ❌ [server]: Error during Data Source initialization: no pg_hba.conf entry for host "***", user "***", database "flowise", no encryption
```

My investigation pinpointed to [this](https://github.com/FlowiseAI/Flowise/commit/735edab9ae5d85be3999978d1ae24b984cfcb5a1) commit, which sets `ssl: true/false`.

Configuring the `DATABASE_SSL` to `true` yields:

```bash
2024-01-23 20:47:26 [ERROR]: ❌ [server]: Error during Data Source initialization: self-signed certificate in certificate chain
Error: self-signed certificate in certificate chain
    at TLSSocket.onConnectSecure (node:_tls_wrap:1659:34)
    at TLSSocket.emit (node:events:517:28)
    at TLSSocket._finishInit (node:_tls_wrap:1070:8)
    at ssl.onhandshakedone (node:_tls_wrap:856:12)
```

It seems like `PGSSLMODE` is now ignored.

**To Reproduce**

1. Set-up a postgresql with a self-signed SSL certificate
2. Configure the postgresql datasource with `DATABASE_SSL=true`
3. Configure `PGSSLMODE=no-verify`

**Expected behavior**
Successful datasource connection.

**Screenshots**
N/A

**Flow**
N/A

**Setup**
-   Installation: docker/k8s with flowiseai/flowise:1.4.10
-   Command: sh -ec "flowise start"
-   Flowise Version 1.4.10
-   OS: Linux
-   Browser N/A

**Additional context**
N/A

HenryHengZJ:
do we have `PGSSLMODE` env variable in Flowise?

Jaredude:
> do we have `PGSSLMODE` env variable in Flowise?

PGSSLMODE is read in by the pg npm library, and it cannot be combined when we add ssl to the connection (line 81 of the connection-parameters.js file in the pg library.

@HenryHengZJ  I'll make another PR for supporting self-signed certificates via an env var with the cert as a base64 string like we discussed in the original PR, following [these guidelines](https://node-postgres.com/features/ssl).

[...]

PR #1613, opened 2024-01-25
adds DATABASE_SSL_KEY_BASE64 for pg connection

Jaredude:
DATABASE_SSL_KEY_BASE64 takes priority over DATABASE_SSL env var

If neither are provided, no ssl value will be used. This allows for the usage of PGSSLMODE

for example, using either of these two will establish a secure connection to the Postgres server
DATABASE_SSL=true
PGSSLMODE=verify-full

Using the following will override the above and allow you to pass your self-signed certificate to create a secure connection
DATABASE_SSL_KEY_BASE64=<Self signed certificate in BASE64>

Using neither of the flags allows using the [unsafe/not recommended](https://www.postgresql.org/docs/current/libpq-ssl.html)  PGSSLMODE=no-verify

Addresses issue #1598


HenryHengZJ (approved):

HenryHengZJ:
awesome this looks good! thanks man @Jaredude !

dkindlund:
Looking forward to using this new feature, once the PR is merged!  👍   Thanks again, @Jaredude!

inno-elon:
how do you obtain the Postgres SSL certificate encoded in base64? I can't get SSL to work with Postgres

dkindlund:
> how do you obtain the Postgres SSL certificate encoded in base64? I can't get SSL to work with Postgres

Hey @inno-elon , you can use a tool like `openssl` to extract the X.509 SSL/TLS certificate from your Postgres server.  The format is this:

```
openssl s_client -showcerts <DATABASE_IP>:5432
```

In that initial output, you should see base64 text returned from the server wrapped in a `-----BEGIN CERTIFICATE-----` and `-----END CERTIFICATE-----` block.  You want to copy that entire block of text (including the BEGIN/END lines) and set that as your `DATABASE_SSL_KEY_BASE64` value, accordingly.

harvetech:
btw is this even a valif falg PGSSLMODE=no-verify ? I don't see this as an option here https://www.postgresql.org/docs/current/libpq-ssl.html

PR #1619, opened 2024-01-26
Bugfix/custom function ending node

HenryHengZJ:


Jaredude:
@HenryHengZJ So glad to see this bug fix! I hadn't pulled the latest from main, and when I did some additional testing just now, I was encountering an error using an agent I hadn't used for the original testing. I was going to comment back on the original PR. This fixes the issue I was encountering on the stale code. Thanks!

HenryHengZJ:
> @HenryHengZJ So glad to see this bug fix! I hadn't pulled the latest from main, and when I did some additional testing just now, I was encountering an error using an agent I hadn't used for the original testing. I was going to comment back on the original PR. This fixes the issue I was encountering on the stale code. Thanks!

Yep i noticed the issue after doing some testing 😁

PR #1634, opened 2024-01-29
mysql-ssl

niztal:
In order to increase the level of security for flowise users who are using **MYSQL** DB, I've added the SSL support.

This PR based on similar PR done for postgres https://github.com/FlowiseAI/Flowise/pull/1613

In order to use this you need to setup **DATABASE_SSL=TRUE** env variable OR **DATABASE_SSL_KEY_BASE64**={YOUR_SSL_BASE64} which overrides the DATABASE_SSL env.

@Jaredude @HenryHengZJ please review

Thanks 🐉

### DIFF
--- a/packages/server/src/DataSource.ts
+++ b/packages/server/src/DataSource.ts
@@ -89,5 +89,5 @@ const getDatabaseSSLFromEnv = () => {
     } else if (process.env.DATABASE_SSL === 'true') {
         return true
     }
-    return {}
+    return undefined
 }
--- a/packages/server/src/DataSource.ts
+++ b/packages/server/src/DataSource.ts
@@ -40,7 +40,19 @@ export const init = async (): Promise<void> => {
                 synchronize: false,
                 migrationsRun: false,
                 entities: Object.values(entities),
-                migrations: mysqlMigrations
+                migrations: mysqlMigrations,
+                ...(process.env.DATABASE_SSL_KEY_BASE64
+                    ? {
+                        ssl: {
+                            rejectUnauthorized: false,
+                            ca: Buffer.from(process.env.DATABASE_SSL_KEY_BASE64, 'base64')
+                        }
+                    }
+                    : process.env.DATABASE_SSL === 'true'
+                        ? {
+                            ssl: true
+                        }
+                        : {}),
             })
             break
         case 'postgres':
@@ -53,16 +65,16 @@ export const init = async (): Promise<void> => {
                 database: process.env.DATABASE_NAME,
                 ...(process.env.DATABASE_SSL_KEY_BASE64
                     ? {
-                          ssl: {
-                              rejectUnauthorized: false,
-                              cert: Buffer.from(process.env.DATABASE_SSL_KEY_BASE64, 'base64')
-                          }
-                      }
+                        ssl: {
+                            rejectUnauthorized: false,
+                            cert: Buffer.from(process.env.DATABASE_SSL_KEY_BASE64, 'base64')
+                        }
+                    }
                     : process.env.DATABASE_SSL === 'true'
-                    ? {
-                          ssl: true
-                      }
-                    : {}),
+                        ? {
+                            ssl: true
+                        }
+                        : {}),
                 synchronize: false,
                 migrationsRun: false,
                 entities: Object.values(entities),
--- a/packages/server/src/DataSource.ts
+++ b/packages/server/src/DataSource.ts
@@ -43,16 +43,16 @@ export const init = async (): Promise<void> => {
                 migrations: mysqlMigrations,
                 ...(process.env.DATABASE_SSL_KEY_BASE64
                     ? {
-                        ssl: {
-                            rejectUnauthorized: false,
-                            ca: Buffer.from(process.env.DATABASE_SSL_KEY_BASE64, 'base64')
-                        }
-                    }
+                          ssl: {
+                              rejectUnauthorized: false,
+                              ca: Buffer.from(process.env.DATABASE_SSL_KEY_BASE64, 'base64')
+                          }
+                      }
                     : process.env.DATABASE_SSL === 'true'
-                        ? {
-                            ssl: true
-                        }
-                        : {}),
+                    ? {
+                          ssl: true
+                      }
+                    : {})
             })
             break
         case 'postgres':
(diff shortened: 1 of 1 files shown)


Jaredude:
I think we should refactor to keep the inline code concise instead of the same code being used twice. Something along these lines:
```
const getDatabaseSSLFromEnv = () => {
  if (process.env.DATABASE_SSL_KEY_BASE64) {
    return {
      ssl: {
        rejectUnauthorized: false,
        ca: Buffer.from(process.env.DATABASE_SSL_KEY_BASE64, 'base64')
      }
    };
  } else if (process.env.DATABASE_SSL === 'true') {
    return { ssl: true };
  }
  return {};
}
```
// Usage then can be:
```
{
  migrations: mysqlMigrations,
  ...getDatabaseSSLFromEnv(),
}

```
and 

```
{
  migrations: postgresMigrations,
  ...getDatabaseSSLFromEnv(),
}
```

niztal:
Hi @Jaredude 

Basically I agree with you it will look better and I must say I started doing it, but there's a small difference between postgres and mysql and that's the pem file they require.

- Postgress (according to your PR) - require the "cert" pem
- While, MYSQL require the "ca" pem

If you still think it worth it, I can think of a valid refactor in order to combine these two options.

Waiting for your reply,

Thanks

HenryHengZJ:
I am okay with this since mysql use `ca` and postgres use `cert`, refractoring it might not be necessary

niztal:
> I am okay with this since mysql use `ca` and postgres use `cert`, refractoring it might not be necessary

@HenryHengZJ can you please approve?

Thanks

Jaredude:
Postgres can be done the same way by using the ca instead of cert.  It might be beneficial to use the same for both so that it's the same standard pem file the user would base64 encode. Anyone who implemented the existing cert env value would need to change to the ca value, but the pg library will work as well using ca.

Jaredude:
Related note: switching pg to using ca just like the mysql implementation likely wouldn't cause a breaking change for existing pg users who have been using self-signed pg because in the prior thread #1598 it was mentioned PGSSLMODE=no-verify was being used to bypass the certificate all together. The code will still allow for that env var to be used by existing pg users. mysql2 lib doesn't have equivalent env var checks, so there's no conflict there either.

niztal:
Ok @Jaredude if that's the case I'll unify both cases as you suggested and let you and @HenryHengZJ review it

niztal:
@Jaredude @HenryHengZJ 

Please review after refactoring, thanks 🌈

packages/server/src/DataSource.ts
```diff
@@ -51,18 +52,7 @@ export const init = async (): Promise<void> => {
...
-                              rejectUnauthorized: false,
-                              cert: Buffer.from(process.env.DATABASE_SSL_KEY_BASE64, 'base64')
-                          }
-                      }
-                    : process.env.DATABASE_SSL === 'true'
-                    ? {
-                          ssl: true
-                      }
-                    : {}),
+                ssl: getDatabaseSSLFromEnv(),
```
Jaredude, line 55: This will result in breaking changes for existing Flowise pg installs given the way pg node uses env vars internally on connection instantiation.

Using ssl: getDatabaseSSLFromEnv() will result in ssl: {} when there are no Flowise env vars present. That will in turn not allow the usage of PGSSLMODE, which will be a breaking change for existing Flowise installs using pg. 

getDatabaseSSLFromEnv should be spread instead of being a fixed value for ssl so that pg node continues to function as is. In turn getDatabaseSSLFromEnv should return ssl as the top level object when the Flowise env vars are present.

packages/server/src/DataSource.ts
```diff
@@ -89,3 +79,15 @@ export function getDataSource(): DataSource {
     }
     return appDataSource
 }
+
+const getDatabaseSSLFromEnv = () => {
+    if (process.env.DATABASE_SSL_KEY_BASE64) {
+        return {
```
Jaredude, line 85: See comment above. getDatabaseSSLFromEnv should return
```
{
  ssl: {
    self signed ssl attributes
  }
}
```

and below should return 
{
  ssl: true
}

the final return of {} will return an empty object that will spread to nothing and allow pg node to continue to use PGSQLMODE as the lib is designed.

mysql2 does not appear to be impacted by using the same approach and should work by using the same spread methodology. mysql2 does not reference any env var internally.

packages/server/src/DataSource.ts
niztal: @Jaredude what if we'll return undefined in case no env vars present?

packages/server/src/DataSource.ts
```diff
@@ -89,3 +79,15 @@ export function getDataSource(): DataSource {
...
+            rejectUnauthorized: false,
+            ca: Buffer.from(process.env.DATABASE_SSL_KEY_BASE64, 'base64')
```
niztal, line 87: @Jaredude have you tested it? 

I think it should be like this:

`Buffer.from(process.env.DATABASE_SSL_KEY_BASE64, 'base64').toString('utf-8')`

packages/server/src/DataSource.ts
Jaredude: Yes, returning undefined instead of {} will work 
Screenshot 2024-01-30 at 3 34 13 PM

packages/server/src/DataSource.ts
niztal: @Jaredude WDYT?

Jaredude: I don't see anything in the pg docs or mysql2 docs requiring the data to be utf8

niztal: @Jaredude ok thanks , highly appreciated

One last question, can you please explain how does it work when we provide only ssl:true but no certificate?

Jaredude: @niztal Great question... Hope this helps: the reason that ssl: true works without supplying a ca or cert is that each library will reference a trust store(s) to verify the certificate coming from the server. The best analogy I can think of would be visiting a website over https instead of http. The web browser validates the ssl connection by checking to make sure the authority of the certificate is indeed valid. If it's not, you get the screen showing you the certificate is not valid. The pg and mysql2 libraries have their own trust stores that they use, and that allows a simple ssl: true property to work. In the case of the mysql2 lib, certs from major providers are directly in the managed source code for mysql2. You can view them under the files in node_modules/mysql2.

niztal: Wow amazing explanation @Jaredude thank you so much

niztal:
@HenryHengZJ can you please approve this PR?

It was reviewed by @Jaredude 

Thank you both 🙏🏼

HenryHengZJ (approved):

PR #1667, opened 2024-02-02
Fix error in yarn build due to missing dependency

tirongi:
When ran yarn build it went into error due to missing DynamicStructuredTool available in langchain/core/tools and used in RetrieverTool.ts.
This had brought to fail the retriever to work properly.
Solution was to add the @langchain/core dependency.


eleonel:
I had the same problem and this fix solves the issue.

HenryHengZJ:
thanks! we'll be doing a big upgrade to langchain v0.1.0 and that includes this dependency as well

HenryHengZJ:
closing this as moved to https://github.com/FlowiseAI/Flowise/pull/1691

PR #1672, opened 2024-02-03
Add a zod-schema based JSON output parser ("Advanced Structured Output Parser")

KennyVaneetvelde:
Title says it all, I added an "Advanced Structured Output Parser" that uses Zod directly.

At first I had modified the original node - but I figured it's probably much nicer to have it as a separate one, so I did just that, and added an extra template for it as well with a nice example.

Beside being an alternative way of specifying the schema, it also is more flexible due to the fact that we can use more of Zod's functionality, we can have nested objects, enums, ....

image
image
image

### DIFF
--- a/packages/components/nodes/outputparsers/StructuredOutputParserAdvanced/StructuredOutputParserAdvanced.ts
+++ b/packages/components/nodes/outputparsers/StructuredOutputParserAdvanced/StructuredOutputParserAdvanced.ts
@@ -46,9 +46,8 @@ class AdvancedStructuredOutputParser implements INode {
         "Action", "Comedy", "Drama", "Fantasy", "Horror",
         "Mystery", "Romance", "Science Fiction", "Thriller", "Documentary"
     ]).array().max(2), // Array of genres, max of 2 from the defined enum
-    shortDescription: z.string().max(500) // Short description, max 150 characters
-})`,
-                additionalParams: true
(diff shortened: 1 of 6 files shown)


packages/components/nodes/outputparsers/StructuredOutputParserAdvanced/StructuredOutputParserAdvanced.ts
```diff
@@ -0,0 +1,80 @@
...
+                default: `z.object({
+    title: z.string(), // Title of the movie as a string
+    yearOfRelease: z.number().int(), // Release year as an integer number,
+    genres: z.enum([
+        "Action", "Comedy", "Drama", "Fantasy", "Horror",
+        "Mystery", "Romance", "Science Fiction", "Thriller", "Documentary"
+    ]).array().max(2), // Array of genres, max of 2 from the defined enum
+    shortDescription: z.string().max(500) // Short description, max 150 characters
+})`,
+                additionalParams: true
```
HenryHengZJ, line 51: Think we can probably don't display as additional parameter. Would like to display it on the node itself, since user must enter the code specifying the zod schema

KennyVaneetvelde: Fair enough! Updated it

HenryHengZJ:
thanks @KennyVaneetvelde !

PR #1684, opened 2024-02-05
add kotlin files to folder-loader

ill-yes:
I noticed that Kotlin files were not loaded via the folder loader.
This PR now adds this file type.


HenryHengZJ (approved):
thanks!

PR #1687, opened 2024-02-06
Fix: relative links method and limit not applying to manage links

0xi4o:
- Fixes `relativeLinksMethod` and `limit` values set in Additional Parameters in `Cheerio`, `Playwright`, and `Puppeteer` nodes.


chungyau97:
The error that I found was in [Cheerio.ts Line 146](https://github.com/FlowiseAI/Flowise/blob/7faaf13ccc3705030a4dd87d05cd60e9f69772c3/packages/components/nodes/documentloaders/Cheerio/Cheerio.ts#L146).

Steps to reproduce error:
1. Set Relative Links Method `webCrawl` and Relative Links Limit to `10`
2. Fetch Links through Manage Links and Click Save
3. Set Relative Links Method `null` and Relative Links Limit to `2`
4. Save ChatFlow and Click Upsert
5. It will return 10 links instead of 2 

Change to loop this `selectedLinks.slice(0, limit)` instead of the whole selectedLinks

HenryHengZJ (approved):

chungyau97 (approved):

PR #1697, opened 2024-02-08
DynamoDB Chat Memory fix

### LINKED ISSUE #1624: [BUG] Dynamodb Chat Memory broken
**Describe the bug**
When attaching my existing and previously working dynamodb chat hisrory in the canva / getting an immediate error from the chat

Getting the following message:
Cannot convert undefined or null to object

**To Reproduce**
Attach the Dynanmodb chat history to Convesation Chain (any type of node would do)
When submitting question: Cannot convert undefined or null to object

1. Go to any canva with DynamoDB chat history
2. Click on question in the chat
3. See error

**Expected behavior**
Should roll back to functionning state

**Setup**

Did test on my 2 setups:
1 on render (last update - today)
1 on AWS with last container pull from yesterday
Using Chrome (did try all the refresh and di try on Firefox=same behaviour)
Both machines do run on linux
My local machines are on both linux and windows

Thanks!

Jaredude:
Fixes #1624

Please note comment in class BufferMemoryExtended for further discussion, if necessary

### DIFF
--- a/packages/components/nodes/memory/DynamoDb/DynamoDb.ts
+++ b/packages/components/nodes/memory/DynamoDb/DynamoDb.ts
@@ -148,9 +148,10 @@ interface DynamoDBSerializedChatMessage {
 }
 
 class BufferMemoryExtended extends FlowiseMemory implements MemoryMethods {
-    tableName = ''
-    partitionKey = ''
-    dynamoKey: Record<string, AttributeValue>
+    private tableName = ''
+    private partitionKey = ''
+    private dynamoKey: Record<string, AttributeValue>
+    private messageAttributeName: string
     sessionId = ''
     dynamodbClient: DynamoDBClient
 
@@ -221,9 +222,9 @@ class BufferMemoryExtended extends FlowiseMemory implements MemoryMethods {
     async getChatMessages(overrideSessionId = '', returnBaseMessages = false): Promise<IMessage[] | BaseMessage[]> {
         if (!this.dynamodbClient) return []
 
-        const dynamoKey = overrideSessionId ? this.overrideDynamoKey(overrideSessionId) : (this as any).dynamoKey
+        const dynamoKey = overrideSessionId ? this.overrideDynamoKey(overrideSessionId) : this.dynamoKey
         const tableName = this.tableName
-        const messageAttributeName = (this as any).messageAttributeName
+        const messageAttributeName = this.messageAttributeName
 
         const params: GetItemCommandInput = {
             TableName: tableName,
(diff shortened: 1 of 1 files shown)


Jaredude:
@HenryHengZJ This should fix #1624 I put a comment in packages/components/nodes/memory/DynamoDb/DynamoDb.ts on line 156 for further clarity. It might just be me not understanding the init flow of the BufferMemoryExtended class distinct from the init of the DynamoDb_Memory node. Any clarity around fields.chatHistory and fields.dynamodbClient is appreciated if you have time. I have tested this against an existing flow I have with DynamoDB memory, and it's working as expected. Sessions are being maintained, and overriding the sessionId is working as expected.

Screenshot 2024-02-07 at 8 32 21 PM
Screenshot 2024-02-07 at 8 33 04 PM

packages/components/nodes/memory/DynamoDb/DynamoDb.ts
```diff
@@ -142,18 +142,26 @@ interface DynamoDBSerializedChatMessage {
...
+    dynamoKey = ''
+    partitionKey = ''
 
     constructor(fields: BufferMemoryInput & BufferMemoryExtendedInput) {
         super(fields)
         this.sessionId = fields.sessionId
         this.dynamodbClient = fields.dynamodbClient
+
+        // These fields are coming in on chatHistory, but should they be on the dynamodbClient instead?
+        this.partitionKey = (fields?.chatHistory as unknown as { partitionKey: string }).partitionKey
```
HenryHengZJ, line 157: I'm not sure why its coming from `chatHistory`, I think a better way is to pass through the initilization of `BufferMemoryExtended` at line 121?

Jaredude: Thanks for pointing that out. I have b better understanding now. I'll update the PR...

Jaredude: During the init, chatHistory is assigned the value of dynamoDb, which is the result of DynamoDBChatMessageHistory. That's the traceback for why fields.chatHistory has all those attributes. Placed the pass through on the initialization of BufferMemoryExtended. Code is working as expected and much easier to follow.

HenryHengZJ: there are other places we are still using `(this as any).dynamoKey` or `partitionKey` or `tableName`, can we just do `this` as well?

Jaredude: You mean change it back to (this as any) instead of adding them as variables to the class? I had added them as variables to the class to avoid using (this as any), but I should have added them as private since they really aren't used outside the class.

HenryHengZJ: No I meant we probably can get rid of using `this as any` and just use `this` instead

Jaredude: Oh, man! I was tired when I was reading that comment last night. I see what you mean.... Yes, I'll update in other uses! 😀

Jaredude: @HenryHengZJ replaced all (this as any) usage in BufferMemoryExtended. Tested using session and override session again, and all chat threads continue to produce the expected results.

HenryHengZJ (approved):

packages/components/nodes/memory/DynamoDb/DynamoDb.ts
HenryHengZJ: tysm!

Issue #96, opened 2023-05-12
[BUG] PayloadTooLargeError: request entity too large


kristianjackson:
**Describe the bug**
Loading in a 48M PDF file is producing the following error:

Failed to save chatflow: <!DOCTYPE html> <html lang="en"> <head> <meta charset="utf-8"> <title>Error</title> </head> <body> <pre>PayloadTooLargeError: request entity too large<br> &nbsp; &nbsp;at readStream (/home/kristian/.nvm/versions/node/v18.12.1/lib/node_modules/flowise/node_modules/raw-body/index.js:156:17)<br> &nbsp; &nbsp;at getRawBody (/home/Kristian/.nvm/versions/node/v18.12.1/lib/node_modules/flowise/node_modules/raw-body/index.js:109:12)<br> &nbsp; &nbsp;at read (/home/Kristian/.nvm/versions/node/v18.12.1/lib/node_modules/flowise/node_modules/body-parser/lib/read.js:79:3)<br> &nbsp; &nbsp;at jsonParser (/home/Kristian/.nvm/versions/node/v18.12.1/lib/node_modules/flowise/node_modules/body-parser/lib/types/json.js:135:5)<br> &nbsp; &nbsp;at Layer.handle [as handle_request] (/home/Kristian/.nvm/versions/node/v18.12.1/lib/node_modules/flowise/node_modules/express/lib/router/layer.js:95:5)<br> &nbsp; &nbsp;at trim_prefix (/home/Kristian/.nvm/versions/node/v18.12.1/lib/node_modules/flowise/node_modules/express/lib/router/index.js:328:13)<br> &nbsp; &nbsp;at /home/Kristian/.nvm/versions/node/v18.12.1/lib/node_modules/flowise/node_modules/express/lib/router/index.js:286:9<br> &nbsp; &nbsp;at Function.process_params (/home/Kristian/.nvm/versions/node/v18.12.1/lib/node_modules/flowise/node_modules/express/lib/router/index.js:346:12)<br> &nbsp; &nbsp;at next (/home/Kristian/.nvm/versions/node/v18.12.1/lib/node_modules/flowise/node_modules/express/lib/router/index.js:280:10)<br> &nbsp; &nbsp;at expressInit (/home/Kristian/.nvm/versions/node/v18.12.1/lib/node_modules/flowise/node_modules/express/lib/middleware/init.js:40:5)</pre> </body> </html>

**To Reproduce**

[...]

PR #1707, opened 2024-02-10
add FLOWISE_FILE_SIZE_LIMIT variable

chungyau97:
Issue #96 

Allow easier way to setup Flowise file size limit.


packages/server/.env.example
```diff
@@ -20,6 +20,7 @@ PORT=3000
 # FLOWISE_USERNAME=user
 # FLOWISE_PASSWORD=1234
 # FLOWISE_SECRETKEY_OVERWRITE=myencryptionkey
+# FLOWISE_FILE_SIZE_LIMIT=50mb
```
HenryHengZJ, line 23: do a search on other variable name like `FLOWISE_USERNAME`, there are other places we need to add this env var too

HenryHengZJ (approved):

PR #1709, opened 2024-02-11
persist overrideConfig in chat_message table to support multi tenancy

niztal:
In order to support multi tenancy in flowise, meaning that a single flow may server multiple customers we can deliver the `overrideConfig` parameter to the `prediction`/`internal-prediction` APIs, so you can overload additional `variable` of `customer_id` (for example, or you can give whatever alias you would like to).

But with the current implementation and in order to support the predictions analytics/logs you won't be able to differ such `chat_message`s among your different tenants/customers.

Hence, I added this overrideConfig to be persisted which will bring us the ability to QUERY by your given `variable` (e.g. `customer_id`)

So for example, if your API would called with such `overrideConfig`:

image

You can later query that watching the final result persisted on DB:

image

### DIFF
--- a/packages/server/src/Interface.ts
+++ b/packages/server/src/Interface.ts
@@ -36,6 +36,7 @@ export interface IChatMessage {
     memoryType?: string
     sessionId?: string
     createdDate: Date
+    overrideConfig?: string
 }
 
 export interface ITool {
--- a/packages/server/src/database/entities/ChatMessage.ts
+++ b/packages/server/src/database/entities/ChatMessage.ts
@@ -40,4 +40,7 @@ export class ChatMessage implements IChatMessage {
 
     @CreateDateColumn()
     createdDate: Date
+
+    @Column({ nullable: true })
+    overrideConfig?: string
 }
--- a/packages/server/src/database/migrations/mysql/1702200925471-AddVariableEntity.ts
+++ b/packages/server/src/database/migrations/mysql/1702200925471-AddVariableEntity.ts
@@ -2,7 +2,7 @@ import { MigrationInterface, QueryRunner } from 'typeorm'
 
 export class AddVariableEntity1699325775451 implements MigrationInterface {
     public async up(queryRunner: QueryRunner): Promise<void> {
-        await queryRunner.query(
+                await queryRunner.query(
             `CREATE TABLE IF NOT EXISTS \`variable\` (
                 \`id\` varchar(36) NOT NULL,
                 \`name\` varchar(255) NOT NULL,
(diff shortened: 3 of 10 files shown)


niztal:
@HenryHengZJ @Jaredude FYI

Jaredude:
It's failing CI Build, but aside from that, perhaps the overrideConfig column should take advantage of database specific data types? Sqlite doesn't have a json data type, but MySQL does and Postgres supports the jsonb data tyle. These data types are more optimized for json querying than having to convert the text to json to use json querying.

chungyau97:
Hi @niztal,
image
you can pass in your custom `customer_id` value as `chatId`
Then you can query by `chatId` to identify your tenants/customers 

In our roadmap we will be adding a new table to store `customerInfo` which will have relationship with `chatId`. This will require your tenants/customers to enter their info before they are able to enter the embedChat that you shared.

Is this what you're trying to achieve?

niztal:
> It's failing CI Build, but aside from that, perhaps the overrideConfig column should take advantage of database specific data types? Sqlite doesn't have a json data type, but MySQL does and Postgres supports the jsonb data tyle. These data types are more optimized for json querying than having to convert the text to json to use json querying.

It's just a string column data type, as you can see on my commit I've made sure to handle all db types (postgres, MySQL, SQLite) . 

I don't have postgres besides that I've tested it on MySQL and SQLite and it worked fine.

I'll fix the ci failures.

Thanks @Jaredude

niztal:
> Hi @niztal, image you can pass in your custom `customer_id` value as `chatId` Then you can query by `chatId` to identify your tenants/customers
> 
> In our roadmap we will be adding a new table to store `customerInfo` which will have relationship with `chatId`. This will require your tenants/customers to enter their info before they are able to enter the embedChat that you shared.
> 
> Is this what you're trying to achieve?

Mainly yes, it sounds like a workaround, but for my use case I can use chatId.
Just one verification, the chatId can be overriden as part of the overrideConfig or just pass chatId as is to the prediction API?

Do you have eta regarding the customerInfo table?

@chungyau97 

Thanks

chungyau97:
This is my example last time:
```
{
    "chatId": "842a4db3-82b0-4d99-b213-b0df56ac540a",
    "question": "What's Jane?",
    "apiType": "embedded",
}
```
More Information:
Just think of `chatId` as a new chat started by your user
If `userA` started the chat, you can just pass in `chatId: userA`
One thing to make your `chatId` more unique is by combining the `userId_chatflowId`, this will prevent `userA` message get mix between different chatflow

Probably between 27Feb-5Mar, if everything went smooth.

Jaredude:
@chungyau97 Thanks for the input on this.

@niztal I found a bug where if you query the chatmessages API endpoint by chatId, the result is always an empty aray. #1716 Resolves that issue in case you encounter it too

niztal:
> @chungyau97 Thanks for the input on this.
> 
> @niztal I found a bug where if you query the chatmessages API endpoint by chatId, the result is always an empty aray. #1716 Resolves that issue in case you encounter it too

Thanks @chungyau97 @Jaredude 

I will close this PR.

PR #1716, opened 2024-02-13
fixes bug where querying by chatId results in no records

Jaredude:
previous logic for where clause of memoryType checked for existence of chatId and would query for memoryType is null This where clause logic results in an empty result for any request to the endpoint that filters by chatId

Getting messages without any filters
Screenshot 2024-02-12 at 8 14 32 PM

filtering by known chatId value using existing where clause
Screenshot 2024-02-12 at 8 16 09 PM

filtering by known chatId value using new where clause logic for memoryType
Screenshot 2024-02-12 at 8 15 09 PM

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -1453,7 +1453,7 @@ export class App {
                 chatflowid,
                 chatType,
                 chatId,
-                memoryType: memoryType ?? (chatId ? IsNull() : undefined),
+                memoryType: memoryType ?? undefined,
                 sessionId: sessionId ?? undefined,
                 createdDate: toDate && fromDate ? Between(fromDate, toDate) : undefined
             },
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -10,7 +10,7 @@ import logger from './utils/logger'
 import { expressRequestLogger } from './utils/logger'
 import { v4 as uuidv4 } from 'uuid'
 import OpenAI from 'openai'
-import { Between, IsNull, FindOptionsWhere } from 'typeorm'
+import { FindOptionsWhere, MoreThanOrEqual, LessThanOrEqual } from 'typeorm'
 import {
     IChatFlow,
     IncomingInput,
@@ -1464,7 +1464,8 @@ export class App {
                 chatId,
                 memoryType: memoryType ?? undefined,
                 sessionId: sessionId ?? undefined,
-                createdDate: toDate && fromDate ? Between(fromDate, toDate) : undefined
+                ...(fromDate && { createdDate: MoreThanOrEqual(fromDate) }),
+                ...(toDate && { createdDate: LessThanOrEqual(toDate) })
             },
             order: {
                 createdDate: sortOrder === 'DESC' ? 'DESC' : 'ASC'
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -1442,11 +1442,20 @@ export class App {
         startDate?: string,
         endDate?: string
     ): Promise<ChatMessage[]> {
+        const setDateToStartOrEndOfDay = (dateTimeStr: string, setHours: 'start' | 'end') => {
+            const date = new Date(dateTimeStr)
+            if (isNaN(date.getTime())) {
+                return undefined
+            }
+            setHours === 'start' ? date.setHours(0, 0, 0, 0) : date.setHours(23, 59, 59, 999)
+            return date
+        }
+
         let fromDate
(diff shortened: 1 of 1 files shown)


HenryHengZJ (approved):

packages/server/src/index.ts
```diff
@@ -1453,7 +1453,7 @@ export class App {
                 chatflowid,
                 chatType,
                 chatId,
-                memoryType: memoryType ?? (chatId ? IsNull() : undefined),
+                memoryType: memoryType ?? undefined,
```
HenryHengZJ, line 1456: couldn't remember why we have the `isNull()` in the first place, but would you be able to check from the View Message dialog on UI, that it still works as expected

Jaredude: It works as expected.
https://github.com/FlowiseAI/Flowise/assets/10228950/481bf246-e62a-4ec9-b91f-7ac73fd8c6b5

I found a bug with the way the chatType is filtered. I'll add another commit to this PR to resolve that issue as well.

Jaredude:
@HenryHengZJ It turned out that the bug I thought was related to the source type was not an issue with source type at all but rather an issue with the date filter. If an invalid date was passed in for startDate or endDate, getChatMessage was not checking whether new Date() was resulting in an invalid date or not. The date picker control on the start date is defaulting to a epoch timestamp that's not compatible with new Date(), so the bug only happens when neither of the date fields are changed.

Bug (causing server crash):
https://github.com/FlowiseAI/Flowise/assets/10228950/46116b2c-1d9f-4cc3-9936-94b40a2375ea

Fixed:
https://github.com/FlowiseAI/Flowise/assets/10228950/a18b3856-9410-45a3-86bb-71c23a2f398c

date filtering working with >= <= logic instead of between to handle more robust date filtering:
https://github.com/FlowiseAI/Flowise/assets/10228950/a13cfe9c-5600-47b2-a126-f397b149fcd4

HenryHengZJ:
@Jaredude I'm confused how to replicate the issue, it seems to be working for me when filter by dates:
image

Jaredude:
@HenryHengZJ I think the key to reproduce might be in live reload vs hot reload and the initial state of the date fields. Here's my steps to reproduce:
1. yarn build
2. yarn start
3. go into flow
4. view messages
5. Filter by type

The above steps will crash the server.

I am able to reliably reproduce using the steps above so long as I don't change the start date first. Let me know if you're unable to reproduce using the steps above, and I'll look at things again.
https://github.com/FlowiseAI/Flowise/assets/10228950/1b9f7a5b-282d-44a5-980f-4976fdc1982c

HenryHengZJ:
> @HenryHengZJ Here's my steps to reproduce:
> 
> 1. yarn build
> 2. yarn start
> 3. go into flow
> 4. view messages
> 5. Filter by type
> 
> The above steps will crash the server.
> 
> Changing the start date will not result in a crash, as it will provide a date in the proper format at that point. I am able to reliably reproduce using the steps above so long as I don't change the start date first. Let me know if you're unable to reproduce using the steps above, and I'll look at things again. [video example of server crash](https://github.com/FlowiseAI/Flowise/assets/10228950/1b9f7a5b-282d-44a5-980f-4976fdc1982c)
> 
> Edit: You can also cause a crash by passing an invalid date to the start date via the API.

Which branch are you on? I've tried multiple times on the `main` branch and never bumped into any issues. Filter by source works just fine

Jaredude:
I'm on main. I just did a pull, to make sure I'm up to date, and there were no files updated.

I thought the issue was filtering by source, but the issue is really with the handling of the date fields. Currently, we don't check whether new Date(startDate) or new Date(endDate) produce valid dates, so when an invalid date value gets passed into the .find method, the server will crash.

In my browser (Chrome or Safari), when I don't change the value for start date the call to new Date(startDate) is converting to "Invalid Date". This looks like it's a localization issue (possibly with the date picker library). But the server should be performing the date validation anyway since it implicitly knows the field must be a valid date to include it in the sql query.

Without changing the value of the start date the server will crash even by changing the end date and note even the type. [Video of server crash on end date change without start date change](https://github.com/FlowiseAI/Flowise/assets/10228950/4db8ab72-0b82-4826-8d7f-7bbab72e8c9f).

This crash can also be produced on Postman using any invalid values for either startDate or endDate.
Screenshot 2024-02-15 at 12 11 57 PM

HenryHengZJ (approved):

Jaredude:
Whew! 😅I was starting to worry something was wrong with my laptop!

PR #1717, opened 2024-02-13
Allows query chatmessage API endpoint by messageId

Jaredude:
This will be useful when the exact message is required by another system. For example, you want a microservice that performs additional work (search query caching, convert to voice, create a video, etc) and it need to be associated to that exact message in some sort of backend (likely graphql) that merges the messages with the augmented data.


HenryHengZJ:
Thanks @Jaredude !

PR #1733, opened 2024-02-15
Feature: Chat message feedback

0xi4o:
- Adds a dialog where chat feedback settings can be controlled.
- Adds feedback option to chat messages in embed - thumbs up or thumbs down. When one of these are clicked, an input dialog is show where users can leave written feedback.
- Shows feedback in `View Messages` dialog. 

flowise-chat-feedback-settings
flowise-chat-feedback-settings-2
flowise-chat-feedback-view-messages


packages/server/src/index.ts
```diff
@@ -1460,6 +1534,28 @@ export class App {
...
+                .where('chat_message.chatflowid = :chatflowid', { chatflowid })
+                .andWhere(chatType ? 'chat_message.chatType = :chatType' : 'TRUE', { chatType })
+                .andWhere(chatId ? 'chat_message.chatId = :chatId' : 'TRUE', { chatId })
+                .andWhere(memoryType ? 'chat_message.memoryType = :memoryType' : 'TRUE', {
+                    memoryType: memoryType ?? (chatId ? IsNull() : undefined)
+                })
+                .andWhere(sessionId ? 'chat_message.sessionId = :sessionId' : 'TRUE', {
+                    sessionId: sessionId ?? (chatId ? IsNull() : undefined)
+                })
+                .andWhere(fromDate && toDate ? 'chat_message.createdDate = :createdDate' : 'TRUE', {
```
HenryHengZJ, line 1550: filtering by date is not working anymore, please check.
image

packages/server/src/index.ts
```diff
@@ -1460,6 +1534,28 @@ export class App {
         let toDate
         if (endDate) toDate = setDateToStartOrEndOfDay(endDate, 'end')
 
+        if (feedback) {
+            const messages = await this.AppDataSource.getRepository(ChatMessage)
+                .createQueryBuilder('chat_message')
+                .leftJoinAndMapOne('chat_message.feedback', ChatMessageFeedback, 'feedback', 'feedback.messageId = chat_message.id')
+                .where('chat_message.chatflowid = :chatflowid', { chatflowid })
+                .andWhere(chatType ? 'chat_message.chatType = :chatType' : 'TRUE', { chatType })
```
HenryHengZJ, line 1542: question: what is the `TRUE` here?

packages/server/src/index.ts
```diff
@@ -599,10 +605,76 @@ export class App {
...
 
+            // remove all related feedback records
+            const feedbackDeleteOptions: FindOptionsWhere<ChatMessageFeedback> = { chatId }
+            await this.AppDataSource.getRepository(ChatMessageFeedback).delete(feedbackDeleteOptions)
+
             const results = await this.AppDataSource.getRepository(ChatMessage).delete(deleteOptions)
             return res.json(results)
         })
 
+        // ----------------------------------------
```
HenryHengZJ, line 616: @ocflowiseai also new endpoints that needs to be updated once its merged - https://github.com/FlowiseAI/Flowise/pull/1806

packages/server/src/index.ts
0xi4o: If a value doesn't exist, `TRUE` will return the value as if there wasn't a filter. Anyway, I've updated how the query is built for optional filters so it's clearer.

packages/server/src/index.ts
0xi4o: Fixed this issue.

packages/ui/src/ui-component/dialog/ViewMessagesDialog.js
```diff
@@ -449,7 +463,33 @@ const ViewMessagesDialog = ({ show, dialogProps, onCancel }) => {
...
+                        style={{
+                            display: 'grid',
+                            gridTemplateColumns: 'repeat(3, minmax(0, 1fr))',
+                            gap: 10,
+                            marginBottom: 16,
+                            marginLeft: 8,
+                            marginRight: 8
+                        }}
+                    >
+                        <StatsCard title='Total Messages (External)' stat={`${stats.totalMessages}`} />
```
HenryHengZJ, line 476: what does the (External) means?

HenryHengZJ (approved):

packages/ui/src/ui-component/dialog/ViewMessagesDialog.js
0xi4o: Embed since feedback is not available in the UI. I'll change the text.

saatchi-david:
This is really cool. Curious if there's an ETA, and if the data can be passed to Langsmith?

0xi4o:
@saatchi-david Should be available in the next release. We only have our own implementation at the moment. Passing the feedback to langsmith will come in a future release.

0xi4o:
@HenryHengZJ I've added the feedback functionality to the UI too. Also updated the `ViewMessagesDialog` - date and source filters will apply to stats too now.

Flowise-Low-code-LLM-apps-builder (7)
Flowise-Low-code-LLM-apps-builder (6)

saatchi-david:
Awesome, thanks @0xi4o!

packages/ui/src/ui-component/dialog/ChatFeedbackContentDialog.jsx
```diff
@@ -0,0 +1,76 @@
...
+    const onSave = () => {
+        onConfirm(feedbackContent)
+    }
+
+    useEffect(() => {
+        if (show) dispatch({ type: SHOW_CANVAS_DIALOG })
+        else dispatch({ type: HIDE_CANVAS_DIALOG })
+        return () => {
+            dispatch({ type: HIDE_CANVAS_DIALOG })
+            setFeedbackContent('')
```
HenryHengZJ, line 32: always remember to clear the state when we exit the dialog

HenryHengZJ (approved):

PR #1750, opened 2024-02-19
fix Dockerfile to actually start server

takaf3:
If you run the docker image built with current Dockerfile, container simply outputs the output of `flowise` command to its log and exit. This is due to `CMD` in Dockerfile just points `flowise` instead of `flowise start` which actually starts the server. Simple update on Dockerfile to make it actually start the server.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -18,4 +18,4 @@ RUN npm install -g flowise
 
 WORKDIR /data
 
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
make sure you have the `bin/dev` and `bin/run` files to have `LF` and not `CRLF`:
image

PR #1786, opened 2024-02-22
handles scenario where NodesPool require call fails

Jaredude:
Currently, the following file results in an error: Flowise/packages/components/dist/nodes/documentloaders/Subtitles/Subtitles.js:3:15 The result is that initializeNodes fails completely, and all other code in initDatabase fails to load


HenryHengZJ (approved):

HenryHengZJ:
thanks @Jaredude! yeah wrapping in try catch can at least let the data source initialiation finish

PR #1800, opened 2024-02-23
Allow default variable values ​​when testing CustomFunction

mokeyish:
Testing code is more cumbersome when CustomFunction requires input variables.

```js
const question = $question ?? 'my default question'

```

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -307,7 +307,10 @@ export class App {
         // execute custom function node
         this.app.post('/api/v1/node-custom-function', async (req: Request, res: Response) => {
             const body = req.body
-            const nodeData = { inputs: body }
+            const functionInputVariables = Object.fromEntries(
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
The pre-requisite is that you need to have the variables already specified in the UI, right?

mokeyish:
> The pre-requisite is that you need to have the variables already specified in the UI, right?

When CustomFunction needs to receive variables from chatflow, testing CustomFunction often reports errors. This can be forcibly defined as `undefined`. If the variable is undefined, the default value in the code will be used.

This is only a temporary fix for better testing CustomFunction.

- before

   ```js
   // const question = $question  // we need comment this line to avoid error.
   const question = "my default question"
   ```
   
- after

   ```js
   const question = $question ?? "my default question"
   ```

HenryHengZJ (approved):

PR #1814, opened 2024-02-26
Fix CustomFunction deserializing escaped JSON strings

mokeyish:
This PR relates to https://github.com/FlowiseAI/Flowise/pull/1539 . If it is json after unescaping, it will be parsed into a json object.

### DIFF
--- a/packages/components/nodes/utilities/CustomFunction/CustomFunction.ts
+++ b/packages/components/nodes/utilities/CustomFunction/CustomFunction.ts
@@ -90,13 +90,15 @@ class CustomFunction_Utilities implements INode {
 
         // Some values might be a stringified JSON, parse it
         for (const key in inputVars) {
-            if (typeof inputVars[key] === 'string' && inputVars[key].startsWith('{') && inputVars[key].endsWith('}')) {
+            value = handleEscapeCharacters(value, true)
+            if (value.startsWith('{') && value.endsWith('}')) {
                 try {
-                    inputVars[key] = JSON.parse(inputVars[key])
+                    value = JSON.parse(value)
                 } catch (e) {
-                    continue
+                    // ignore
                 }
             }
+            inputVars[key] = value
         }
 
         let sandbox: any = { $input: input }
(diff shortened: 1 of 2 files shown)


HenryHengZJ:
can you provide more description or elaboration what this PR trying to solve? screenshot/example of bug/errors/how to replicate the issue

mokeyish:
@HenryHengZJ This PR relates to #1539 . If it is json after unescaping, it will be parsed into a json object.

mokeyish:
图片

HenryHengZJ:
@mokeyish do you have some examples chatflow that I can replicate the issue?

mokeyish:
> @mokeyish do you have some examples chatflow that I can replicate the issue?

In my case it's LLMChain that outputs a JSON string, and then the next node is CustomFunction.

HenryHengZJ:
> > @mokeyish do you have some examples chatflow that I can replicate the issue?
> 
> In my case it's LLMChain that outputs a JSON string, and then the next node is CustomFunction.

does the JSON string contains the escaped characters like `FLOWISE_DOUBLE_QUOTE` ? I was just thinking any downside of handling the conversion of escaped charactes before JSON parsing it

mokeyish:
Yes, it contains `FLOWISE_DOUBLE_QUOTE`, so JSON parsing failed. then in the CustomFunction, it json string other than json object.

HenryHengZJ (approved):

HenryHengZJ:
We will have to do the same for `IfElseFunction.ts` as well

mokeyish:
> We will have to do the same for `IfElseFunction.ts` as well

Did you reply to the wrong place? did you mean #1832 ? IfElseFunction I understand that there are two EndNode branches, and which branch is decided based on the result.  But #1832 says reusing the calculation result of the previous node.

HenryHengZJ:
we have the same logic here - https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/utilities/IfElseFunction/IfElseFunction.ts#L104

PR #1820, opened 2024-02-27
Feature: Add support for accessing NodeData via json path

mokeyish:
This PR supports `{{<variableNodeId>.data.instance.pathtokey}}`.

图片

---

This is useful for PromptTemplate to get the value of one of its fields from the previous node json value output.

### DIFF
--- a/packages/server/src/utils/index.ts
+++ b/packages/server/src/utils/index.ts
@@ -541,7 +541,7 @@ export const getVariableValue = (
                 if (variableFullPathParts.length > 3) {
                     let variableObj = null
                     switch (typeof variableValue) {
-                        case 'string':
+                        case 'string': {
                             const unEscapedVariableValue = handleEscapeCharacters(variableValue, true)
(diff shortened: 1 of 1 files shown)


packages/server/src/utils/index.ts
```diff
@@ -528,11 +528,45 @@ export const getVariableValue = (
...
+            // 2: <variableNodeId>.data.instance.pathtokey
+            const variableFullPathParts = variableFullPath.split('.')
+            const variableNodeId = variableFullPathParts[0]
             const executedNode = reactFlowNodes.find((nd) => nd.id === variableNodeId)
             if (executedNode) {
-                const variableValue = get(executedNode.data, 'instance')
+                let variableValue = get(executedNode.data, 'instance')
+
+                // Handle path such as `<variableNodeId>.data.instance.key`
+                if (variableFullPathParts.length > 3) {
```
HenryHengZJ, line 541: should we check if the previous `data.instance` are present in the `variableFullPathParts` ?

mokeyish: Currently it seems to be all "data.instance".

PR #1832, opened 2024-02-28
Fix the logic error of `getStartingNodes`

mokeyish:
For the flow below, the original logic for calculating depth is wrong.

```mermaid
  graph TD;
      EndNode-->Y;
      Y-->Z;
      EndNode-->B;
      B-->C;
      C-->D;
      D-->Y;
```

Y will be skipped early, in fact, it has the greatest depth.

### DIFF
--- a/packages/server/src/utils/index.ts
+++ b/packages/server/src/utils/index.ts
@@ -162,45 +162,32 @@ export const constructGraphs = (
  * @param {string} endNodeId
  */
 export const getStartingNodes = (graph: INodeDirectedGraph, endNodeId: string) => {
-    const visited = new Set<string>()
-    const queue: Array<[string, number]> = [[endNodeId, 0]]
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
im not sure if such flow is allowed, the only way you can split the path is using If Else node, and even after splitting the path, they remain on seperate path and are not allowed to rejoin. do you have a working chatflow that shows this type of flow?

mokeyish:
The `Y&Z`, you can think it as the global initializaton/configuration(it very slow). It would be better to reuse the results and avoid double calculations.

It requires PR #1834.

HenryHengZJ:
> @HenryHengZJ Hi, what are your concerns about this PR? Currently this PR and #1834 are working fine in my local evironment.

sorry for late getting back. Left a comment on #1834

PR #1834, opened 2024-02-28
Fix duplicated calculation of nodes and dependent nodes are not initialized

mokeyish:
Same flow graph as #1832

### DIFF
--- a/packages/server/src/utils/index.ts
+++ b/packages/server/src/utils/index.ts
@@ -299,7 +299,7 @@ export const buildFlow = async (
         exploredNode[startingNodeIds[i]] = { remainingLoop: maxLoop, lastSeenDepth: 0 }
     }
 
-    const initializedNode: Set<string> = new Set()
+    const initializedNodes: Set<string> = new Set()
     const nonDirectedGraph = constructGraphs(reactFlowNodes, reactFlowEdges, { isReversed: true }).graph
     while (nodeQueue.length) {
         const { nodeId, depth } = nodeQueue.shift() as INodeQueue
@@ -386,7 +386,7 @@ export const buildFlow = async (
                 flowNodes[nodeIndex].data.instance = outputResult
 
                 logger.debug(`[server]: Finished initializing ${reactFlowNode.data.label} (${reactFlowNode.data.id})`)
-                initializedNode.add(reactFlowNode.data.id)
+                initializedNodes.add(reactFlowNode.data.id)
             }
         } catch (e: any) {
             logger.error(e)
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
@mokeyish the code make sense to avoid re-calculation of node once it has been initialized. However I'd like to double confirm if I can replicate this issue with a chatflow. So far from my testing, I've not come across with a chatflow that has this issue.

Do you have an example chatflow JSON template?

mokeyish:
My chatflow is more complicated, but I can take a screenshot.

图片

HenryHengZJ:
wow thats quite a complex one! from your chatflow, which nodes were re-calculated?

packages/server/src/utils/index.ts
```diff
@@ -299,6 +299,8 @@ export const buildFlow = async (
         exploredNode[startingNodeIds[i]] = { remainingLoop: maxLoop, lastSeenDepth: 0 }
     }
 
+    const initializedNodes: Set<string> = new Set()
+    const nonDirectedGraph = constructGraphs(reactFlowNodes, reactFlowEdges, { isReversed: true }).graph
```
HenryHengZJ, line 303: can we rename to `reversedGraph`?

packages/server/src/utils/index.ts
```diff
@@ -406,6 +409,8 @@ export const buildFlow = async (
         for (let i = 0; i < neighbourNodeIds.length; i += 1) {
             const neighNodeId = neighbourNodeIds[i]
             if (ignoreNodeIds.includes(neighNodeId)) continue
+            if (initializedNodes.has(neighNodeId)) continue
+            if (nonDirectedGraph[neighNodeId].some((dependId) => !initializedNodes.has(dependId))) continue
```
HenryHengZJ, line 413: I dont quite understand this line, why its it necessary? and can you walk me through a simple example

mokeyish:
The node that did the initialization repeated the calculation. In fact, it went to the database to read the entire configuration, and subsequent nodes will use the configuration it read.  Without this PR, it would read many times and be very slow.

packages/server/src/utils/index.ts
mokeyish: please look at this graph

```mermaid
  graph LR;
      Configration --> Initialization
      Initialization -- Reuse results--> EndNode
      Initialization --> A
      A --> B
      B --> C
      C --> EndNode
```

EndNode will be calculated via path from Initializaton to EndNode, at this time, C is unintialized.

So it would better to wait until from path C to EndNode.

packages/server/src/utils/index.ts
mokeyish: I copied it here.

https://github.com/FlowiseAI/Flowise/blob/13ee0d0c7b4c325100ec8dbd3919a26dd0db8893/packages/server/src/index.ts#L2025

HenryHengZJ: ah we should rename both as well, as its a reversed graph, not a non directed graph

packages/server/src/utils/index.ts
HenryHengZJ: Okay, for future reference:

When we arrive at `Initialization`, we have 2 neighbours: `A` and `EndNode`.

For `EndNode`:
- `reverseGraph[EndNode] = [Initialization, C]`
- `initializedNodes = [Initialization]`
- Since C is not part of `initializedNodes`, we skip processing `EndNode`

For `A`:
- `reverseGraph[A] = [Initialization]`
- `initializedNodes = [Initialization]`
- Continue processing A

HenryHengZJ (approved):

PR #1843, opened 2024-03-01
Fix the failure to resolve variables for OverrideConfig(Json Object)

mokeyish:
When defining chatflowConfig as below, `{{question}}` was not replaced with user input.

```js
Chatbot.initFull({
      chatflowid: 'e8c221b3-a47e-4c05-bcbc-86491b52bea9',
      apiHost: 'http://127.0.0.1:8711',
      isFullPage: true,
      chatflowConfig: {
        functionInputVariables: {
          topk: '7', // {{question}}
        }
      },
})
```

I found out that `getVariableValue` only handles strings. This PR solves the problem.


mokeyish:
@HenryHengZJ It seems that the failure of CI is not due to this code modification.

图片

HenryHengZJ:
You mean the `topK` provided below is not being captured?
```
Chatbot.initFull({
      chatflowid: 'e8c221b3-a47e-4c05-bcbc-86491b52bea9',
      apiHost: 'http://127.0.0.1:8711',
      isFullPage: true,
      chatflowConfig: {
        functionInputVariables: {
          topk: '7', // {{question}}
        }
      },
})
```

mokeyish:
`topk` is captured, it's no problem, but CustomFunction received `{{question}}`.

HenryHengZJ:
> `topk` is captured, it's no problem, but CustomFunction received `{{question}}`.

hey sorry for late getting back. what do you mean by CustomFunction received `{{question}}` ?

I tested the following chatflow:
image

```
{
    "question": "how are u",
    "overrideConfig": {
        "functionInputVariables": {
            "character": "sarcastic"
        }
    }
}
```

It is able to override the variable `character`

mokeyish:
Add `{question}` to `Format Prompt Values`, and then only overwrite `character`, CustomFunction will not be able to get the value of `{question}`.

mokeyish:
This is because `getVariableValue` receives paramValue of type object. `paramValue.length` returns undefined, so it will not be abble to resolve `{question}`.

HenryHengZJ:
> Add `{question}` to `Format Prompt Values`, and then only overwrite `character`, CustomFunction will not be able to get the value of `{question}`.

actually it was a bug on UI, the dialog was meant to be from the Input Variables from Custom Function:
image

HenryHengZJ (approved):

HenryHengZJ:
For future reference, the issue was:

When you have an overrideConfig which is a JSON type. For example: `functionInputVariables`, and you have `{{question}}` as one of the variables value:
image

The code does not replace `{{question}}`:
image

This fix is to stringify the JSON object, allowing `{{question}}` to be replaced correctly

HenryHengZJ:
@mokeyish can you pull the latest changes from main so we can have [this](https://github.com/FlowiseAI/Flowise/pull/1914) to let the CI run successfully

mokeyish:
sure

mokeyish:
@HenryHengZJ BTW,  It would be better to use `squash+rebase` when merging code into the main branch, so that you can have a clearer git history.

HenryHengZJ:
> @HenryHengZJ BTW, It would be better to use `squash+rebase` when merging code into the main branch, so that you can have a clearer git history.

thanks for the suggestion! will definitely take a look!

PR #1802, opened 2024-02-23
Feature/Indexing

HenryHengZJ:
Add indexing record manager components:

Screenshot 2024-02-24 013603

image

PR #1891, opened 2024-03-06
Fix for 1890 upsert returns text not JSON

### LINKED ISSUE #1890: [BUG] The upsert API examples assume a JSON response, but the response is not JSON so the examples fail when used
**Describe the bug**
If you open the upsert modal in a chatflow and click the 'Show API' button, the Python and JavaScript samples assume a JSON response:

```
import requests

API_URL = "http://localhost:3000/api/v1/vector/upsert/026f4b68-8a0b-418e-bd5d-99f3882a462f"

def query(payload):
    response = requests.post(API_URL, json=payload)
    return response.json()

output = query({
    "overrideConfig": {
        "basePath": "example",
        "topK": 1,
        "azureOpenAIApiKey": "example",
        "azureOpenAIApiInstanceName": "example",
    }
})
```

If you actually use this sample you'll see the following error:
`Uncaught (in promise) SyntaxError: Unexpected token 'S', "Successful"... is not valid JSON`

The reason is the response in code is not JSON, it's just a string (`upsertVector` in `index.ts`):
`return res.status(201).send('Successfully Upserted')`

**To Reproduce**

1. Open any chatflow and click the upsert icon, then 'Show API'
2. Copy & paste the JavaScript sample into the console and run it

**Expected behavior**
The sample should not assume a JSON response.

automaton82:
Fixes #1890 as upsert returns text, not JSON as the result.

### DIFF
--- a/packages/ui/src/views/vectorstore/VectorStoreDialog.js
+++ b/packages/ui/src/views/vectorstore/VectorStoreDialog.js
@@ -109,7 +109,7 @@ API_URL = "${baseURL}/api/v1/vector/upsert/${dialogProps.chatflowid}"
 
 def query(payload):
     response = requests.post(API_URL, json=payload)
-    return response.json()
+    return response.text
 
 output = query({
     ${isMultiple ? `"stopNodeId": "${vectorNodeId}",\n    ` : ``}"overrideConfig": {${getConfigExamplesForPython(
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
@automaton82 in this [PR](https://github.com/FlowiseAI/Flowise/pull/1802/files#diff-8e6dfbbfb522575fe7c568c2de518bf9de351de83c9dbb48eaa865cd54450eb3R1836) we will be returning the upsert result as a JSON.

Closing this for now

PR #1898, opened 2024-03-07
Claude 3 - Adding Support for Image Uploads / Multimodal Inputs

vinodkiran:
Claude 3 - Adding Support for Image Uploads / Multimodal Inputs


packages/components/src/IVisionChatModal.ts
```diff
@@ -0,0 +1,12 @@
+import { IMultiModalOption } from './Interface'
```
HenryHengZJ, line 1: Can we move this to `Interface.ts`?

packages/components/nodes/chatmodels/AzureChatOpenAI/AzureChatOpenAI.ts
```diff
@@ -1,8 +1,9 @@
-import { AzureOpenAIInput, ChatOpenAI, OpenAIChatInput } from '@langchain/openai'
+import { AzureOpenAIInput, ChatOpenAI as LangchainChatOpenAI, OpenAIChatInput } from '@langchain/openai'
```
HenryHengZJ, line 1: I'm getting access to Azure Vision model, in the meantime we can get rid of this first, and create another PR once we have access

HenryHengZJ (approved):

PR #1910, opened 2024-03-10
Fixed a bug about conversation dates

JoeABCDEF:
Fixed the issue of inconsistent date creation, if the running server and the database are in different regions, the generated date will be inconsistent, and the generated date will be inconsistent


HenryHengZJ:
hmm I'm not sure if we need to do this since the date is automatically generated when a record is created in database - https://github.com/FlowiseAI/Flowise/blob/main/packages/server/src/database/entities/ChatMessage.ts#L44

HenryHengZJ:
@JoeABCDEF can you pull the latest changes from main so we can have this [bugfix](https://github.com/FlowiseAI/Flowise/pull/1914) to let the CI run successfully

JoeABCDEF:
@HenryHengZJ I re-pulled

PR #1943, opened 2024-03-13
CONSTRAINT ID already in use - in postgresql migration - 1707213601923-AddFeedback.ts

matthias:
There is a conflict with AddVariableEntity1699325775451 which is using the same CONSTRAINT ID

I manually changed it. But there may be a better way to fix this (directly in the /entities/..) or when creating these migrations.


HenryHengZJ:
@matthias should be resolved with this [PR](https://github.com/FlowiseAI/Flowise/pull/1938)

matthias:
Thanks. Yes that's the same fix.

There are more issues for postgresql (supabase) with the new feedback feature. Will add a PR for that.

PR #1944, opened 2024-03-13
Fix: Faiss vector store throws error when using pnpm

0xi4o:
Fixes an error thrown by the Faiss node when using pnpm to install packages.


0xi4o:
@HenryHengZJ This fixes the error when upserting to Faiss vector store.

PR #1959, opened 2024-03-15
Fix#1947

emrahtoy:
This PR fixes 2 problem
1. type problem while calculating "a month ago", old code were return timestamp in numbers but it must be Date 
2. Migrations to the uuid and timestamp with time zone Entity types with migration file

This PR adds 2 files
1. util file for getting data source to generate typeorm migrations
2. Migration file for type changes 

This PR adds a npm command 
1. in Server package "typeorm:migration-generate" added to have easier typeorm data source with the help of new utility file


emrahtoy:
I forgat linting, will reopen soon.

PR #1946, opened 2024-03-13
Fix/supabase database conflicts introduced with ChatMessageFeedback

matthias:
Fixed issues for PostgreSQL (supabase) for ChatFeedback
- fixed foreign key (uuid) type mismatch
- fixed timestamp date/time field handling (value out of range / date conversion)
- added debug logging for database
- added sourcemap support for /server

Using supabase (postgresql) there where some issues introduced. 

To be able to track them down I added debugging support for TypeORM (if DEBUGU=true)
and sourcemap support for /server (as a dev dependency)

See logs:

Issue 1) Foreign key (uuid) type mismatch
```
flowise:dev: QueryFailedError: operator does not exist: character varying = uuid
flowise:dev:     at PostgresQueryRunner.query 

flowise:dev: query failed: SELECT "chat_message"."id" AS "chat_message_id", "chat_message"."role" AS "chat_message_role", "chat_message"."chatflowid" AS "chat_message_chatflowid", "chat_message"."content" AS "chat_message_content", "chat_message"."sourceDocuments" AS "chat_message_sourceDocuments", "chat_message"."usedTools" AS "chat_message_usedTools", "chat_message"."fileAnnotations" AS "chat_message_fileAnnotations", "chat_message"."fileUploads" AS "chat_message_fileUploads", "chat_message"."chatType" AS "chat_message_chatType", "chat_message"."chatId" AS "chat_message_chatId", "chat_message"."memoryType" AS "chat_message_memoryType", "chat_message"."sessionId" AS "chat_message_sessionId", "chat_message"."createdDate" AS "chat_message_createdDate", "feedback"."id" AS "feedback_id", "feedback"."chatflowid" AS "feedback_chatflowid", "feedback"."chatId" AS "feedback_chatId", "feedback"."messageId" AS "feedback_messageId", "feedback"."rating" AS "feedback_rating", "feedback"."content" AS "feedback_content", "feedback"."createdDate" AS "feedback_createdDate" FROM "chat_message" "chat_message" LEFT JOIN "chat_message_feedback" "feedback" ON "feedback"."messageId" = "chat_message"."id" WHERE "chat_message"."chatflowid" = $1 AND "chat_message"."chatType" = $2 AND "chat_message"."createdDate" BETWEEN $3 AND $4 ORDER BY "chat_message"."createdDate" ASC -- PARAMETERS: ["7a6a33cf-512b-4130-8f24-30ee699202f0","INTERNAL",1707857744097,"2024-03-13T20:55:44.097Z"]

[...]

PR #1960, opened 2024-03-15
Fix#1947

emrahtoy:
This PR fixes 2 problem
1. type problem while calculating "a month ago", old code were return timestamp in numbers but it must be Date 
2. Migrations to the uuid and timestamp with time zone Entity types with migration file

This PR adds 2 files
1. util file for getting data source to generate typeorm migrations
2. Migration file for type changes 

This PR adds a npm command 
1. in Server package "typeorm:migration-generate" added to have easier typeorm data source with the help of new utility file


emrahtoy:
NOTE : This PR does not include Sqllite migration.

HenryHengZJ:
coicidentally this [PR](https://github.com/FlowiseAI/Flowise/pull/1946) also fixing the same issue

HenryHengZJ:
@emrahtoy here's a list of supported types - https://orkhan.gitbook.io/typeorm/docs/entities#column-types

type `uuid` supports
- [x] postgres
- [x] mysql
- [ ] sqlite

type `timestamp with time zone` supports
- [x] postgres
- [ ] mysql
- [ ] sqlite

I tested sqlite it seems to work, maybe because of the `Init.ts` migration script so it ignores the column types specified in entities. Have you tried on mysql?

emrahtoy:
> @emrahtoy here's a list of supported types - https://orkhan.gitbook.io/typeorm/docs/entities#column-types
> 
> type `uuid` supports
> 
> * [x]  postgres
> * [x]  mysql
> * [ ]  sqlite
> 
> type `timestamp with time zone` supports
> 
> * [x]  postgres
> * [ ]  mysql
> * [ ]  sqlite
> 
> I tested sqlite it seems to work, maybe because of the `Init.ts` migration script so it ignores the column types specified in entities. Have you tried on mysql?

uuid becomes varchar find default that is why sqlite doesn't care about it.
timestamp with timezone type is not necessary as it is handled on the code side by returning actual date instead of numbers.

I haven't tested it on MySQL as it does not exist in my stack. 

It seems like both PR are almost identical.

emrahtoy:
> > > @emrahtoy here's a list of supported types - https://orkhan.gitbook.io/typeorm/docs/entities#column-types
> > > type `uuid` supports
> > > 
> > > * [x]    postgres
> > > * [x]    mysql
> > > * [ ]    sqlite
> > > 
> > > type `timestamp with time zone` supports
> > > 
> > > * [x]    postgres
> > > * [ ]    mysql
> > > * [ ]    sqlite
> > > 
> > > I tested sqlite it seems to work, maybe because of the `Init.ts` migration script so it ignores the column types specified in entities. Have you tried on mysql?
> > 
> > 
> > uuid becomes varchar find default that is why sqlite doesn't care about it. timestamp with timezone type is not necessary as it is handled on the code side by returning actual date instead of numbers.
> > I haven't tested it on MySQL as it does not exist in my stack.
> > It seems like both PR are almost identical but mine has changed all the entities.
> 
> okay, in that case can we get rid of the `timestamp with time zone` as its not necessary and I'm afraid we will bump into issues supporting mysql.

Yes for sure we can use `timestamp`

emrahtoy:
I have just changed `timestamp with timezone` sections to `timestamp`

HenryHengZJ:
@emrahtoy i dont think `timestamp` is supported for sqlite

emrahtoy:
> @emrahtoy i dont think `timestamp` is supported for sqlite. Actually do we really need the column type there?

I believe giving a type is important. It gives the opportunity of being future proof. 
Decorators has been handling different depending on the driver you pick ( MySQL, sqlite etc.. ) with this way TypeOrm know what to do with the value and does the necessary conversions.

Even if it doesn't give support at all we are still able to create our own custom handlers.

HenryHengZJ (approved):

HenryHengZJ:
thanks @emrahtoy !

neliocaa:
when changing the table to uuid, whoever is upgrading receives an error, since they already have data in another format, especially when it comes externally,
So, changing the chatId and sessionId fields to uuid doesn't make sense, and causes a bug in the upgrade

emrahtoy:
In my case, they were always UUID. @HenryHengZJ which one of these chatId and sessionId is actually a user variable?

Issue #1247, opened 2023-11-17
It takes ages to load chatflows view


zeitderforschung:
**Describe the bug**
Since the the chat history feature is available it takes several minutes to load the Chatflows view. The /api/v1/chatflows endpoint returns ca. 30mb of data, although we only have a dozen chatflows configured. We use the chatbot in production and the chat history / view messages tab contains a lot of conversations.

**To Reproduce**
Load Chatflows view.

**Expected behavior**
Chatflow view should load instantly and history shouldn't be included.

Is there a way to disable the chat history completely? BTW we also use the langsmith integration.

HenryHengZJ:
The reason of the laggy performance is because of the document loaders that you've uploaded with your file. If you do a little experiment and remove the document loaders, you might find it load the chatflows fast.

Flowise currently save all the uploaded documents as base64 string to your chatflow. We will have an update to let users store to a local path and that will resolve the issues.

PR #1976, opened 2024-03-17
Feature/externalize files from chatflow - do not save as base64

vinodkiran:
Currently Flowise saves all uploaded docs as base64 string to the chatflow. It is not scalable with large documents. Issue - https://github.com/FlowiseAI/Flowise/issues/1247

**Solution**: save the files to a local path and only save the filepath to the chatflow.

This PR extracts the base64 string and saves it to the path specified in env.BLOB_STORAGE_PATH (creates a sub-folder with the chatflowid)

This is backwards compatible, when a current chatflow (with base64 in flowData) is opened and saved, the files are extracted and saved.


HenryHengZJ:
if you do a search `type: 'file'`, there are other places that allow uploading files as well

vinodkiran:
> if you do a search `type: 'file'`, there are other places that allow uploading files as well

Yes, working on the rest of the item. Draft PR to ensure that the loaders work as expected with this change.

vinodkiran:
All nodes with **`type: 'file'`** are refactored, ready for review and merge.

HenryHengZJ (approved):

Issue #845, opened 2023-08-29
[BUG] CTRL + Backspace Deletes Nodes


vexersa:
**Describe the bug**
The common key combination of CTRL + Backspace to remove a word closes dialogue windows and removes nodes when there are no words left to remove (although it seems this is sometimes ignored). I think that this is a bad UX pattern which should be removed or modified to respect the classic/expected behaviour of removing the previous word, and when there are no more words to remove, does nothing.

**To Reproduce**

1. Open a configuration window for a node
2. Type a word or two
3. Press CTRL + Backspace to remove the previously typed words, when there are no words left to remove (although it seems this is sometimes ignored), the dialogue window will close and the node will be removed. 

**Expected behavior**
The dialogue window does not close, and the node is not removed. Words are removed according to typical CTRL + Backspace behaviour.

**Screenshots**

https://github.com/FlowiseAI/Flowise/assets/7994227/8598c058-c451-4d76-92a8-4d9c6c36fb94

**Flow**
If applicable, add exported flow in order to help replicating the problem.

**Setup**
-   Installation : Docker
-   Flowise Version : 1.3.4
-   OS: Windows
-   Browser: Chrome

**Additional context**

HenryHengZJ:
we've removed the shortcut backspace for deleting nodes - https://github.com/FlowiseAI/Flowise/pull/1983 , it should resolve the issue

PR #2008, opened 2024-03-21
fix: render the dialog only if it open

thvroyal:
Fix the issue: https://github.com/FlowiseAI/Flowise/issues/845


HenryHengZJ:
we've removed the shortcut backspace for deleting nodes - https://github.com/FlowiseAI/Flowise/pull/1983 , it should resolve the issue

PR #2018, opened 2024-03-21
feature/fetch openaiKey via overrideConfig (vars)

niztal:
In order to support multi-tenancy and having the ability that a single chatFlow will support multiple customers with each customer potentially providing their own OpenAI key, we can implement the "customer_id" overrideConfig (as shown in the screenshot below). In this setup, the Flowise server will attempt to locate the corresponding credential in the Credentials table based on the provided `customer_id`. The `customer_id` supplied should match the name of the credential in the table.

For instance, if we use this embed code:

```
<script type="module">
        import Chatbot from "https://cdn.jsdelivr.net/npm/flowise-embed/dist/web.js"
        Chatbot.init({
            chatflowid: "e683f528-4f5f-4cf8-8f5b-25f7802c7de1",
            apiHost: "http://localhost:3000",
            chatflowConfig: {
                vars: {
                    customer_id: 'test'
                }
            }
        })
    </script>
```

and assuming I have an openai key credential already persisted by the name of "test" (for example in the below screenshot):

image

image

Then flowise server will decrypt this row's value and use this openai key.


niztal:
@HenryHengZJ please review

Jaredude:
@niztal What's the result if a user passes in a customer_id value that's not in the credentials table?

Also, what prevents someone from changing the overrideconfig value and using someone else's customer_id, which would allow them to use someone else's API Key? Implemented as is, appears to me, would create a vulnerability into Flowise.

HenryHengZJ:
yeah I agree with @Jaredude, we can't tie the logic of retrieving credential to values set in `vars`. A better approach would be just allow users to pass in as part of overrideConfig:
```
<script type="module">
        import Chatbot from "https://cdn.jsdelivr.net/npm/flowise-embed/dist/web.js"
        Chatbot.init({
            chatflowid: "e683f528-4f5f-4cf8-8f5b-25f7802c7de1",
            apiHost: "http://localhost:3000",
            chatflowConfig: {
                openAIApiKey: "abc"
            }
        })
    </script>
```

niztal:
> yeah I agree with @Jaredude, we can't tie the logic of retrieving credential to values set in `vars`. A better approach would be just allow users to pass in as part of overrideConfig:
> 
> ```
> <script type="module">
>         import Chatbot from "https://cdn.jsdelivr.net/npm/flowise-embed/dist/web.js"
>         Chatbot.init({
>             chatflowid: "e683f528-4f5f-4cf8-8f5b-25f7802c7de1",
>             apiHost: "http://localhost:3000",
>             chatflowConfig: {
>                 openAIApiKey: "abc"
>             }
>         })
>     </script>
> ```

If user will provide the openai API key as plain text it won't be safe , rather than store it encrypted on flowise and fetch it by some unique id.

@HenryHengZJ

niztal:
> @niztal What's the result if a user passes in a customer_id value that's not in the credentials table?
> 
> Also, what prevents someone from changing the overrideconfig value and using someone else's customer_id, which would allow them to use someone else's API Key? Implemented as is, appears to me, would create a vulnerability into Flowise.

If user will provide non existing customer_id, the fallback would be to the provided  API key over the node @HenryHengZJ

HenryHengZJ:
> > yeah I agree with @Jaredude, we can't tie the logic of retrieving credential to values set in `vars`. A better approach would be just allow users to pass in as part of overrideConfig:
> > ```
> > <script type="module">
> >         import Chatbot from "https://cdn.jsdelivr.net/npm/flowise-embed/dist/web.js"
> >         Chatbot.init({
> >             chatflowid: "e683f528-4f5f-4cf8-8f5b-25f7802c7de1",
> >             apiHost: "http://localhost:3000",
> >             chatflowConfig: {
> >                 openAIApiKey: "abc"
> >             }
> >         })
> >     </script>
> > ```
> 
> If user will provide the openai API key as plain text it won't be safe , rather than store it encrypted on flowise and fetch it by some unique id.
> 
> @HenryHengZJ

in that case, are you dynamically generating new credential for every user? does that means if you have 1000 users, you will have 1000 credentials?

niztal:
> > > > yeah I agree with @Jaredude, we can't tie the logic of retrieving credential to values set in `vars`. A better approach would be just allow users to pass in as part of overrideConfig:
> > > > ```
> > > > <script type="module">
> > > >         import Chatbot from "https://cdn.jsdelivr.net/npm/flowise-embed/dist/web.js"
> > > >         Chatbot.init({
> > > >             chatflowid: "e683f528-4f5f-4cf8-8f5b-25f7802c7de1",
> > > >             apiHost: "http://localhost:3000",
> > > >             chatflowConfig: {
> > > >                 openAIApiKey: "abc"
> > > >             }
> > > >         })
> > > >     </script>
> > > > ```
> > > 
> > > 
> > > If user will provide the openai API key as plain text it won't be safe , rather than store it encrypted on flowise and fetch it by some unique id.
> > > @HenryHengZJ
> > 
> > 
> > in that case, are you dynamically generating new credential for every user? does that means if you have 1000 users, you will have 1000 credentials?
> 
> No,
> 
> I won't dynamically generate new credentials for every user.
> 
> Each customer has its own API key, and will get charge to openai by his usage.
> 
> Imagine one customer use 10 tokens a month and another one will use 10K tokens a month they will get charged accordingly. If I'll have 1000 customers I will have 1000 credentials, yes. I don't see it as a problem, SQL will query that easily from the Credentials table.
> 
> Moreover I really think moving a variable named "openAiKey" will be vulnerable since it will be visible publicly to everyone, where "customer_id" is less explicit.
> 
> WDYT? @Jaredude @HenryHengZJ

Guys,

I feel there are few options here, please let me know what's your favorite:

1. Moving the `openAiKey` as plain text - I think this is really bad since each user can view it via the browser's network and use it for his own purposes.
2. Instead of `openAiKey` move `credentialId` which will represent the customer's persisted credentialId on the DB and change [line](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/chatmodels/ChatOpenAI/ChatOpenAI.ts#L224) to look for this as well from the `nodeData.inputs?.credentialId` From there it will work the same no other changes needed, the key is not exposed and each customer has its own credential's uuid 
3. Move the `openAiKey` encrypted and change [line](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/src/utils.ts#L524) to call decrypt function if the param was taken from the `nodeData.inputs?[paramName]`
4. Use my original solution and move another implicit input (e.g. id) and use this to search it over the db (very similar to the 2nd solution)


@Jaredude @HenryHengZJ

Thanks

HenryHengZJ:
So if we go with solution 2, you'll have to implement your logic to retrieve the credential ID for the customer and pass in as:

```
<script type="module">
        import Chatbot from "https://cdn.jsdelivr.net/npm/flowise-embed/dist/web.js"
        Chatbot.init({
            chatflowid: "e683f528-4f5f-4cf8-8f5b-25f7802c7de1",
            apiHost: "http://localhost:3000",
            chatflowConfig: {
                "credentialId": {
                    "chatOpenAI_0": "abc123"
                }
            }
        })
</script>
```

Then we can implement the logic here:
```
if (nodeData.inputs?.credentialId) {
    nodeData.credential = nodeData.inputs?.credentialId
}
const credentialData = await getCredentialData(nodeData.credential ?? '', options)
const openAIApiKey = getCredentialParam('openAIApiKey', credentialData, nodeData)
```

niztal:
> So if we go with solution 2, you'll have to implement your logic to retrieve the credential ID for the customer and pass in as:
> 
> ```
> <script type="module">
>         import Chatbot from "https://cdn.jsdelivr.net/npm/flowise-embed/dist/web.js"
>         Chatbot.init({
>             chatflowid: "e683f528-4f5f-4cf8-8f5b-25f7802c7de1",
>             apiHost: "http://localhost:3000",
>             chatflowConfig: {
>                 "credentialId": {
>                     "chatOpenAI_0": "abc123"
>                 }
>             }
>         })
> </script>
> ```
> 
> Then we can implement the logic in `ChatOpenAI.ts`:
> 
> ```
> if (nodeData.inputs?.credentialId) {
>     nodeData.credential = nodeData.inputs?.credentialId
> }
> const credentialData = await getCredentialData(nodeData.credential ?? '', options)
> const openAIApiKey = getCredentialParam('openAIApiKey', credentialData, nodeData)
> ```

sure @HenryHengZJ 

can you please elaborate further what's the `chatOpenAI_0` you mentioned?

Thanks

niztal:
> > So if we go with solution 2, you'll have to implement your logic to retrieve the credential ID for the customer and pass in as:
> > ```
> > <script type="module">
> >         import Chatbot from "https://cdn.jsdelivr.net/npm/flowise-embed/dist/web.js"
> >         Chatbot.init({
> >             chatflowid: "e683f528-4f5f-4cf8-8f5b-25f7802c7de1",
> >             apiHost: "http://localhost:3000",
> >             chatflowConfig: {
> >                 "credentialId": {
> >                     "chatOpenAI_0": "abc123"
> >                 }
> >             }
> >         })
> > </script>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then we can implement the logic in `ChatOpenAI.ts`:
> > ```
> > if (nodeData.inputs?.credentialId) {
> >     nodeData.credential = nodeData.inputs?.credentialId
> > }
> > const credentialData = await getCredentialData(nodeData.credential ?? '', options)
> > const openAIApiKey = getCredentialParam('openAIApiKey', credentialData, nodeData)
> > ```
> 
> sure @HenryHengZJ
> 
> can you please elaborate further what's the `chatOpenAI_0` you mentioned and why does the `credentialId` over the `chatflowConfig` is an object and not a string?
> 
> Thanks

@HenryHengZJ 

I think it's wrong, while running your approach I'm getting the following error:

> flowise:dev: query failed: SELECT `Credential`.`id` AS `Credential_id`, `Credential`.`name` AS `Credential_name`, `Credential`.`credentialName` AS `Credential_credentialName`, `Credential`.`encryptedData` AS `Credential_encryptedData`, `Credential`.`createdDate` AS `Credential_createdDate`, `Credential`.`updatedDate` AS `Credential_updatedDate` FROM `credential` `Credential` WHERE ((`Credential`.`id` = ?)) LIMIT 1 -- PARAMETERS: [{"chatOpenAI_0":"abc123"}]
> flowise:dev: error: Error: ER_BAD_FIELD_ERROR: Unknown column 'chatOpenAI_0' in 'where clause'
> flowise:dev: 2024-03-25 22:08:48 [ERROR]: QueryFailedError: ER_BAD_FIELD_ERROR: Unknown column 'chatOpenAI_0' in 'where clause'

I think the embed code should be as following:

`><script type="module">
 >       import Chatbot from "https://cdn.jsdelivr.net/npm/flowise-embed/dist/web.js"
 >     Chatbot.init({
 >        chatflowid: "e683f528-4f5f-4cf8-8f5b-25f7802c7de1",
 >            apiHost: "http://localhost:3000",
 >           chatflowConfig: {
 >               "credentialId": "abc123" //The customer's credential uuid in the DB
 >           }
 >       })
 > </script>`

niztal:
> > So if we go with solution 2, you'll have to implement your logic to retrieve the credential ID for the customer and pass in as:
> > ```
> > <script type="module">
> >         import Chatbot from "https://cdn.jsdelivr.net/npm/flowise-embed/dist/web.js"
> >         Chatbot.init({
> >             chatflowid: "e683f528-4f5f-4cf8-8f5b-25f7802c7de1",
> >             apiHost: "http://localhost:3000",
> >             chatflowConfig: {
> >                 "credentialId": {
> >                     "chatOpenAI_0": "abc123"
> >                 }
> >             }
> >         })
> > </script>
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > Then we can implement the logic in `ChatOpenAI.ts`:
> > ```
> > if (nodeData.inputs?.credentialId) {
> >     nodeData.credential = nodeData.inputs?.credentialId
> > }
> > const credentialData = await getCredentialData(nodeData.credential ?? '', options)
> > const openAIApiKey = getCredentialParam('openAIApiKey', credentialData, nodeData)
> > ```
> 
> sure @HenryHengZJ
> 
> can you please elaborate further what's the `chatOpenAI_0` you mentioned and why does the `credentialId` over the `chatflowConfig` is an object and not a string?
> 
> Thanks

@HenryHengZJ I think you missed this:

`
if (nodeData.inputs?.credentialId && nodeData.inputs?.credentialId[nodeData.id]) {
    nodeData.credential = nodeData.inputs?.credentialId[nodeData.id]
}...
`

niztal:
@HenryHengZJ it's works like a charm ✨ 

Please review, highly appreciated 🙏

HenryHengZJ (approved):

HenryHengZJ:
Glad to hear!

niztal:
@HenryHengZJ 

can you please tell when this PR is going to be released?

Thanks

PR #2026, opened 2024-03-24
Bugfix/Postgres migration column type

HenryHengZJ:


schieck0:
Thank you very much, thanks to you I spent 2 days trying to identify a bug "invalid input syntax for type uuid" to discover that someone changed the field type in a migration instead of creating another.
[issue](https://github.com/FlowiseAI/Flowise/issues/3014)

otalekar:
I am trying to setup Postgres db for Flowise and keep getting this error on the startup

`[ERROR]: :x: [server]: Error during Data Source initialization: relation "UQ_6352078b5a294f2d22179ea7956" already exists`

Do you think we should add this command before adding the constraint?

`ALTER TABLE "chat_message_feedback" DROP CONSTRAINT IF EXISTS "UQ_6352078b5a294f2d22179ea7956";`

Did anyone face this problem?

PR #2055, opened 2024-03-27
chore: upgrade langfuse to support metadata and tags

marcklingen:


marcklingen:
still used yarn, migrated change to pnpm lock

HenryHengZJ (approved):

PR #2101, opened 2024-04-04
Chore/consistent services and error handlers

ocflowiseai:
Implement consistent services and error definitions

### DIFF
--- a/packages/server/src/controllers/apikey/index.ts
+++ b/packages/server/src/controllers/apikey/index.ts
@@ -28,10 +28,10 @@ const createApiKey = async (req: Request, res: Response, next: NextFunction) => 
 // Update api key
 const updateApiKey = async (req: Request, res: Response, next: NextFunction) => {
     try {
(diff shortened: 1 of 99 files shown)


packages/server/src/controllers/apikey/index.ts
```diff
@@ -61,9 +61,6 @@ const verifyApiKey = async (req: Request, res: Response, next: NextFunction) =>
             new ApiError(StatusCodes.PRECONDITION_FAILED, `Error: apikeyController.verifyApiKey - apiKey not provided!`)
```
HenryHengZJ, line 61: should we `throw new ApiError` or just `throw new Error`? Since the other controllers are just throwing Error

ocflowiseai: We now have the same error throwing mechanism in the entire server package

0xi4o (approved):

0xi4o (approved):

HenryHengZJ (approved):

PR #2132, opened 2024-04-09
Feature: Add BasePath parameter setting support to ChatOpenAI LlamaIndex

louyongjiu:
Add BasePath parameter setting support to ChatOpenAI LlamaIndex


HenryHengZJ (approved):
thanks @louyongjiu !

PR #2133, opened 2024-04-09
Optimize `getEndingNodes`

mokeyish:
If there are multiple end nodes, only one needs to pass verification, and the others are automatically ignored. 

This allows us to temporarily disconnect a node without deleting it.

### DIFF
--- a/packages/server/src/services/chatflows/index.ts
+++ b/packages/server/src/services/chatflows/index.ts
@@ -41,40 +41,20 @@ const checkIfChatflowIsValidForStreaming = async (chatflowId: string): Promise<a
         const edges = parsedFlowData.edges
         const { graph, nodeDependencies } = constructGraphs(nodes, edges)
 
-        const endingNodeIds = getEndingNodes(nodeDependencies, graph)
-        if (!endingNodeIds.length) {
-            throw new InternalFlowiseError(StatusCodes.NOT_FOUND, `Ending nodes not found`)
-        }
-
-        const endingNodes = nodes.filter((nd) => endingNodeIds.includes(nd.id))
+        const endingNodes = getEndingNodes(nodeDependencies, graph, nodes)
 
         let isStreaming = false
-        let isEndingNodeExists = endingNodes.find((node) => node.data?.outputs?.output === 'EndingNode')
-
         for (const endingNode of endingNodes) {
             const endingNodeData = endingNode.data
-            if (!endingNodeData) {
-                throw new InternalFlowiseError(StatusCodes.NOT_FOUND, `Ending node ${endingNode.id} data not found`)
-            }
-
             const isEndingNode = endingNodeData?.outputs?.output === 'EndingNode'
-
-            if (!isEndingNode) {
-                if (
-                    endingNodeData &&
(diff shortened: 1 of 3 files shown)


HenryHengZJ:
You mean this allow us to disconnect one of the ending nodes without deleting it? But if you disconnect it, its no longer part of the endingNodes calculation no? Do you have a flow that can better showcase the issue?

mokeyish:
When testing the prompt, the examples(for few-shot prompting) is no longer needed. ChatPromptTemplate node needs to disconnect the CustomFunction node(fetch the exmaples for vector database) so that it will no longer be executed.

The current version will consider this CustomFunction to be an endingNode and prompt an error `Ending node must be either a Chain or Agent`.

packages/server/src/utils/buildChatflow.ts
```diff
@@ -187,45 +187,21 @@ export const utilBuildChatflow = async (req: Request, socketIO?: Server, isInter
...
-                        return {
-                            executionError: true,
-                            status: 500,
-                            msg: `Ending node must be either a Chain or Agent`
-                        }
-                    }
-
                     if (
                         endingNodeData.outputs &&
                         Object.keys(endingNodeData.outputs).length &&
```
mokeyish, line 207: @HenryHengZJ I'm not sure if this needs to be moved to "getEndingNodes" as well

packages/server/src/services/chatflows/index.ts
```diff
@@ -41,35 +41,14 @@ const checkIfChatflowIsValidForStreaming = async (chatflowId: string): Promise<a
...
-        }
-
-        const endingNodes = nodes.filter((nd) => endingNodeIds.includes(nd.id))
+        const endingNodes = getEndingNodes(nodeDependencies, graph, nodes)
 
         let isStreaming = false
-        let isEndingNodeExists = endingNodes.find((node) => node.data?.outputs?.output === 'EndingNode')
+        let isEndingNodeExists = endingNodes.some((node) => node.data?.outputs?.output === 'EndingNode')
 
         for (const endingNode of endingNodes) {
```
HenryHengZJ, line 49: Can we change the naming from `isEndingNodeExists` to `isCustomFunctionEndingNode`? this can better reflects the purpose of this check.

I think once `isEndingNodeExists`, we dont have to run the `for` loop below right? we can directly return `{ isStreaming : false }`

packages/server/src/utils/buildChatflow.ts
HenryHengZJ: @mokeyish we can leave it here, because its not required when checking if chatflow is streamable

HenryHengZJ (approved):

HenryHengZJ:
this looks much neater! thanks a lot @mokeyish !

While only making check to get one validation for now, in future we should implement a validation on every component nodes level, instead of the `getEndingNodes` function at server level.

llattarini1982:
Hello guys, I am getting this issue after the commit of this:failed to solve: process "/bin/sh -c pnpm build" did not complete successfully: exit code: 1-

Any suggestion how to solve it?
Thanks

PR #2151, opened 2024-04-10
fix: add migration to modify chatflowconfig to longtext for mysql

sickOscar:
Using big CSV files as document loaders my break mysql chat flow because of the TEXT size of the column and base64 conversion.

Here a migration has been added to change the column type from TEXT to LONGTEXT (4GB of max data)

### DIFF
--- a/packages/server/src/database/migrations/mysql/1712762777471-ExpandChatflowConfig.ts
+++ b/packages/server/src/database/migrations/mysql/1712762777471-ExpandChatflowConfig.ts
@@ -1,0 +1,11 @@
+import { MigrationInterface, QueryRunner } from 'typeorm'
+
+export class ExpandChatFlowConfig1712762777471 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE \`chat_flow\` MODIFY \`chatbotConfig\` LONGTEXT;`)
+    }
+
+    public async down(queryRunner: QueryRunner): Promise<void> {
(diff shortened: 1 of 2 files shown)


HenryHengZJ:
@sickOscar thanks for the fix! However with this [PR](https://github.com/FlowiseAI/Flowise/pull/1976) merged and released, everything is now stored under a local file path instead of a base64 string, hence this fix is no longer needed

HenryHengZJ:
having said that, if you still have a lot of nodes and using MySQL, you might still face the issue. See https://github.com/FlowiseAI/Flowise/issues/2159.

So we can still proceed with this change

packages/server/src/database/migrations/mysql/1712762777471-ExpandChatflowConfig.ts
```diff
@@ -0,0 +1,11 @@
+import { MigrationInterface, QueryRunner } from 'typeorm'
+
+export class ExpandChatFlowConfig1712762777471 implements MigrationInterface {
+    public async up(queryRunner: QueryRunner): Promise<void> {
+        await queryRunner.query(`ALTER TABLE \`chat_flow\` MODIFY \`chatbotConfig\` LONGTEXT;`)
```
HenryHengZJ, line 5: is it `chatbotConfig` that should be changed? isnt it `flowData`?

seonghobae: I think that’s May purpose of prevention of exceptions.

HenryHengZJ:
closing as https://github.com/FlowiseAI/Flowise/pull/2908 merged

PR #2113, opened 2024-04-06
Feature/externalize model list

vinodkiran:
Moving all model names to an external configuration (JSON) file.

PR #2165, opened 2024-04-12
feat: add gpt-4-trubo (new updated model by OpenAI)

AsharibAli:
**Add: New "GPT-4-Turbo" with Vision Model in the ChatOpenAI Node.**

Yesterday, they have released a stable version of gpt-4-turbo series with a context window size of 128,000 tokens and updated training data up to December 2023.

**Official Release Reference:**
#[https://platform.openai.com/docs/models/continuous-model-upgrades](https://platform.openai.com/docs/models/continuous-model-upgrades)


HenryHengZJ:
thanks @AsharibAli! we have a working PR here - https://github.com/FlowiseAI/Flowise/pull/2113, it moves models name into a JSON file, so next time there is a new model, all we need to do is just update the JSON file

AsharibAli:
Hmm, understood.

PR #2182, opened 2024-04-14
Unable to view upsert history dialog, when InMemoryVector Store is used

vinodkiran:
Steps to reproduce:
1) Create a new chatflow with InMemory VectorStore and do an upsert.
2) Click on Upsert History Dialog from settings
3) Expand Accordion - Details 
4) Expand the InMemory Accordion


HenryHengZJ:
good catch, because the `"paramValues":[]` is empty

HenryHengZJ (approved):

PR #2205, opened 2024-04-18
Feature: Add pdfUsage parameter setting support to folderFiles

louyongjiu:
Folder with Files
By default, a separate document is created for each page in the PDF file. 
This setting can be modified to create one document per PDF file.

### DIFF
--- a/packages/components/nodes/documentloaders/Folder/Folder.ts
+++ b/packages/components/nodes/documentloaders/Folder/Folder.ts
@@ -47,6 +47,22 @@ class Folder_DocumentLoaders implements INode {
                 optional: true
             },
             {
+                label: 'Pdf Usage',
+                name: 'pdfUsage',
+                type: 'options',
+                options: [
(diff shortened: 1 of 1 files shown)


packages/components/nodes/documentloaders/Folder/Folder.ts
```diff
@@ -46,6 +46,22 @@ class Folder_DocumentLoaders implements INode {
...
+                    {
+                        label: 'One document per page',
+                        name: 'perPage'
+                    },
+                    {
+                        label: 'One document per file',
+                        name: 'perFile'
+                    }
+                ],
+                default: 'perPage'
```
HenryHengZJ, line 63: can we add `additionalParams: true`?

HenryHengZJ: also we'll have to increment the version at the top whenever we add a new parameter

PR #2211, opened 2024-04-19
Feature: Add pdfUsage parameter setting support to folderFiles

louyongjiu:
Folder with Files 
By default, a separate document is created for each page in the PDF file. 
This setting can be modified to create one document per PDF file


HenryHengZJ (approved):

HenryHengZJ:
thanks @louyongjiu !

Issue #2191, opened 2024-04-15
[BUG]Error: 400 Bad Request: Payload error: JSON payload (95879532 bytes) is larger than allowed (limit: 33554432 bytes).


prithvi151080:
**Describe the bug**
Cannot upsert big documents with more than 100k words. if the document is in docx format get the error "Error: 400 Bad Request: Payload error: JSON payload (95879532 bytes) is larger than allowed (limit: 33554432 bytes)." 

If the same file is converted pdf or txt file than get this error "Error: Undefined error code fetch failed: undefined"

In all these cases the vector collection is being created in the vector DB but the collection is empty. 

**To Reproduce**
Its a locally hosted environment using mxbai_large as the embedding model and tinyllama as the LLM. Qdrant is being used as the vector DB. 
However, smaller document documents gets embedded and upserted successfully whether they are docx/pdf/txt file using the same workflow without any error

**Expected behavior**
successful upsert of document

**Screenshots**
docx upsert error:
![1](https://github.com/FlowiseAI/Flowise/assets/157370999/9784d4ef-4841-4110-baa6-8407ce02a720)
![3](https://github.com/FlowiseAI/Flowise/assets/157370999/e843e693-cf55-4058-ac72-26a2184312a9)

pdf upsert error:
![2](https://github.com/FlowiseAI/Flowise/assets/157370999/26411fdd-d7a6-462a-afa1-1ed825b04086)

txt upsert error:
![4-text](https://github.com/FlowiseAI/Flowise/assets/157370999/652e0c36-4adc-4ae4-bbde-f90adc307ca7)

mxbai_embedding logs:
![image](https://github.com/FlowiseAI/Flowise/assets/157370999/a5944c25-f99d-41b1-bfce-df1d18f0b5ff)

**Flow**
[pdf extract vector Chatflow.json](https://github.com/FlowiseAI/Flowise/files/14981867/pdf.extract.vector.Chatflow.json)

**Setup**

-   Installation: Docker 25.0.3
-   Flowise Version 1.6.0 linux-x64 node-v18.19.1
-   OS: windows
-   Browser chrome

**Additional context**
Even if the upsert process fails the embedding llm does undertake the process of embedding...its only when the entire document is embedded that this error comes.
additionally getting this warning [WARNING]: Importing from "langchain/schema" is deprecated.Instead, please import from the appropriate entrypoint in "@langchain/core" or "langchain".This will be mandatory after the next "langchain" minor version bump to 0.2.

[...]

Issue #1824, opened 2024-02-27
[BUG] Error: Error: 400 Bad Request: Payload error: JSON payload (37327342 bytes) is larger than allowed (limit: 33554432 bytes).


vipervs:
Im trying to embed a vector db (quadrant) with langchain docs with sitemap.xml link extraction.

everything works fine and the embedding starts, but after a while I get this error:

Error: Error: 400 Bad Request: Payload error: JSON payload (37327342 bytes) is larger than allowed (limit: 33554432 bytes).

is it possible to increase the JSON payload so it can manage all links?

<img width="1650" alt="image" src="https://github.com/FlowiseAI/Flowise/assets/33532138/32100f01-ef1c-4e9a-b515-a892a12122e3">

HenryHengZJ:
Is the error coming from Qdrant not able to handle too large of a payload? Or its from Flowise side that is not able to pass the payload over to Qdrant?

mrabbah:
I got the same error, I think the qdrant client must send the data in batch mode, and I think we must be able to customize the batch_size in the qdrant node

prithvi151080:
@vipervs: solution found for the problem. plz check https://github.com/FlowiseAI/Flowise/issues/2191

xmaiconx:
Increase QDrant limit size is not a solution. I am trying to upload 211kb csv file and I am getting the same error. 
Very strange behaviour.

prithvi151080:
@xmaiconx: increase the max file upload size for Qdrant in the custom config.yaml for Qdrant as per ur use case...this is not a limitation from flowise but Qdrant...i faced problem with a txt file of less than 1 mb. use the config.yaml for configuring Qdrant to accept big files. size of the source file is not the issue but the no of vectors that are being generated from the content of the file that is being pushed in Qdrant in a single go is the bottleneck...Qdrant by default accepts max 32 MB (33554432 bytes) in a single go...increase that size as per ur requirement and the problem goes away.

HenryHengZJ:
should be fixed via [PR](https://github.com/FlowiseAI/Flowise/pull/2215), closing for now

PR #2215, opened 2024-04-19
Add batchSize parameter to Qdrant_VectorStores and QdrantUpsert_Vecto…

mrabbah:
Resolve the bug related to #2191 and #1824 :
Error: 400 Bad Request: Payload error: JSON payload (xxxxxx bytes) is larger than allowed (limit: 33554432 bytes)


packages/components/nodes/vectorstores/Qdrant/Qdrant.ts
```diff
@@ -257,7 +265,10 @@ class Qdrant_VectorStores implements INode {
 
                     return res
                 } else {
-                    await QdrantVectorStore.fromDocuments(finalDocs, embeddings, dbConfig)
+                    for(let i=0; i<finalDocs.length; i+=batchSize) {
```
HenryHengZJ, line 268: if no `batchSize` is found, can we just do it without for loop?

I tried upserting 1k+ chunks, it works fine:
image

Instead of doing this 10 times in for loop, I think its better for other to specify the batch size once they faced the error, if everything works fine for the first try, then thats great

packages/components/nodes/vectorstores/Qdrant/Qdrant.ts
```diff
@@ -86,6 +86,13 @@ class Qdrant_VectorStores implements INode {
                 default: 1536,
                 additionalParams: true
             },
+            {
+                label: 'Batch Size',
+                name: 'batchSize',
+                type: 'number',
+                default: 100,
+                additionalParams: true
```
HenryHengZJ, line 94: can we have an `optional: true`

xmaiconx:
This is a very important fix..

mrabbah:
In my last commit the batchSize is not required, and I try to POST all at once first and if it fail I fall back to batch mode

HenryHengZJ (approved):
thanks guys for the PR, merging!

PR #2238, opened 2024-04-23
feat(ui): auto proxy backend

mokeyish:
This PR removes hard code backend port and automatically reads the port from server's env file.

### DIFF
--- a/packages/ui/src/store/constant.js
+++ b/packages/ui/src/store/constant.js
@@ -4,10 +4,7 @@ export const drawerWidth = 260
 export const appDrawerWidth = 320
 export const headerHeight = 80
 export const maxScroll = 100000
-export const baseURL =
-    import.meta.env.PROD === true
-        ? window.location.origin
-        : window.location.origin.replace(`:${import.meta.env.VITE_PORT ?? '8080'}`, ':3000')
+export const baseURL = window.location.origin
 export const uiBaseURL = window.location.origin
 export const FLOWISE_CREDENTIAL_ID = 'FLOWISE_CREDENTIAL_ID'
 export const REDACTED_CREDENTIAL_VALUE = '_FLOWISE_BLANK_07167752-1a71-43b1-bf8f-4f32252165db'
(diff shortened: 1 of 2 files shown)


nstankov-bg:
I second on this one. It was a weird hoop to jump through for such a low barrier of entry project.

HenryHengZJ (approved):
thank you @mokeyish !

packages/ui/src/store/constant.js
```diff
@@ -4,10 +4,7 @@ export const drawerWidth = 260
 export const appDrawerWidth = 320
 export const headerHeight = 80
 export const maxScroll = 100000
-export const baseURL =
-    import.meta.env.PROD === true
-        ? window.location.origin
-        : window.location.origin.replace(`:${import.meta.env.VITE_PORT ?? '8080'}`, ':3000')
+export const baseURL = window.location.origin
 export const uiBaseURL = window.location.origin
```
HenryHengZJ, line 8: @mokeyish just noticed there are 2 urls:
`baseURL` and `uiBaseURL` with same value `window.location.origin`, do we need both?

Also when running `pnpm dev` on `localhost:8080`, the `baseURL` becomes `localhost:8080`. This makes the `ChatMessage.jsx` not able to receive streaming because the server is `localhost:3000` - https://github.com/FlowiseAI/Flowise/blob/main/packages/ui/src/views/chatmessage/ChatMessage.jsx#L609

mokeyish: This may be because vite does not support proxy SSE or because it only proxy the path of /api/?

HenryHengZJ: perhaps it doenst support socket proxy. we probably need to get the server port and replace it `const serverPort = parseInt(dotenv.config({ processEnv: {}, path: '../server/.env' }).parsed?.['PORT'])`

mokeyish: I think the possible reason is that vite only proxy the `api/` prefix.


图片

HenryHengZJ: @mokeyish do you have any fix for that?

mokeyish: No, I'm on a business trip recently, so it's not convenient.

mokeyish: Fix by #2396

PR #2282, opened 2024-04-29
Fixed Decision Flow

Vertirax:
- the fixed decision flow is useful to define a predetermined flow - you add the tool and attach choices to it, which then gets matched to the AI's input once your tool is called
- added FixedDecisionTool and FixedDecisionChoice with icons
- added the returned text to the setMessages since the returned answer wasn't updated, probably because it wasn't sent through the websocket


Vertirax:
Hi guys,

I've implemented a simple fixed decision flow.
image

There was an issue with the frontend not setting the latest message's content, most likely because of the `returnDirect`, please pay extra attention to reviewing `packages/ui/src/views/chatmessage/ChatMessage.jsx` (that solved it, it only showed the tool usage). I've put the Choice node into the Tools, since I wasn't sure it would fit into Utilities either.

I've attached a simple example use case export: [Fixed Decision Flow Simple Example Chatflow.json](https://github.com/FlowiseAI/Flowise/files/15150216/Fixed.Decision.Flow.Simple.Example.Chatflow.json)
(using the return direct and setting the buffer window memory size to 0 is highly advised)

HenryHengZJ:
Thanks for the suggestion @Vertirax !

I was thinking of how do we make it work regardless of the chain/agent, because right now it relies on the tool agent. The goal of this is to allow user to skip calling LLM if certain input question is asked, and we match that to a set of predefined output. To do that, we need a type of sequential flow approach. We're currently working on that feature. For now, will keep this PR in draft mode first until we have the new sequential flow approach feature

PR #2284, opened 2024-04-29
Upstash Vector Integration

fahreddinozcan:
This PR adds Upstash Vector integration as a vector store.


packages/components/credentials/UpstashVectorApi.credential.ts
```diff
@@ -0,0 +1,29 @@
...
+	description: string
+	inputs: INodeParams[]
+
+	constructor() {
+		this.label = 'Upstash Vector API'
+		this.name = 'upstashVectorApi'
+		this.version = 1.0
+		this.inputs = [
+			{
+				label: 'UPSTASH_VECTOR_REST_URL',
```
HenryHengZJ, line 16: Can we change the label to `Upstash Vector REST URL`. Same with `Upstash Vector REST Token`

packages/components/nodes/vectorstores/Upstash/Upstash.ts
```diff
@@ -0,0 +1,181 @@
...
+				name: 'document',
+				type: 'Document',
+				list: true,
+				optional: true
+			},
+			{
+				label: 'Embeddings',
+				name: 'embeddings',
+				type: 'Embeddings',
+				optional: true
```
HenryHengZJ, line 59: `embeddings` is a must no? if so, we have to get rid of the `optional: true`

packages/components/nodes/vectorstores/Upstash/Upstash.ts
```diff
@@ -0,0 +1,181 @@
...
+			index: upstashIndex
+		}
+
+		if (upstashMetadataFilter) {
+			obj.filter = upstashMetadataFilter
+		}
+
+		const vectorStore = await UpstashVectorStore.fromExistingIndex(embeddings, obj)
+
+		return resolveVectorStoreOrRetriever(nodeData, vectorStore)
```
HenryHengZJ, line 177: can we also add `obj.filter`:

`return resolveVectorStoreOrRetriever(nodeData, vectorStore, obj.filter)`

HenryHengZJ:
would be great if you can pull the latest changes and resolve the pnpm lock conflict 👍

packages/components/nodes/vectorstores/Upstash/Upstash.ts
HenryHengZJ: ahh i see didnt know that, thanks!

fahreddinozcan:
Hey, I removed the optional field from the embeddings for now, since langchain integration for Upstash Embeddings is not released. We can just proceed without Upstash Embeddings now, and once it's available I'll create another PR for that.

HenryHengZJ:
@fahreddinozcan can we do a `yarn lint-fix`?

HenryHengZJ (approved):
thank you @fahreddinozcan !

PR #2300, opened 2024-05-01
FIX Web Crawler doesn't include search params in URLs

ahmosman:
There is an issue with Puppeter, Playwright and Cherioo Web Scrappers, they doesn't include search params from URLs.

E. g. I'd like to scrap URL https://wiki.minthcm.org/index.php?title=Process:Running_on_Docker and in te result this URL which is scrapped is truncated to https://wiki.minthcm.org/index.php. Current implemetation truncates all the parameters after "?".

Provided fix will cover this issue.

Here is the flow I was using
web-crawler-flow


HenryHengZJ:
awesome thanks @ahmosman !

for future reference:
`const urlObj = new URL('https://example.org/abc?123')`

`urlObj.hostname = example.org`

`urlObj.pathname= /abc`

`urlObj.search= ?123`

HenryHengZJ (approved):

PR #2328, opened 2024-05-05
models.json support for Groq

ml4den:
Added support for `models.json` with GroqChat and included the current models in the list.


HenryHengZJ (approved):
thank you!

Issue #2343, opened 2024-05-07
[BUG] CSV document loader marks text splitter as optional, but it fails if it's not specified


automaton82:
**Describe the bug**
If you use the CSV document loader node with a CSV file and **do not add a text splitter** (which seems to be okay, it's marked as `optional`), you'll get the following error:

```
2024-05-07 00:26:21 [ERROR]: [server]: Error: TypeError: splitter.splitDocuments is not a function
Error: TypeError: splitter.splitDocuments is not a function
```

**To Reproduce**

1. Add a CSV document loader and any sort of vector storage
2. Use ChatOpenAI and a conversational retrieval QA node
3. Do not add a text splitter
4. Try to ask a question

**Expected behavior**
It should work. If you add a text splitter it will then work properly.

**Screenshots**
![image](https://github.com/FlowiseAI/Flowise/assets/4467711/9db6e23a-44fc-4b00-8b8e-46b068720a82)

**Setup**

-   Installation`pnpm start`
-   Flowise Version latest as of https://github.com/FlowiseAI/Flowise/commit/a84eabbef2c53a36a90e11084af071f1f54d7398
-   OS: Windows 11
-   Browser all

**Additional context**
This looks like a typo in the document loader. I can submit a PR to fix

automaton82:
Submitted PR #2344 to fix this issue.

PR #2344, opened 2024-05-07
Fixes 2343 CSV error with no text splitter

automaton82:
This fixes the typo where the CSV document loader doesn't work if a text splitter isn't specified.


automaton82:
Should fix #2343

HenryHengZJ (approved):

PR #2360, opened 2024-05-08
fix initDatabase function by proper use of await in try catch

saurabhg22:
The initDatabase function in App class was not awaiting for its completion. Added a await keyword and refactored the catch in try catch block.


HenryHengZJ (approved):
thanks @saurabhg22 !

Issue #2341, opened 2024-05-07
[BUG] CSV agent is not working with OpenAI ChatGPT


automaton82:
**Describe the bug**
If you attempt to use the CSV agent with the attached public csv city data file, you get the following error:

Question: `what cities are there`

Response:
```
{'LatD': 'int64', ' "LatM"': 'int64', ' "LatS"': 'int64', ' "NS"': 'object', ' "LonD"': 'int64', ' "LonM"': 'int64', ' "LonS"': 'int64', ' "EW"': 'object', ' "City"': 'object', ' "State"': 'object'}
2024-05-06 19:54:32 [ERROR]: [server]: Error: Sorry, I'm unable to find answer for question: "what cities do you have" using following code: "df["City"].unique()"
Error: Sorry, I'm unable to find answer for question: "what cities do you have" using following code: "df["City"].unique()"
    at CSV_Agents.run (C:\git\flowise\packages\components\dist\nodes\agents\CSVAgent\CSVAgent.js:165:23)
    at async utilBuildChatflow (C:\git\flowise\packages\server\dist\utils\buildChatflow.js:232:15)
    at async createInternalPrediction (C:\git\flowise\packages\server\dist\controllers\internal-predictions\index.js:7:29)
```

All questions will fail though.

CSV file: [cities.csv](https://github.com/FlowiseAI/Flowise/files/15227621/cities.csv)

**To Reproduce**

1. Create a chatflow and add a `CSV agent` node
2. Use OpenAI ChatGPT `Chat Model` or `LLM` node, it's the same result for both
3. Upload the attached csv file
4. Ask "what cities are there"

**Expected behavior**
This should work.

**Setup**

-   Installation `pnpm start`
-   Flowise Version latest as of https://github.com/FlowiseAI/Flowise/commit/ef532866fde6b32305eadf076a220926c3c732a6
-   OS: Windows 11
-   Browser all

**Additional context**
Aside from this problem, it also appears the uploaded file does not go into DISK BLOB storage. It goes into the database as base64. This behaviour is different than the CSV document loader which does go into DISK BLOB storage immediately when uploaded. Not sure if this is intended or not.

HenryHengZJ:
OpenAI suddenly gives the prefix `python`, so the whole code becomes:

```
import pandas as pd
```python
print(df["Company Name"])
` ` `
```
which causes error when trying to run on pyodide
[...]

Issue #2364, opened 2024-05-08
[BUG] Airtable Agent: Only returning a maximum of 100 records


toi500:
**Describe the bug**
I am facing a bug when the aitable agent is limited to read up to 100 results, no matter how you configure its parameters.

Also, some times it gives the following error:

<img width="353" alt="image" src="https://github.com/FlowiseAI/Flowise/assets/138339291/e1e12288-9c72-4aae-8651-fbd17dcc825b">

**To Reproduce**
1. Connect an airtable with more than 100 entries, lets say 500.
2. Configure it for "return all".
3. Ask the agent how many records there are in the database
4. The response will be 100

**Screenshots**
<img width="457" alt="image" src="https://github.com/FlowiseAI/Flowise/assets/138339291/2a9af923-084a-489b-8ef9-337bcca02a0c">
<img width="645" alt="image" src="https://github.com/FlowiseAI/Flowise/assets/138339291/56035e96-f0aa-4f2b-98a9-a83c8a1fbe1d">

**Setup**
-   Flowise Version is up tp date

**Additional context**
It happens with all models, chatgpt, gemini, etc.

**Dummy data**
[customers.csv](https://github.com/FlowiseAI/Flowise/files/15250247/customers.csv)

PR #2393, opened 2024-05-12
Fixed CSV agent not working with OpenAI GPT

Hariharan1828:
Pull Request Description:

This pull request addresses an issue #2341 encountered with the CSV agent in Flowise when processing CSV data containing prefixed column names. The issue manifested as an error preventing successful retrieval of information from the CSV file, specifically when querying for city data.

Changes Made:

    Regex Function Implementation:
        Introduced a removePrefix function utilizing a regular expression pattern to remove any prefixes from column names. The regex pattern effectively targets and removes prefixes added by OpenAI to the column names in the CSV data.

    Map Function Application:
        Applied the removePrefix function to each node's label property using the map function. This ensures that any prefixes present in the labels of the starting nodes are removed before they are added to the activeChatflows object.

    Logger Info Enhancement:
        Added logging statements to provide visibility into the addition and removal of chatflows from the ChatflowPool. These statements aid in tracking the execution flow and monitoring the status of chatflows.

Expected Outcome:

With these changes, the prefixes from column names in the CSV data will be effectively removed, enabling the CSV agent to accurately process the data without encountering errors. This should resolve the issue described in the bug report, allowing for successful retrieval of city data from CSV files.

If there are any further questions or adjustments needed, please feel free to discuss them in the comments.

### DIFF
--- a/packages/server/src/ChatflowPool.ts
+++ b/packages/server/src/ChatflowPool.ts
@@ -17,10 +17,21 @@ export class ChatflowPool {
      * @param {ICommonObject} overrideConfig
      */
     add(chatflowid: string, endingNodeData: INodeData | undefined, startingNodes: IReactFlowNode[], overrideConfig?: ICommonObject) {
+        const removePrefix = (colName: string) => colName.replace(/^"[^"]*"/, ''); // Regex to remove prefix
+
+        const updatedStartingNodes = startingNodes.map(node => ({
+            ...node,
+            data: {
+                ...node.data,
+                label: removePrefix(node.data.label),
+                // Add more properties if needed
+            },
+        }));
+
         this.activeChatflows[chatflowid] = {
-            startingNodes,
+            startingNodes:updatedStartingNodes,
             endingNodeData,
-            inSync: true
+            inSync: true,
         }
         if (overrideConfig) this.activeChatflows[chatflowid].overrideConfig = overrideConfig
         logger.info(`[server]: Chatflow ${chatflowid} added into ChatflowPool`)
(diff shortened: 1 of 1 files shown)


Hariharan1828:
Opps, there is a lint error, let me fix

packages/server/src/ChatflowPool.ts
```diff
@@ -17,8 +17,19 @@ export class ChatflowPool {
      * @param {ICommonObject} overrideConfig
      */
     add(chatflowid: string, endingNodeData: INodeData | undefined, startingNodes: IReactFlowNode[], overrideConfig?: ICommonObject) {
+        const removePrefix = (colName: string) => colName.replace(/^"[^"]*"/, '') // Regex to remove prefix
```
HenryHengZJ, line 20: thanks for the suggested fix!

why do we have the fix here though? I thought it should be in the `CSVAgent.ts` line 178 to check if the `pythonCode` contains any prefix

Hariharan1828: I thought of changing it directly from the Pool because not only CSV AGENT AirtableAgent also had same issue, but yes I do find changing from CSV AGENT is precise, will revert back and do he checks in both CSV and AirtableAgent

toi500: Any news about this fix?

toi500:
@Hariharan1828 @HenryHengZJ any news about this fix?

Also, the airtable node has been bugged for a while now https://github.com/FlowiseAI/Flowise/issues/2364

prakashcs:
Can anyone tell how to use it? I am uploading a normal CSV file and tried using both Open AI and Azure Open AI Chat with GPT 4o but getting strange errors

HenryHengZJ:
Closing as https://github.com/FlowiseAI/Flowise/pull/2983 opened here

PR #2403, opened 2024-05-13
feature/Add ChatOllama Function

HenryHengZJ:
image


AtmehEsraa:
What if I want to engage in conversation without needing to rely on a tool @HenryHengZJ

manuelbachl:
I always receive the response "Agent stopped due to max iterations" (Using llama3 as model)

dentro-innovation:
does ollama function support streaming?
Doesn't seem to me, but wanted to double check

erlebach:
The ChatOllamaFunction seems to have disappeared. It was there in commit 0de7fb85. I could not find it in the code base in `main.cpp`. It might have been deleted inadvertently. @HenryHengZJ

dentro-innovation:
Chat Ollama is very useful btw for using self hosted LLMs, hope we can get it back soon!

PR #2426, opened 2024-05-16
Updated bedrock model list

QuinnGT:
**Description:**

In the "models" array:
- Added a "description" field for each bedrock model, providing a brief overview of the model's capabilities.
- Updated the "name" field for some models by appending a version number (e.g., ":0", ":1", ":2") to the model name.
- Added new models:
    - "anthropic.claude-3-opus"
    - "mistral.mistral-large-2402-v1:0"
- Note: I didn't add the cohere command R models as langchain doesn't have support for their needed parameters (plus they rank low, so is it a loss?)

In the embeddings section:
- Added a new model "amazon.titan-embed-text-v2:0" with an embedding dimension of 1024.
- Added a "description" field for each embedding model, specifying the embedding dimensions.

Each model has been tested on my local with success.

### DIFF
--- a/packages/components/models.json
+++ b/packages/components/models.json
@@ -5,47 +5,68 @@
             "models": [
                 {
                     "label": "anthropic.claude-3-haiku",
-                    "name": "anthropic.claude-3-haiku-20240307-v1:0"
+                    "name": "anthropic.claude-3-haiku-20240307-v1:0",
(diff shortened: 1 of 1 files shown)


HenryHengZJ (approved):
thanks, this is helpful!

Issue #1400, opened 2023-12-15
[FEATURE] Support for TogetherAI


HenryHengZJ:
Add support for TogetherAI under LLM nodes - https://js.langchain.com/docs/integrations/llms/togetherai

- Must have access to API to test it out
- Must enable streaming mode - https://js.langchain.com/docs/integrations/llms/togetherai#streaming

jiabaow:
I can work on this.

HenryHengZJ:
Thanks! We have the integration in already, shouldve close this issue

jiabaow:
I see chatmodels/ChatTogetherAI and embeddings/TogetherAIEmbedding, but not llms/TogetherAI, did I miss something?

HenryHengZJ:
Oh yeah we don't have that yet, feel free to add the integration for LLM node

PR #2454, opened 2024-05-21
Together AI llms

jiabaow:
#1400 
This PR adds support for TogetherAI LLM.
Screenshot 2024-05-20 at 20 50 34

TODO:
- The streaming mode is currently unusable. When activated, it generates an error message indicating issues with JSON parsing: [ERROR]: [server]: Error: Unexpected token 'd', "data: {"id"... is not valid JSON. It appears to come from the server side.
- safetyModel and logprobs parameters are not included in the current implementation.


HenryHengZJ (approved):
looks good! thanks @jiabaow !

HenryHengZJ:
@jiabaow in order to add streaming, can you add `togetherAI` to this line [here](https://github.com/FlowiseAI/Flowise/blob/main/packages/server/src/utils/index.ts#L1092)

jiabaow:
@HenryHengZJ Thanks for your reply, however, the problem persists after adding `togetherAI` to src/utils: 
Screenshot 2024-05-23 at 14 52 07

HenryHengZJ:
I see, its okay we can remove it the line then. Lets have it not streaming for now. Since the ChatTogetherAI is able to stream

HenryHengZJ (approved):

PR #2458, opened 2024-05-21
Added Opensearch credential

patrickreinan:
Added Opensearch credential and updated Opensearch component.


packages/components/nodes/vectorstores/OpenSearch/OpenSearch.ts
```diff
@@ -3,8 +3,9 @@ import { Client } from '@opensearch-project/opensearch'
 import { Document } from '@langchain/core/documents'
```
HenryHengZJ, line 3: curious: why the need to move the URL into credential?

patrickreinan: Hi... when we are using OpenSearch with username and password, we need to pass credentials through URL. This situation is similar to MongoDb Atlas connection string. Because this I think OpenSearch URL as a connection string and should be a credential.

HenryHengZJ: got it make sense, thanks! Also can we increment the version number of the node

patrickreinan: Good. I will fix some errors I found.

HenryHengZJ (approved):
thanks @patrickreinan !

PR #2461, opened 2024-05-22
Fix JSON escaping

mokeyish:
This PR fixes the JSON parssing error that `variableValue` contains newlines

```json
{
    "text": "a\\nb" 
}
```

### DIFF
--- a/packages/server/src/utils/index.ts
+++ b/packages/server/src/utils/index.ts
@@ -780,7 +780,7 @@ export const getVariableValue = (
             const variableValue = variableDict[path]
             // Replace all occurrence
             if (typeof variableValue === 'object') {
-                returnVal = returnVal.split(path).join(JSON.stringify(variableValue).replace(/"/g, '\\"'))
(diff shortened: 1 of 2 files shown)


HenryHengZJ:
thanks! what is the expected output from your example?

image

mokeyish:
> thanks! what is the expected output from your example?

@HenryHengZJ  The second. This is because the paramValue is:

```json
{
    "functionInputVariables": {
        "config": "{{<variableNodeId>.data.instance}}"
    }
}
```
and the `<variableNodeId>.data.instance` is a json object: 
```json
{
    "text": "a\nb" 
}
```

---

Another solution： 

when `{{<variableNodeId>.data.instance}}` is a json object, remove the double quotes around it.

mokeyish:
The original method considers the escaping of double quotes, but does not consider the escaping of newlines.

HenryHengZJ:
got it:
image

HenryHengZJ (approved):

Issue #2466, opened 2024-05-22
[FEATURE] Support AWS IAM role-based authentication for storage S3 bucket and components


danieldabate:
**Describe the feature you'd like**
It would be great to have the alternative of not configuring access key and secret for AWS, and using IAM role authentication instead when possible. This is the preferred way of managing permissions in some companies because of the risk of access keys getting leaked.

By default, the AWS SDK will use IAM role auth if no credentials are passed, so just making AWS credentials optional for all services that support it should be enough to get this working. 

**Additional context**
A good example of a component that is doing this is the S3 File Loader component ([source](https://github.com/FlowiseAI/Flowise/blob/ff2381741e7a8a610c8ba0e31958bba32af846ca/packages/components/nodes/documentloaders/S3File/S3File.ts#L481)), which has credentials as optional for this reason. I tested it and it works.

Some places where credentials are currently mandatory, but could be optional:
- S3 storage bucket: [source](https://github.com/FlowiseAI/Flowise/blob/ff2381741e7a8a610c8ba0e31958bba32af846ca/packages/components/src/storageUtils.ts#L309)
- DynamoDB chat memory: [source](https://github.com/FlowiseAI/Flowise/blob/ff2381741e7a8a610c8ba0e31958bba32af846ca/packages/components/nodes/memory/DynamoDb/DynamoDb.ts#L94)

PR #2470, opened 2024-05-23
Feature: Support role-based authentication for AWS

danieldabate:
This pull request addresses issue #2466.

### Context
Currently, Flowise has four features or components that use AWS services:
- Flowise Storage
- DynamoDB Chat Memory
- AWS Bedrock
- S3 file

However, only AWS Bedrock and S3 file support AWS role-based authentication by making access keys optional. Since the AWS SDK supports seeking credentials from multiple sources, if access keys are not provided during client initialization, it will search for other authentication methods.

### What does this pull request do?
This PR makes credentials optional for Flowise Storage and DynamoDB Chat Memory, allowing for role-based authentication.

### Why is this useful?
Role-based authentication is particularly useful when deploying to K8S, as roles can be assumed by pods (Flowise) by configuring a ServiceAccount. It is also one of the preferred methods for granting access to applications, as AWS advises against generating access keys that are not regularly rotated. Additionally, many companies mandate the use of role-based authentication, prohibiting the creation of access keys.

### Was it tested?
Yes, it was tested in K8S with a ServiceAccount attached, and it works for both Flowise Storage and DynamoDB Chat Memory.

### DIFF
--- a/packages/components/nodes/memory/DynamoDb/DynamoDb.ts
+++ b/packages/components/nodes/memory/DynamoDb/DynamoDb.ts
@@ -46,7 +46,8 @@ class DynamoDb_Memory implements INode {
             label: 'Connect Credential',
             name: 'credential',
             type: 'credential',
-            credentialNames: ['dynamodbMemoryApi']
+            credentialNames: ['dynamodbMemoryApi'],
+            optional: true
         }
         this.inputs = [
             {
@@ -102,13 +103,18 @@ const initializeDynamoDB = async (nodeData: INodeData, options: ICommonObject): 
     const accessKeyId = getCredentialParam('accessKey', credentialData, nodeData)
     const secretAccessKey = getCredentialParam('secretAccessKey', credentialData, nodeData)
 
-    const config: DynamoDBClientConfig = {
-        region,
-        credentials: {
+    let credentials: DynamoDBClientConfig['credentials'] | undefined;
+    if (accessKeyId && secretAccessKey) {
+        credentials = {
             accessKeyId,
             secretAccessKey
-        }
+        };
     }
+    
+    const config: DynamoDBClientConfig = {
+        region,
+        credentials
+    };
 
     const client = new DynamoDBClient(config ?? {})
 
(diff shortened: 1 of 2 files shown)


packages/components/nodes/memory/DynamoDb/DynamoDb.ts
```diff
@@ -102,13 +103,18 @@ const initializeDynamoDB = async (nodeData: INodeData, options: ICommonObject):
...
+        credentials = {
             accessKeyId,
             secretAccessKey
-        }
+        };
     }
+    
+    const config: DynamoDBClientConfig = {
+        region,
+        credentials
```
HenryHengZJ, line 116: Curious if the credentials is undefined, what do we use? IAM role specified from your local env variables?

danieldabate: Good question! The AWS SDK checks a series of different sources for credentials, in a specific order. You can find more information in the [AWS docs](https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/setting-credentials-node.html#credchain).

Passing the credentials through code is the way that takes precedence, but by making it optional, we allow users to choose any other preferred authentication method. In my case, that's using an IAM OIDC provider inside the K8S cluster. But there are many other methods.

I used the S3 file component code as a reference for my changes.

danieldabate: Also, I'm going to open an issue to discuss how to add AWS authentication to the OpenSearch Vector Store component: [OpenSearch Docs](https://opensearch.org/docs/latest/clients/javascript/index/#authenticating-with-amazon-opensearch-service-aws-signature-version-4). Because I'm not sure if we should add it as an option to the existing component or create a new one called AWS OpenSearch.

HenryHengZJ (approved):
thanks @danieldabate !

HenryHengZJ:
some minor linting error, if you can do `pnpm lint-fix` @danieldabate

danieldabate:
> some minor linting error, if you can do `pnpm lint-fix` @danieldabate

Done! Thanks!!

PR #2473, opened 2024-05-24
Add LLM Chain in JSON single line

patrickreinan:
Add a single line json log with input and output of llm chain. It is very interesting to monitoring what is passing through LLM. It is single line because is a easy way to read these logs with a external log agent (fluentbit, fluentd, etc)

### DIFF
--- a/packages/components/nodes/chains/LLMChain/LLMChain.ts
+++ b/packages/components/nodes/chains/LLMChain/LLMChain.ts
@@ -141,9 +141,10 @@ class LLMChain_Chains implements INode {
         promptValues = injectOutputParser(this.outputParser, chain, promptValues)
         const res = await runPrediction(inputVariables, chain, input, promptValues, options, nodeData)
         // eslint-disable-next-line no-console
-        console.log('\x1b[93m\x1b[1m\n*****FINAL RESULT*****\n\x1b[0m\x1b[0m')
+        //console.log('\x1b[93m\x1b[1m\n*****FINAL RESULT*****\n\x1b[0m\x1b[0m')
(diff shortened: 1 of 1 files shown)


packages/components/nodes/chains/LLMChain/LLMChain.ts
```diff
@@ -141,9 +141,10 @@ class LLMChain_Chains implements INode {
         promptValues = injectOutputParser(this.outputParser, chain, promptValues)
         const res = await runPrediction(inputVariables, chain, input, promptValues, options, nodeData)
         // eslint-disable-next-line no-console
-        console.log('\x1b[93m\x1b[1m\n*****FINAL RESULT*****\n\x1b[0m\x1b[0m')
+        //console.log('\x1b[93m\x1b[1m\n*****FINAL RESULT*****\n\x1b[0m\x1b[0m')
         // eslint-disable-next-line no-console
-        console.log(res)
+        //console.log(res)
+        options.logger.info(JSON.stringify({ input, output: res }))
```
HenryHengZJ, line 147: you shouldn't need to log it here, its automatically log from `handler.ts`

patrickreinan: I got it. But handler.ts is writing logs in multiline format, because this is impossible to regroup all messages and insert in a data to future analysis. I could add this same log to handler.ts as logger.info and a single line json object. What do you think about it?

patrickreinan: Or... What do you think about change handler.ts loggings to be single line?

PR #2483, opened 2024-05-24
Added LOG_JSON_SPACES to control json beautify

patrickreinan:
Added: environment var LOG_JSON_SPACES with default value 2. Used to no beautify JSON on handler.ts
Fix: logger.verbose was not working because default log level was info
Updated: CONTRIBUITING.md

### DIFF
--- a/packages/components/src/handler.ts
+++ b/packages/components/src/handler.ts
@@ -14,16 +14,25 @@ import { ChainValues } from '@langchain/core/utils/types'
 import { AgentAction } from '@langchain/core/agents'
 import { LunaryHandler } from '@langchain/community/callbacks/handlers/lunary'
 
-import { getCredentialData, getCredentialParam } from './utils'
+import { getCredentialData, getCredentialParam, getEnvironmentVariable } from './utils'
 import { ICommonObject, INodeData } from './Interface'
 
 interface AgentRun extends Run {
     actions: AgentAction[]
 }
 
+function tryGetJsonSpaces() {
+    try {
+        return parseInt(getEnvironmentVariable('LOG_JSON_SPACES') ?? '2')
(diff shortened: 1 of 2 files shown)


packages/components/src/handler.ts
```diff
@@ -49,6 +58,7 @@ export class ConsoleCallbackHandler extends BaseTracer {
     constructor(logger: Logger) {
         super()
         this.logger = logger
+        logger.level = getEnvironmentVariable('LOG_LEVEL') ?? 'info'
```
HenryHengZJ, line 61: we might not need to specify this, as its already specified [here](https://github.com/FlowiseAI/Flowise/blob/bugfix/Remove-Doc-Store-File/packages/server/src/utils/config.ts#L12)

patrickreinan: When I tried to run with log level verbose, handler.ts used value info instead of verbose. Log level verbose just worked when I put this line.

patrickreinan: @HenryHengZJ I just check again and when I remove line 61, application stop to log verbose (or others levels), using default value info to log.

HenryHengZJ: if you try to set these env variables:
```
DEBUG=true
LOG_LEVEL=debug
```
you should be able to see the verbose logs no?

patrickreinan: I did it, but logs were shown multiline. DEBUG=true enables Langchain logs. It would be very good have these logs multiline to do a best ingestion in log analysis tools.

HenryHengZJ (approved):
looks good, thanks @patrickreinan !

patrickreinan:
😍

Issue #1220, opened 2023-11-13
[FEATURE] Enable test for Flowise


nobody4t:
**Describe the feature you'd like**

Currently, Flowise got no tests.
So enable the tests framework for this repos.

nobod4t:
@HenryHengZJ Is there any module/function which should be in 1st priority for testing. From my opinion, backend is.
Any advice?

HenryHengZJ:
backend definitely, functions in /utils

PR #2484, opened 2024-05-24
server-util-logger test

mingxin0607:
#1220 
Add test cases for packages/server/src/utils/logger.ts

Run with:

```
npx ts-node packages/server/test/utils/logger.test.ts
```

Output:
```
Running Test: Should call next() for a valid API call
next() called
```


HenryHengZJ:
thank you! some minor lint issues @mingxin0607

PR #2498, opened 2024-05-28
Fix JSON escaping (better)

mokeyish:
This PR fixes that #2461 cannot handle newline(`\r\n`) on windows.

### DIFF
--- a/packages/server/src/utils/index.ts
+++ b/packages/server/src/utils/index.ts
@@ -780,7 +780,7 @@ export const getVariableValue = (
             const variableValue = variableDict[path]
             // Replace all occurrence
             if (typeof variableValue === 'object') {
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
hmm by double stringify it? how does that work?

mokeyish:
It will do the escaping thing.

HenryHengZJ:
result: 
image

HenryHengZJ (approved):

mokeyish:
> result: 结果： image

It seems that we should strip the `"` from the result.

HenryHengZJ:
> @HenryHengZJ It seems that we should strip the `"` from the result.

yep, some users reported it doesnt work.

image
[Debug_Chatflow.json](https://github.com/user-attachments/files/15552309/Debug_Chatflow.json)

PR #2521, opened 2024-05-29
Bugfix/Upserting same id with supabase

HenryHengZJ:


opexchanger:
Hi, i'm not entirely sure if it has to do with this commit, but coincidentally, now when I upsert a record to my supabase vectors table, it always overwrite the old data. It's like the embeddings are coming with predefined ids starting from 0 and then replacing the ones that currently exist.

I'm sorry if this is not the correct place, if it's better i can open an issue about it. thanks!

PR #2526, opened 2024-05-30
Fix for Whisper Error: 'File is not defined' when using Speech to Text

raffareis:
Fix for error reported on Discord:

image

toFile() is a helper by openai to create the correct data to send to whisper.

Below is after the fix (disconsider the pt-br thing, I used the wrong ISO):

image


HenryHengZJ:
oh didnt notice they have this function, thats a big help!

PR #2530, opened 2024-05-30
feat(ci): enable `pnpm` caching in CI

SauravMaheshkar:
Set the `cache` argument to enable caching to speed up CI times. Similar workflow is adopted by [`nextauthjs/next-auth`](https://github.com/nextauthjs/next-auth/blob/d3eb3f9f07d1d19b11d5397bb9bdabf31dc72c0e/.github/workflows/release.yml#L241)

Request for Review: @HenryHengZJ


HenryHengZJ (approved):
thank you @SauravMaheshkar !

Issue #2533, opened 2024-05-30
[BUG] Custom JS Function not updating variable values after a second run


daniel-dabate-scopely:
**Describe the bug**
If variable values (static or dynamic) are updated or created, Custom JS Functions don't get the updated value after a second run.

**To Reproduce**
1. Create an static variable `test` with value `1`
2. Create a ChatFlow with a Custom JS Function `return $vars.test`
3. Connect it to a prompt that receives the output with the text: `Always reply with this number: {var}`
4. Connect it to an LLM Chain an LLM and save.
5. Send any text to the chat. Reply should be 1.
6. Go to variables, change the variable `test` from `1` to `2`.
7. Go to the ChatFlow and send any text to the chat. Reply is 1, but it should be 2.

**Expected behavior**
Values of variables should be always up to date, both for modification or creation of variables.

**Flow**
[Chatflow.json](https://github.com/user-attachments/files/15506513/Chatflow.json)

**Setup**

-   Installation [`docker`, `pnpm start`]
-   Flowise Version [`1.8.0`]
-   OS: [`macOS`, `Linux`]
-   Browser [`Chrome`]

**Additional context**
I debugged a little bit and found that this is related to the caching of ChatFlows. The code to check if the flow is reusable or not is not taking into account changes in variables. And because the Custom JS Function takes the variables at the time of initialisation, values are never updated in the cached instance.

Maybe more components are affected by this, like Custom Tools. I haven't tested it myself.

**Additional Question**
Besides fixing the bug, would it be possible to add an ENV variable to disable the cache entirely? Considering the usage of these tools is not latency sensitive, I think I might prefer to have always the ChatFlows recreated to reduce the risk of this type of bugs.

HenryHengZJ:
yep that was the reason, because it was reused without getting full execution again. We can add a FLAG to disable this, as more nodes are added, we can't possibly detect every edge cases and decide which one get to re-used

danieldabate:
Cool, I'll open a PR, I have it already implemented and tested. 

Only limitation I see is that in the latest version (1.8.0) the ChatflowPool is also being used to offer an endpoint to "abort" the flows. I think only the ones built with langgraph work, because they are referenced differently in the pool than normal flows. If ChatflowPool is disabled this endpoint won't do anything. 

What do you think about this Henry? Are there any plans to refactor this logic to avoid depending on RAM state?

HenryHengZJ:
[...]

PR #2559, opened 2024-06-03
Feature - Env variable to disable ChatFlow reuse

danieldabate:
Discussed in #2533, this PR adds a new env variable `DISABLE_CHATFLOW_REUSE` to allow users to decide if they want to create a new ChatFlow in each interaction or reuse current one when possible.

This is useful for a few reasons:
- Debugging purposes: If a user sees an issue and we think it might be related to caching, we can ask to disable it temporarily and try again.
- Avoid issues with unsupported flows: Flows like the one described in #2533 are not supported by current logic and users might decide to disable the reuse of ChatFlows.

### Implementation
Current implementation adds a new condition to the `ifFlowReusable` function to check for the new env variable `DISABLE_CHATFLOW_REUSE`.

This approach is simple and keeps the `chatflowPool` active for other use cases, like AgentFlow abortion.

### Alternative implementation
Another possible solution is to make the `chatflowPool` creation optional. At first this was my preferred approach because I think it's cleaner, but it would break the AgentFlow abortion logic. With the PR implementation regular ChatFlows are still stored in the pool but never used.


HenryHengZJ (approved):
thanks @danieldabate !

PR #2563, opened 2024-06-03
fix: Change the CMD flowise command in Dockerfile to be the same as in the document

erasmo-pinheiro:
The Dockerfile wasn't working with the command:
`CMD flowise`

Based on the documentation, to start Flowise, you should use the command "start". In this case, the Dockerfile was updated to use:

ENTRYPOINT ["npx", "flowise", "start"]

Reference:
[get started | Flowise](https://docs.flowiseai.com/getting-started)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -18,4 +18,4 @@ RUN npm install -g flowise
 
 WORKDIR /data
 
-ENTRYPOINT ["npx", "flowise", "start"]
+ENTRYPOINT ["flowise", "start"]
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -18,4 +18,4 @@ RUN npm install -g flowise
 
 WORKDIR /data
 
-CMD "flowise"
+ENTRYPOINT ["npx", "flowise", "start"]
--- a/docker/docker-compose.yml
+++ b/docker/docker-compose.yml
@@ -33,4 +33,4 @@ services:
             - '${PORT}:${PORT}'
         volumes:
             - ~/.flowise:/root/.flowise
-        command: /bin/sh -c "sleep 3; flowise start"
+        entrypoint: /bin/sh -c "sleep 3; flowise start"


docker/Dockerfile
```diff
@@ -18,4 +18,4 @@ RUN npm install -g flowise
 
 WORKDIR /data
 
-CMD "flowise"
+ENTRYPOINT ["npx", "flowise", "start"]
```
danieldabate, line 21: Thanks for this change, I'm using this image in K8S and I'm executing a different command because the one in the current image doesn't work.

But I don't think it's necessary to use `npx` given that Flowise is being installed globally in the previous step (`-g` flag). It should be more straightforward and faster to run `flowise start`. This is the command we are executing.

danieldabate:
@HenryHengZJ could you merge this one? It's an easy one and fixes the docker images which are currently broken.

HenryHengZJ:
Testing on my Windows machine running Docker on WSL, bump into errors:
image

danieldabate:
Ohh.. good catch! In Linux/Mac it's the opposite, it doesn't work without the `start` parameter, so I assumed the image was broken in every platform. 

I'll try to find out what's going, it's the first time I see inconsistent behaviors in different platforms for the same Docker image.

danieldabate:
This is the error that we get in Linux/Mac:
Screenshot 2024-06-04 at 10 48 35

HenryHengZJ:
I'm no Docker expert but suspect we might need `docker-entrypoint.sh`, I can help to test on Windows side, whereas you guys can test on Linux/Mac

danieldabate:
I implemented it with an entrypoint script as @HenryHengZJ suggested. This might solve the issue, WSL seems to be doing something different to run the images.

@HenryHengZJ could you try what I pushed to this branch in my own repository? This is working for me in Mac.
https://github.com/danieldabate/Flowise/tree/docker-image-fix

If it works @erasmo-pinheiro can apply the same changes to this PR or I can create a new PR.

danieldabate:
Ok, you can discard my previous message. I think I found what's going on.

@HenryHengZJ, you are running the Docker image with Docker Compose, right? The issue here is that the Docker Compose is trying to override the COMMAND, when it should actually override the ENTRYPOINT now.

I was testing running the Docker image directly with `docker run`, that's why it was working for me. If I do it from Docker Compose it fails.

Changing `command: /bin/sh -c "sleep 3; flowise start"` to `entrypoint: /bin/sh -c "sleep 3; flowise start"` in `docker-compose.yml` should solve the issue.

erasmo-pinheiro:
> Ok, you can discard my previous message. I think I found what's going on.
> 
> @HenryHengZJ, you are running the Docker image with Docker Compose, right? The issue here is that the Docker Compose is trying to override the COMMAND, when it should actually override the ENTRYPOINT now.
> 
> I was testing running the Docker image directly with `docker run`, that's why it was working for me. If I do it from Docker Compose it fails.
> 
> Changing `command: /bin/sh -c "sleep 3; flowise start"` to `entrypoint: /bin/sh -c "sleep 3; flowise start"` in `docker-compose.yml` should solve the issue.



> Ok, you can discard my previous message. I think I found what's going on.
> 
> @HenryHengZJ, you are running the Docker image with Docker Compose, right? The issue here is that the Docker Compose is trying to override the COMMAND, when it should actually override the ENTRYPOINT now.
> 
> I was testing running the Docker image directly with `docker run`, that's why it was working for me. If I do it from Docker Compose it fails.
> 
> Changing `command: /bin/sh -c "sleep 3; flowise start"` to `entrypoint: /bin/sh -c "sleep 3; flowise start"` in `docker-compose.yml` should solve the issue.

@danieldabate I've updated this Pull Request with the entrypoint in Docker Compose. It's ready to review.

HenryHengZJ:
awesomee it works! Thank you @danieldabate @erasmo-pinheiro !

PR #2565, opened 2024-06-03
Update Unstructured.ts - Add header

mohamed99akram:
Can you please make this header customizable, as it is hardcoded here. There are services that have more secured APIs, and these need different headers. This is helpful for dealing with sensitive documents.

### DIFF
--- a/packages/components/nodes/documentloaders/Unstructured/Unstructured.ts
+++ b/packages/components/nodes/documentloaders/Unstructured/Unstructured.ts
@@ -31,6 +31,8 @@ export class UnstructuredLoader extends BaseDocumentLoader {
 
     private apiUrl = 'https://api.unstructured.io/general/v0/general'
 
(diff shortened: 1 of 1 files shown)


packages/components/nodes/documentloaders/Unstructured/Unstructured.ts
```diff
@@ -67,6 +69,7 @@ export class UnstructuredLoader extends BaseDocumentLoader {
         const options = optionsOrLegacyFilePath
         this.apiKey = options.apiKey
         this.apiUrl = options.apiUrl ?? this.apiUrl
+        this.apiHeader = options.apiHeader ?? 'UNSTRUCTURED-API-KEY'
```
HenryHengZJ, line 72: where does `options.apiHeader` comes from? Don't you need that value from `UnstructuredFile.ts` and `UnstructuredFolder.ts`? And you if we were to make this user-configurable, we have to add a new input there as well

PR #2566, opened 2024-06-03
Update Unstructured.ts

MohamedYasserOaf:
I tried to use our API key with custom header but it always failed, I inspected the code and it turned out that the API header is hardcoded, I added a variable for the custom header to replace the hardcoded API header.


HenryHengZJ:
duplicate with https://github.com/FlowiseAI/Flowise/pull/2565

PR #2589, opened 2024-06-06
Bugfix - JS function "execute" from node crashing the frontend and response is hidden

danieldabate:
When testing a JS function from the node, a message with the results should appear below the code view.

Currently there are two issues:
- When JS execution throws an error, the frontend is crashing leading to a full white screen and losing the progress if not saved.
- Code view has a fixed size that doesn't give enough room for the error message to show. User has to scroll to view it. This makes it hard for users to know that there is actually a message with the response being shown and generates confusion even knowing that this happens. 

This PR tackles both issues:
- Fixes crash by transforming the response object to string to be printed on screen.
- Reduces heigh of code view by 180px to give room to response message. Screenshots below show the before and after.

**Before:**
Code view before execution
Before
Code view after execution
Before - Error

**After:**
Code view before execution
After
Code view after execution
After - Error

### DIFF
--- a/packages/ui/src/ui-component/dialog/ExpandTextDialog.jsx
+++ b/packages/ui/src/ui-component/dialog/ExpandTextDialog.jsx
@@ -152,7 +152,7 @@ const ExpandTextDialog = ({ show, dialogProps, onCancel, onConfirm }) => {
                     <div style={{ marginTop: '15px' }}>
                         <CodeEditor
                             disabled={true}
-                            value={codeExecutedResult}
+                            value={codeExecutedResult.toString()}
                             height='max-content'
                             theme={customization.isDarkMode ? 'dark' : 'light'}
                             lang={'js'}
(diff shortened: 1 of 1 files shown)


packages/ui/src/ui-component/dialog/ExpandTextDialog.jsx
```diff
@@ -100,15 +100,15 @@ const ExpandTextDialog = ({ show, dialogProps, onCancel, onConfirm }) => {
                                     borderColor: theme.palette.grey['500'],
                                     borderRadius: '12px',
                                     height: '100%',
-                                    maxHeight: languageType === 'js' ? 'calc(100vh - 250px)' : 'calc(100vh - 220px)',
+                                    maxHeight: languageType === 'js' ? 'calc(100vh - 430px)' : 'calc(100vh - 400px)',
```
HenryHengZJ, line 103: actually prefer to have `languageType === 'js' ? 'calc(100vh - 330px)' : 'calc(100vh - 220px)'`

so we have more spaces in the editable area:
image

IMO `430` is lil bit small for edtiable area:
image

danieldabate: Sounds good to me, as long as the response is still visible.

Ideally the code view should adapt in size to cover as much as possible depending on the rest of the content that is shown, but my frontend knowledge is not good enough to do it 😅

HenryHengZJ (approved):

packages/ui/src/ui-component/dialog/ExpandTextDialog.jsx
HenryHengZJ: yeah thats good enough for now! thanks!

PR #2592, opened 2024-06-06
Bugfix - Add x-forwarded-proto as source for http protocol for base url and add base url field to ChatflowTool

### LINKED ISSUE #2584: [BUG] ChatFlowTool using wrong URL depending on infrastructure setup
**Describe the bug**
Under special infrastructure setups, ChatFlowTool is trying to use a different or wrong URL which leads to a network failure and no answer by the tool.

An example of this is the following setup:
`Browser` --HTTPS--> `AWS ALB` --HTTP--> `Flowise`

The expected URL to execute would be:
https://someurl.com/path

But the executed URL is:
http://someurl.com/path

Which leads to a failure if HTTPS is enforced in the ALB. There might be more complex scenarios where the URL might also change.

**Expected behavior**
There is no expected behavior really, the code works fine. I just want to discuss this issue and propose a solution to make it more flexible.

I can think about different solutions:
* Adding an optional ENV variable "FLOWISE_BASE_URL" to make this base url configurable. Default would be current behavior.
* Adding an optional field to the ChatFlowTool called Base URL. Default would be current behavior.

Approach 1 covers the issue in a more global and central way, but it's not really clear to understand why changing "FLOWISE_BASE_URL" could affect the ChatFlowTool.

Approach 2 is more direct and clear to understand, but every ChatFlowTool added needs to be changed with the right URL.

I would love to get your thoughts on how to tackled this. Thanks!!

danieldabate:
Fixes #2584 

_Discussion is ongoing in the issue, but I feel this change is good enough while we evaluate if it's needed to change the whole approach._

### DIFF
--- a/packages/server/src/utils/buildChatflow.ts
+++ b/packages/server/src/utils/buildChatflow.ts
@@ -54,7 +54,9 @@ export const utilBuildChatflow = async (req: Request, socketIO?: Server, isInter
     try {
         const appServer = getRunningExpressApp()
         const chatflowid = req.params.id
-        const baseURL = `${req.protocol}://${req.get('host')}`
+
+        const httpProtocol = req.get('x-forwarded-proto') || req.protocol
+        const baseURL = `${httpProtocol}://${req.get('host')}`
 
         let incomingInput: IncomingInput = req.body
         let nodeToExecuteData: INodeData
--- a/packages/components/nodes/tools/ChatflowTool/ChatflowTool.ts
+++ b/packages/components/nodes/tools/ChatflowTool/ChatflowTool.ts
@@ -22,7 +22,7 @@ class ChatflowTool_Tools implements INode {
     constructor() {
         this.label = 'Chatflow Tool'
         this.name = 'ChatflowTool'
-        this.version = 1.0
+        this.version = 2.0
         this.type = 'ChatflowTool'
         this.icon = 'chatflowTool.svg'
         this.category = 'Tools'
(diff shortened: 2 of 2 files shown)


packages/server/src/utils/buildChatflow.ts
```diff
@@ -54,7 +54,9 @@ export const utilBuildChatflow = async (req: Request, socketIO?: Server, isInter
     try {
         const appServer = getRunningExpressApp()
         const chatflowid = req.params.id
-        const baseURL = `${req.protocol}://${req.get('host')}`
+
+        const httpProtocol = req.get('x-forwarded-proto') || req.protocol
```
HenryHengZJ, line 58: this doesn't fix it no? `req.get('x-forwarded-proto')` will always be `http` behind proxy, and we want `htttps`

HenryHengZJ: actually no, `req.get('x-forwarded-proto')` will be `https` behind proxy and `req.protocol` will be `http`. Have you tested it on AWS @danieldabate ?

HenryHengZJ: do we need to check for capital letters as well? `'X-Forwarded-Proto'` ?

danieldabate: "Get" method is case insensitive according to my tests, so current implementation works for all cases.

HenryHengZJ: awesome thanks!

HenryHengZJ (approved):

Issue #1401, opened 2023-12-15
[FEATURE] Add support for Fireworks


HenryHengZJ:
Add support for Fireworks
- LLM (https://js.langchain.com/docs/integrations/llms/fireworks)
- Chat (https://js.langchain.com/docs/integrations/chat/fireworks)

Must have access to API to test it out

jiabaow:
Hello, is anyone working on it already?

HenryHengZJ:
No not yet. Feel free to take a stab!

PR #2596, opened 2024-06-06
Fireworks ai chat model

mingxin0607:
#1401 

Add support for Fireworks AI chat model and manually test its functionality.


HenryHengZJ:
thank you! are you planning to work on the chat model as well ? https://js.langchain.com/v0.1/docs/integrations/chat/fireworks/

HenryHengZJ:
I think it supports streaming as well, can we try to add the name of this new component [here](https://github.com/FlowiseAI/Flowise/blob/main/packages/server/src/utils/index.ts#L1098)

mingxin0607:
> I think it supports streaming as well, can we try to add the name of this new component [here](https://github.com/FlowiseAI/Flowise/blob/main/packages/server/src/utils/index.ts#L1098)

Sure, added the name. Thanks for your reply!

packages/components/nodes/chatmodels/ChatFireworks/ChatFireworks.ts
```diff
@@ -0,0 +1,77 @@
...
+            credentialNames: ['fireworksApi']
+        }
+        this.inputs = [
+            {
+                label: 'Cache',
+                name: 'cache',
+                type: 'BaseCache',
+                optional: true
+            },
+            {
```
HenryHengZJ, line 41: We need to have the model field parameter, example - https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/chatmodels/ChatHuggingFace/ChatHuggingFace.ts#L41

Since there are so many models, we can just make it as a `type: 'string` field, and get rid of the `listModels` method

mingxin0607: I modified that, not sure if I understand it correctly. Please let me know if other modification is needed.

packages/components/nodes/chatmodels/ChatFireworks/ChatFireworks.ts
```diff
@@ -0,0 +1,82 @@
...
+                step: 0.1,
+                default: 0.9,
+                optional: true
+            }
+        ]
+    }
+
+    //@ts-ignore
+    loadMethods = {
+        async listModels(): Promise<INodeOptionsValue[]> {
```
HenryHengZJ, line 59: we can remove the whole `loadMethods`

packages/components/nodes/chatmodels/ChatFireworks/ChatFireworks.ts
```diff
@@ -0,0 +1,82 @@
...
+            return await getModels(MODEL_TYPE.CHAT, 'chatFireworks')
+        }
+    }
+
+    async init(nodeData: INodeData, _: string, options: ICommonObject): Promise<any> {
+        const cache = nodeData.inputs?.cache as BaseCache
+        const temperature = nodeData.inputs?.temperature as string
```
HenryHengZJ, line 66: we need to get the model from the input as well

mingxin0607: Do we need to use "model" or "modelName" as input when constructing ChatFireworks model using LangChain API? Because I didn't notice input parameter like "model" or "modelName" in LangChain's Fireworks chat model api, it only requires temperature and api key.

mingxin0607: https://js.langchain.com/v0.2/docs/integrations/chat/fireworks/

HenryHengZJ: we can use the `model`, updated!

HenryHengZJ (approved):
thank you @mingxin0607 !

PR #2597, opened 2024-06-06
add Fireworks LLM

jiabaow:
#1401 

This PR adds support for Fireworks LLM.
[update 2024-06-07] Support multiple models:
Screenshot 2024-06-07 at 16 32 32


packages/components/nodes/llms/Fireworks/Fireworks.ts
```diff
@@ -0,0 +1,74 @@
...
+            name: 'credential',
+            type: 'credential',
+            credentialNames: ['fireworksApi']
+        }
+        this.inputs = [
+            {
+                label: 'Cache',
+                name: 'cache',
+                type: 'BaseCache',
+                optional: true
```
HenryHengZJ, line 47: We need to have the model field parameter, example - https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/chatmodels/ChatHuggingFace/ChatHuggingFace.ts#L41

Since there are so many models, we can just make it as a type: 'string field, and get rid of the listModels method

jiabaow: done

HenryHengZJ (approved):
thanks @jiabaow !

PR #2598, opened 2024-06-06
Improvement - Reduce size of final docker image with multistage build

danieldabate:
This draft PR transforms the main Dockerfile to a multistage build. The multistage build allows for any unnecessary files, like npm cache files, to be left outside of the runtime image reducing the final size significantly.

I calculated the following sizes before and after the changes (numbers are approximates):

**Before**
Uncompressed: `~3.2GB`
Compressed: `~0.98GB`

**After**
Uncompressed: `~2.3GB`
Compressed: `~0.67GB`

But this could be optimised even more if we review the packages that are being installed with APK:
`git 
python3 
py3-pip 
make 
g++ 
build-base 
cairo-dev 
pango-dev`

These packages do not seem to be necessary when installing Flowise through npm and I doubt they are actually needed in runtime. I'm not a node expert, but I'd think that any dependencies needed are installed with npm. Chrome is definitely needed for Puppeteer, so we need to keep it.

If those APK packages are not installed final image sizes are:
Uncompressed: `~2.0GB`
Compressed: `~0.57GB`

Any thoughts and insights on why those packages are installed is welcome.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -3,6 +3,9 @@ FROM node:20-alpine as build
 
 USER root
 
+# Skip downloading Chrome for Puppeteer (saves build time)
+ENV PUPPETEER_SKIP_DOWNLOAD=true
+
 # Install latest Flowise globally (specific version can be set: flowise@1.0.0)
 RUN npm install -g flowise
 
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
oh wow didnt know such method exsits

those APKs are needed to all the different integration we have on Flowise, like FaiSS, Opensearch, etc. couldnt remember which APK is for which integration to work..

danieldabate:
Cool! We can leave those in the runtime stage then.

I'll make a small change to move the PUPPETEER_SKIP_DOWNLOAD to the build stage, as I incorrectly put it in the runtime stage (it saves some time), and set the PR as ready for review.

HenryHengZJ (approved):

PR #2599, opened 2024-06-07
Searxng Tool implementation

Getty:
A first searxng implementation for the tools


packages/components/nodes/tools/Searxng/Searxng.ts
```diff
@@ -0,0 +1,119 @@
...
+            format,
+            categories,
+            engines,
+            language,
+            pageno,
+            time_range,
+            safesearch
+        }
+
+        const headers = {}
```
HenryHengZJ, line 107: is this needed?

Getty: Not yet but i wanted to keep it there to show the ability that this is there, if someone needs it and want to extend the node.

seonghobae:
@Getty, Can you check #2815 ?

PR #2601, opened 2024-06-07
Add PromptOuput Node for corpus collection

mokeyish:


HenryHengZJ:
do you have a flow or example template?

mokeyish:
> do you have a flow or example template?

I hope to have an IfElse node that receives overideConfig to control whether to run this node and only outputs the dynamically generated Prompt. However, IfElse is not flexible enough and needs to be improved first, so it will be turned into a draft first.

If you have a better, more convenient solution, please let me know.

mokeyish:
@HenryHengZJ Here is the chatflow:
图片

I add `forwardTrue` and `forwardFalse`(type any) to IfElseFunction's input. IfElseFunction will disable the following branch that not match condition.



But I feel like this solution is not very good.


It should be something like 'if' in
https://docs.github.com/zh/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstepsif

HenryHengZJ:
will hold on this for now, we are releasing sequential agents which might be able to solve your needs

PR #2602, opened 2024-06-07
Fix apikey not URL safe

mokeyish:
This PR fixes apikey not URL safe(which make url routing not work). and add keyonly filtering the chatflows.


mokeyish:
I also use apikey to get different kinds of chatflows. if an apikey named RAG-QA, it can get all chatflows that the use this apikey.

HenryHengZJ (approved):

PR #2604, opened 2024-06-07
feat (server)： add support for setting listening host

mokeyish:


HenryHengZJ:
why the need to listen for host?

mokeyish:
> why the need to listen for host?

Because in development mode, the UI will proxy this address. I hope to configure it to listen to 127.0.0.1

mokeyish:
Could you merge this PR？ do you have any other concerns? 

If `HOST` not defined, its behavior is the same as before the change.

My local code has these changes, and it is troublesome to synchronize the code every time.

HenryHengZJ (approved):

HenryHengZJ:
@mokeyish sorry was held up by other things, just approved!

PR #2613, opened 2024-06-08
chore: move configurations to `package.json`

SauravMaheshkar:
Changes proposed by this PR can be summarised as follows :-

* Move the content from `.prettierrc` under the `prettier` key in `package.json`. [Source(prettier docs)](https://prettier.io/docs/en/configuration.html)
* Since `.prettierignore` also reads from `.gitignore` we can delete it.
* Move the content from `.eslintignore` under the `eslintIgnore` key in `package.json`. [Source(eslint docs)](https://eslint.org/docs/latest/use/configure/ignore#using-eslintignore-in-packagejson)
* Move the content from `babel.config.js` under the `babel` key in `package.json`. [Source(babel docs)](https://babeljs.io/docs/configuration#packagejson)

Doing so, would lead to a more minimal project structure without any functional change.

Request for Review: @HenryHengZJ


SauravMaheshkar:
Gentle ping @HenryHengZJ

PR #2628, opened 2024-06-12
Unstructured Upsert bug

mohamed99akram:
When upserting with the API, the uploaded files are of type pdfFile, txtFile, etc. but the code reads only fileObject which is the uploaded file using the button


HenryHengZJ (approved):
good catch! appreciated the fix!

PR #2633, opened 2024-06-13
Feature - Add option to start a new session with each interaction with the Chatflow tool

danieldabate:
This PR adds an option to the Chatflow Tool to decide if the conversation with the Chatflow keeps the same session or is independent with each interaction. This is very useful when using it with a Chatflow that has memory, to avoid it.

My use case is the following:
- Chatflow 1: Is a chat bot that answers questions based on RAG, with memory.
- Chatflow 2: Is an agent that uses the Chatflow 1 as a tool and it's instructed to, based on a list of questions and expected answers, ask them to the Chatflow 1 and evaluate each response.

In scenarios like this where repeated independent tasks are executed, memory is not desirable. Chatflow 1 might get confused about what is being asked in each independent task.

### DIFF
--- a/packages/components/nodes/tools/ChatflowTool/ChatflowTool.ts
+++ b/packages/components/nodes/tools/ChatflowTool/ChatflowTool.ts
@@ -258,7 +258,7 @@ class ChatflowTool extends StructuredTool {
 
         const body = {
             question: inputQuestion,
-            chatId: flowConfig?.chatId,
+            chatId: this.startNewSession ? uuidv4() : flowConfig?.chatId,
             overrideConfig: {
                 sessionId: this.startNewSession ? uuidv4() : flowConfig?.sessionId
             }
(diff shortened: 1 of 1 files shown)


HenryHengZJ (approved):
awesome thanks!

packages/components/nodes/tools/ChatflowTool/ChatflowTool.ts
```diff
@@ -242,7 +260,7 @@ class ChatflowTool extends StructuredTool {
             question: inputQuestion,
             chatId: flowConfig?.chatId,
```
HenryHengZJ, line 261: might need to replace `chatId` with a new uuid as well

danieldabate: That's actually a question I had. What's the difference between ChatID and SessionID? Should both be the same?

danieldabate: From what I tried, replacing just the SessionID is enough to make it work, but I also saw that ChatID is currently the same as SessionID when the tool is called.

HenryHengZJ: yeah for example, when using OpenAI Assistant:
sessionID = thread_abc (this is an ID from OpenAI)
chatId = uuid (this is an ID from Flowise)

in other words, chatId is sessionID specifically for flowise, whereeas sessionID is from 3rd party

danieldabate: Done, I pushed a change generating also a random id for the chatId

saatchi-david:
Hi @danieldabate, this sounds super useful for particular use cases. Are you able to share where this can be configured?

PR #2667, opened 2024-06-18
LunaryAI automatic Thread and User tracking

vincelwt:
Hi!

This PR extends the Lunary integration to automatically report user conversations using the Threads feature (not just the LLM calls), as well as report the user data if lead capture is enabled.

([Lunary](https://lunary.ai/) is an open-source LLM monitoring toolkit :) ). 

This was requested by many of our mutual users.

For example this is a Flowise conversation:
Screenshot 2024-06-18 at 18 58 39

Now visible into Lunary:
Screenshot 2024-06-18 at 18 59 05

Also with full support for tracing each assistant's message:
Screenshot 2024-06-18 at 18 59 18

---

I extensively tested this (both with and without Lunary enabled) but let me know if something else is required!

Cheers


BigFoxMedia:
Great contribution! Is this also supporting the open source variant of lunary? 

So, if we install lunary locally, can we use your integration in this PR to natively connect to the local variant or only the commercial one?

vincelwt:
Hi @BigFoxMedia yes absolutely, this is completely compatible with the open-source version. If you install Lunary locally, you just have to set the correct endpoint to `http://localhost:3333` in the settings :)

packages/server/src/utils/buildChatflow.ts
```diff
@@ -333,29 +343,80 @@ export const utilBuildChatflow = async (req: Request, socketIO?: Server, isInter
...
-                  chatId,
-                  chatflowid,
-                  logger,
-                  appDataSource: appServer.AppDataSource,
-                  databaseEntities,
-                  analytic: chatflow.analytic,
-                  uploads: incomingInput.uploads,
-                  prependMessages
-              })
+        const runParams = {
```
HenryHengZJ, line 346: i think the correct placement is [here](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/src/handler.ts#L357)

vincelwt: Hi @HenryHengZJ are there events for "on user message" and "on assistant response"? Couldn't find it in the callback handlers space, which seems to be more related with Langchain events (as this PR is for tracking the high level user conversation)

HenryHengZJ: We don't need `on user message` as we can already have `input` from the callbacks in [AnalyticHandler](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/src/handler.ts#L317)

For instance, open a thread when `onChainStart`, and close when `onChainEnd`. I'd like to keep 3rd party integration on `components` level and try to reuse the callbacks that are already exposed

vincelwt:
Hi @HenryHengZJ, hope all is well. 

We're looking at this PR now to try and adapt it. We'd also love to support the feedback tracking.

We're checking how we can fit the tracking of threads/messages into `onChainStart` (if we assume that a chain without `parentRunId` is always a chat message) like you suggested.

A few notes:

- for agent flows, a chain w/o parentRunId isn't necessarily a chat message, but there doesn't seem to be a metadata to know that.

- for user tracking, we need to find the lead's email and name, but there does't seem to be info about the chatId inside the data passed to onChainStart.  

Would there be a way to infer the id of the parent `chat_message` from a top level chain ID? 

- for feedback tracking, I'm not sure at how we could listen to that. Is there an event somewhere we could listen to? We could do regular polling but doesn't seem very clean.

Thanks for the help.

vincelwt:
Closing in favor of #3233

PR #2672, opened 2024-06-19
Bug fix - ChatBot Share - CURL with i/p config

HarshavardhanNetha:


HarshavardhanNetha:
image

The following parameter `modelName` value was not shown as a string.
Updated it to show as string with quotes.

Due to the above issue, it raises error when pasted in Postman like tool or when executed directly in CMD.
image

HenryHengZJ (approved):

PR #2674, opened 2024-06-19
Feature/Add filter to Postgres VectorStore

HenryHengZJ:
image

Verified working correctly:

1.) Set filter:
image

2.) Retrieve correctly
image


Astriel:
I've updated flowise and restarted docker, but seems I can't find the feature into the Postgres vector store ? Is there anything else to do ?

PR #2677, opened 2024-06-19
Add LangWatch integration

rogeriochaves:
Hello there 👋 

I'm the creator of [LangWatch](https://github.com/langwatch/langwatch), an LLM Ops and Analytics platform for teams to collaborate, we have some customers excited to use Flowise, so we worked on an integration between Flowise and LangWatch, I have followed the same steps as the LangSmith and LangFuse integrations, hope you can merge our contribution to the ecosystem

We've tested it with a few flows and it all worked 100%, but let us know if there is anything else we need to add, will also work on a PR for FlowiseDocs next


HenryHengZJ (approved):
thanks for the integration!

tested and looks good!

rogeriochaves:
Docs follow up: https://github.com/FlowiseAI/FlowiseDocs/pull/90

PR #2678, opened 2024-06-19
FIX #2617 Cherio Web Crawler doesn't work with large sites

### LINKED ISSUE #2617: [BUG] Big sites scan error: Error: vectorsService.upsertVector - TypeError: (intermediate value) is not iterable (cannot read property undefined)
**Describe the bug**
I scan the page using Cherio web scraper, if the page I want to scrape is large, I get the error as in the screenshot. smaller pages scrape without any problems. Same if i set  Relative Links Limit to 100 - works
Screenshot from 2024-06-10 13-13-42

**To Reproduce**
1. Create workflow with cheerio web scraper, like in screenshot
image

2. Start Upsert Vector Database
3. See error

**Expected behavior**
Database full of vectors :)

**Setup**

-   Installation pnpm start
-   Flowise Version 1.8.1
-   OS: Linux Ubuntu
-   Browser Chrome, Opera

ahmosman:
Hi,

There is a fix for issue https://github.com/FlowiseAI/Flowise/issues/2617. There were two problems. 

Firstly, sometimes cheerioLoader fails to download contnent of page and it returns undefined, so I added assignment of empty arrays if there is undefined. 

Secondly, CheerioWebBaseLoader doesn’t support loading PDF files. It takes a lot to load PDF file and the content is encoded so I believe it shouldn’t be downloaded as Document. So I added condition to avoid downloading PDF files.

I’ve tried it on the same Chatflow as in the issue.

### DIFF
--- a/packages/components/nodes/documentloaders/Cheerio/Cheerio.ts
+++ b/packages/components/nodes/documentloaders/Cheerio/Cheerio.ts
@@ -145,6 +145,7 @@ class Cheerio_DocumentLoaders implements INode {
                 return docs
             } catch (err) {
                 if (process.env.DEBUG === 'true') options.logger.error(`error in CheerioWebBaseLoader: ${err.message}, on page: ${url}`)
+                return []
             }
         }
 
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
do you have an example site that we can test before and after this PR?

ahmosman:
Sure, I tested on this site: https://www.cupraofficial.pl

packages/components/nodes/documentloaders/Cheerio/Cheerio.ts
```diff
@@ -161,17 +165,17 @@ class Cheerio_DocumentLoaders implements INode {
...
-                docs.push(...(await cheerioLoader(page)))
+                docs.push(...((await cheerioLoader(page)) || []))
             }
             if (process.env.DEBUG === 'true') options.logger.info(`Finish ${relativeLinksMethod}`)
         } else if (selectedLinks && selectedLinks.length > 0) {
             if (process.env.DEBUG === 'true')
                 options.logger.info(`pages: ${JSON.stringify(selectedLinks)}, length: ${selectedLinks.length}`)
             for (const page of selectedLinks.slice(0, limit)) {
-                docs.push(...(await cheerioLoader(page)))
+                docs.push(...((await cheerioLoader(page)) || []))
```
HenryHengZJ, line 175: instead of placing the `[]`, I think we should just modify the `cheerioLoader` function to return `[]` when `catch (err)`

ahmosman: I agree with you. I've added a fix as you suggested.

HenryHengZJ (approved):

PR #2696, opened 2024-06-21
Fix `observation?.includes is not a function`

mokeyish:
This error occurs when the result returned by CustomTool is of type number.


HenryHengZJ (approved):
thanks!

PR #2726, opened 2024-06-26
Fix llm instanceof wrong ChatOpenAI type

mokeyish:
update 
```ts
import { ChatOpenAI } from '../../chatmodels/ChatOpenAI/FlowiseChatOpenAI'
```
to 
```ts
import { ChatOpenAI } from '@langchain/openai'
```

### DIFF
--- a/packages/components/nodes/multiagents/Supervisor/Supervisor.ts
+++ b/packages/components/nodes/multiagents/Supervisor/Supervisor.ts
@@ -17,10 +17,10 @@ import { z } from 'zod'
 import { StructuredTool } from '@langchain/core/tools'
 import { AgentExecutor, JsonOutputToolsParser, ToolCallingAgentOutputParser } from '../../../src/agents'
 import { ChatMistralAI } from '@langchain/mistralai'
-import { ChatOpenAI } from '../../chatmodels/ChatOpenAI/FlowiseChatOpenAI'
 import { ChatAnthropic } from '../../chatmodels/ChatAnthropic/FlowiseChatAnthropic'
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
it is the correct import, `FlowiseChatOpenAI` is needed for `multiModalOption`

mokeyish:
> it is the correct import, `FlowiseChatOpenAI` is needed for `multiModalOption`

I see that the actual instance of llm is LanhchainChatOpenAI. Should logical OR be used here?

HenryHengZJ:
> > it is the correct import, `FlowiseChatOpenAI` is needed for `multiModalOption`
> 
> I see that the actual instance of llm is LanhchainChatOpenAI. Should logical OR be used here?

which line?

mokeyish:
> > > it is the correct import, `FlowiseChatOpenAI` is needed for `multiModalOption`
> > 
> > 
> > I see that the actual instance of llm is LanhchainChatOpenAI. Should logical OR be used here?
> 
> which line?

https://github.com/FlowiseAI/Flowise/blob/15a416a58f4865a07b35443a563d546e31e92d47/packages/components/nodes/multiagents/Supervisor/Supervisor.ts#L225

mokeyish:
@HenryHengZJ Is there any problem with this?

HenryHengZJ:
if you look at this [line](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/multiagents/Supervisor/Supervisor.ts#L419), `llm.multiModalOption` is only a method from the custom `FlowiseChatOpenAI` [here](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/chatmodels/ChatOpenAI/FlowiseChatOpenAI.ts#L32)

so for any LLMs that is able to process images, we need the custom override class, cant use the direct import

Issue #2733, opened 2024-06-27
[BUG] Broken Icon Display on Marketplace Tab


KonstantinOhrimenko:
Hello,

I've encountered an issue with the icon display on the Marketplace tab when the filter by type is set to 'Tool'.

Issue Details:
 - When the filter by type is active with the filter value 'Tool', a broken icon is displayed.
 - The icon is not adjusted to fit the container and is displayed as-is.
 - Other icons have the background property value 'center center / contain no-repeat', but this particular icon does not.

Steps to Reproduce:
1. Navigate to the Marketplace tab.
2. Set the filter by type to 'Tool'.
3. Observe the icon display.

Expected Behavior:
The icon should be adjusted to the container with the background property value center center / contain no-repeat.

Screenshots:
<img src="https://github.com/FlowiseAI/Flowise/assets/75303199/dc75005c-f583-42d9-b61d-c2e01a87ffcf" alt="Broken Icon 01" width=320>
<img src="https://github.com/FlowiseAI/Flowise/assets/75303199/72040d28-a665-4d03-b8e8-9aed276cc701" alt="Broken Icon 02" width=320>

Additional Information:
Browser: Google Chrome 125.0.6422.76
OS: Ubuntu 20.04.6 LTS
Flowise: flowise@1.8.2

PR #2737, opened 2024-06-27
Bugfix/broken icon display on market place tab

Mubashirshariq:
fixes: #2733 

changed the conflicted background name to backgroundImg

1)light mode
Screenshot from 2024-06-27 20-15-25

2)dark mode
Screenshot from 2024-06-27 20-15-18


packages/ui/src/ui-component/cards/ItemCard.jsx
```diff
@@ -72,7 +72,7 @@ const ItemCard = ({ data, images, onClick }) => {
                                         flexShrink: 0,
                                         marginRight: 10,
                                         borderRadius: '50%',
-                                        background: data.color
+                                        backgroundImage: data.color
```
HenryHengZJ, line 75: this should still be `background` no?

Mubashirshariq: I did this to make the code consistent,when their is no image then also instead of background i changed the name to backgroundimage

HenryHengZJ: but `backgroundImage` is only when we have an image url. For this case, its just a color, we should use `background`

HenryHengZJ (approved):
thank you!

PR #2743, opened 2024-06-28
Fix docker command

WilliamEspegren:
Noobs will struggle on this part a long time if they are using apt, since docker-compose actually is a package which is shown when you try to run `docker-compose up -d`

image


README.md
```diff
@@ -44,9 +44,9 @@ Download and Install [NodeJS](https://nodejs.org/en/download) >= 18.15.0
 
 1. Go to `docker` folder at the root of the project
 2. Copy `.env.example` file, paste it into the same location, and rename to `.env`
-3. `docker-compose up -d`
+3. `docker compose up -d`
```
HenryHengZJ, line 47: I think it depends on the OS, for example: when using ec2 instance with Linux, it is `docker-compose`

WilliamEspegren: Are you sure this isn't just an apt package that redirects to docker-compose?

WilliamEspegren: @HenryHengZJ docker's [official installation documentation](https://docs.docker.com/compose/install/linux/) uses `docker compose`

HenryHengZJ: if we were to update, there are other README that needed to be update, and Flowise Docs as well - https://github.com/FlowiseAI/FlowiseDocs

WilliamEspegren: I will look into it :)

WilliamEspegren: I made a PR in the FlowiseDocs as well: https://github.com/FlowiseAI/FlowiseDocs/pull/95

WilliamEspegren:
@HenryHengZJ Ready to merge along with https://github.com/FlowiseAI/FlowiseDocs/pull/95

HenryHengZJ (approved):

PR #2744, opened 2024-06-29
Bugfix/observation-includes-not-function

### LINKED ISSUE #2299: POST API chain issue, observation?.includes is not a function on large data
Hi, i am using post api chain, i am retrieving property information from an api, each property has around 100 fields that i am getting,sending payload and getting response, issue is when the api response contain large data like 10 properties returning i am facing error of "observation?.includes is not a function" when i ask questions in chatbot, but when it has 2 or 3 properties it predicts the answer right and doesn't throw error.
On large data response i am getting issue.

On small api response:
small

On large api response:
large

On both cases the format of response is same.

arunlodhi:
Fixes #2299


HenryHengZJ (approved):
thank you!

PR #2783, opened 2024-07-09
Feature: Option to add apikeys to the DB instead of api.json.

vinodkiran:


packages/server/src/services/apikey/index.ts
```diff
@@ -1,25 +1,87 @@
...
+import { APIKEYS_STORAGE_DB, APIKEYS_STORAGE_JSON } from '../../AppConfig'
+import { randomBytes } from 'crypto'
+import { Not, IsNull } from 'typeorm'
 
 const getAllApiKeys = async () => {
     try {
-        const keys = await getAPIKeys()
-        const dbResponse = await addChatflowsCount(keys)
-        return dbResponse
+        if (APIKEYS_STORAGE_JSON) {
```
HenryHengZJ, line 24: so if user that is using existing version, and upgrade to new version which includes this change, they wont have the storage type set right? from the code this goes into else block, and throw an error.

should we do something similar like Yau was doing. If no key is set initially, we migrate the existing apikey.json to database, then the database approach will be the default unless specified otherwise

vinodkiran: for the first time user, the key is set to APIKEYS_STORAGE_JSON. This is true for when the value is 'json' or when the key is missing. [AppConfig.ts - Line 1](https://github.com/FlowiseAI/Flowise/pull/2783/files#diff-cb7edb0c2cb215e1fc2456c1cc8ee8c7de92f29ba9d9e9f53d4efd1989f4c224R1)

packages/server/src/AppConfig.ts
```diff
@@ -0,0 +1,3 @@
+export const APIKEYS_STORAGE_JSON = !process.env.APIKEY_STORAGE_TYPE || process.env.APIKEY_STORAGE_TYPE === 'json'
```
vinodkiran, line 1: @HenryHengZJ - check this

packages/server/src/AppConfig.ts
```diff
@@ -0,0 +1,3 @@
...
+export const APIKEYS_STORAGE_DB =
+    process.env.APIKEY_STORAGE_TYPE === 'db' || process.env.APIKEY_STORAGE_TYPE === 'DB' || process.env.APIKEY_STORAGE_TYPE === 'Db'
```
HenryHengZJ, line 3: should we make this more geneic?

```
export const appConfig = {
    apiKeys: {
        storageType: process.env.APIKEY_STORAGE_TYPE ? process.env.APIKEY_STORAGE_TYPE.toLowerCase() : 'json',
    }
    // todo: add more config options here like database, log, storage, credential and allow modification from UI
}
```

HenryHengZJ: Also, this new env variable needed to add at other places like `start.ts`, `env.example`, etc

HenryHengZJ: > Also, this new env variable needed to add at other places like `start.ts`, `env.example`, etc

changes pushed

vinodkiran: Changes pushed!!!

packages/server/src/services/apikey/index.ts
```diff
@@ -28,32 +97,147 @@ const createApiKey = async (keyName: string) => {
...
+                            // should not reach here as we have already checked for existing keys
+                            throw new Error(`Key with name ${key.keyName} already exists`)
+                        }
+                        default: {
+                            throw new Error(`Unknown overwrite option ${body.importMode}`)
+                        }
+                    }
+                } else {
+                    const newKey = new ApiKey()
+                    newKey.id = key.id
```
vinodkiran, line 204: @HenryHengZJ I have an edge case, lets discuss!

CONTRIBUTING.md
```diff
@@ -121,7 +121,7 @@ Flowise has 3 different modules in a single mono repository.
 Flowise support different environment variables to configure your instance. You can specify the following variables in the `.env` file inside `packages/server` folder. Read [more](https://docs.flowiseai.com/environment-variables)
 
 | Variable                     | Description                                                                                     | Type                                             | Default                             |
-| ---------------------------- | ----------------------------------------------------------------------------------------------- | ------------------------------------------------ | ----------------------------------- |
+| ---------------------------- | ----------------------------------------------------------------------------------------------- | ------------------------------------------------ | ----------------------------------- | --- | --- |
```
HenryHengZJ, line 124: is this last few lines redundant?

vinodkiran: yes, and a few other lines needed fixing, pushed changes.

HenryHengZJ (approved):

PR #2796, opened 2024-07-12
Feature/extract import all

chungyau97:
Allow users to export and import all chatflows.
UI: 
image

### DIFF
--- a/packages/server/src/routes/chatflows/index.ts
+++ b/packages/server/src/routes/chatflows/index.ts
@@ -4,6 +4,7 @@ const router = express.Router()
 
 // CREATE
 router.post('/', chatflowsController.saveChatflow)
+router.post('/all', chatflowsController.saveChatflows)
 
 // READ
 router.get('/', chatflowsController.getAllChatflows)
--- a/packages/server/src/services/chatflows/index.ts
+++ b/packages/server/src/services/chatflows/index.ts
@@ -1,19 +1,18 @@
-import { StatusCodes } from 'http-status-codes'
-import { InternalFlowiseError } from '../../errors/internalFlowiseError'
-import { getRunningExpressApp } from '../../utils/getRunningExpressApp'
-import { ChatflowType, IChatFlow } from '../../Interface'
-import { ChatFlow } from '../../database/entities/ChatFlow'
-import { getAppVersion, getTelemetryFlowObj, isFlowValidForStream, constructGraphs, getEndingNodes } from '../../utils'
-import logger from '../../utils/logger'
 import { removeFolderFromStorage } from 'flowise-components'
-import { IReactFlowObject } from '../../Interface'
-import { utilGetUploadsConfig } from '../../utils/getUploadsConfig'
+import { StatusCodes } from 'http-status-codes'
+import { ChatflowType, IChatFlow, IReactFlowObject } from '../../Interface'
+import { ChatFlow } from '../../database/entities/ChatFlow'
 import { ChatMessage } from '../../database/entities/ChatMessage'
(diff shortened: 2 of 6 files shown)


packages/ui/src/layout/MainLayout/Header/ProfileSection/index.jsx
```diff
@@ -135,6 +242,29 @@ const ProfileSection = ({ username, handleLogout }) => {
...
+                                                <ListItemButton
+                                                    sx={{ borderRadius: `${customization.borderRadius}px` }}
+                                                    onClick={() => {
+                                                        getAllChatflowsApi.request()
+                                                    }}
+                                                >
+                                                    <ListItemIcon>
+                                                        <IconFileExport stroke={1.5} size='1.3rem' />
+                                                    </ListItemIcon>
+                                                    <ListItemText primary={<Typography variant='body2'>Export Chatflows</Typography>} />
```
HenryHengZJ, line 254: By hardcoding `Export Chatflows`, even if user is on Tools page, they will sstill see this. Should we have a conditon check and only display this import export when user is on `/chatflow`  or the main `/` page?

packages/server/src/services/chatflows/index.ts
```diff
@@ -198,6 +197,60 @@ const saveChatflow = async (newChatFlow: ChatFlow): Promise<any> => {
     }
 }
 
+const saveChatflows = async (newChatflows: Partial<ChatFlow>[]): Promise<any> => {
```
HenryHengZJ, line 200: should we rename to `importChatflows`? `saveChatflows` sounds like updating existing chatflows

packages/ui/src/api/chatflows.js
```diff
@@ -10,6 +10,8 @@ const getSpecificChatflowFromPublicEndpoint = (id) => client.get(`/public-chatfl
 
 const createNewChatflow = (body) => client.post(`/chatflows`, body)
 
+const createNewChatflows = (body) => client.post(`/chatflows/all`, body)
```
HenryHengZJ, line 13: will `importChatflows` a better name?

packages/server/src/routes/chatflows/index.ts
```diff
@@ -4,6 +4,7 @@ const router = express.Router()
 
 // CREATE
 router.post('/', chatflowsController.saveChatflow)
+router.post('/all', chatflowsController.saveChatflows)
```
HenryHengZJ, line 7: is `/import` better naming? not sure whats the standard best practices

packages/ui/src/layout/MainLayout/Header/ProfileSection/index.jsx
chungyau97: Nope, the reason is because this will soon be replace into "Import/Export All" for POC 2; Then it will be similar to "About Flowise", it can happen in any page.

HenryHengZJ (approved):

packages/ui/src/layout/MainLayout/Header/ProfileSection/index.jsx
```diff
@@ -55,6 +75,104 @@ const ProfileSection = ({ username, handleLogout }) => {
...
+
+            const exportFileDefaultName = 'AllChatflows.json'
+
+            const linkElement = document.createElement('a')
+            linkElement.setAttribute('href', dataUri)
+            linkElement.setAttribute('download', exportFileDefaultName)
+            linkElement.click()
+            exportChatflowsSuccess()
+        }
+    }, [getAllChatflowsApi.error, getAllChatflowsApi.data])
```
HenryHengZJ, line 174: `// eslint-disable-next-line react-hooks/exhaustive-deps`

PR #2807, opened 2024-07-16
Result object of endpoint function unused

28Smiles:


HenryHengZJ:
thanks for the PR, I opened one here that fix the issue where custom endpoint is not working - https://github.com/FlowiseAI/Flowise/pull/2811

PR #2844, opened 2024-07-22
Release/2.0.0

HenryHengZJ:
🥳 flowise@2.0.0 release

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
     "name": "flowise",
-    "version": "1.8.4",
+    "version": "2.0.0",
     "private": true,
     "homepage": "https://flowiseai.com",
     "workspaces": [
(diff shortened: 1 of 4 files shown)


ClanSpace:
Hi, sorry in advance if this is not the right place to put this, but I don't understand github much, I'm learning how to use it.I have a few questions. I'm looking to change the colors for the full-page implementation. Specifically, I'm having trouble changing the color of the top section of the chatbot in the full-page mode. It works fine for the pop-up mode. Is this a bug, or can it be modified from the base code? If it can, where should I look to make these changes?
Also, regarding the "agent message" that appears when the chatbot responds with an OpenAI agent after integration into a host, is it possible to modify this message? If so, could you point me to the relevant documentation or files? Lastly, where can I find the color sources and design files for the chatbots?

Issue #2847, opened 2024-07-23
[BUG] Agentflow Not Functioning Properly in Flowise 2.0.0


seonghobae:
**Description:** After upgrading to Flowise 2.0.0, we have encountered an issue where the existing Agentflow is not functioning correctly. Specifically, messages generated by the Worker are not appearing in the final output.

**Steps to Reproduce:**

1. Upgrade to Flowise 2.0.0.
2. Execute an existing Agentflow.
3. Observe that messages from the Worker do not appear in the final result.

**Expected Behavior:** Messages generated by the Worker should be included in the final output.

**Actual Behavior:** Messages from the Worker are missing in the final output.

**Environment:**

- Flowise Version: 2.0.0
- Operating System: Ubuntu
- Browser: Edge

**Additional Context:** This issue was not present in the previous version of Flowise. It appears to be a regression introduced in the 2.0.0 update.

**Screenshots:** If applicable, add screenshots to help explain your problem.
![image](https://github.com/user-attachments/assets/4e22ff5f-f8e8-4c28-abb8-00b76dedbe63)

HenryHengZJ:
This is the new behaviour. Since multi agents are working back and forth between workers, we think the final output should be a summarisation of the whole conversation.

For example. If you have a coder agent and a reviewer agent, and you ask the team to generate piece of code. The expected final output should be the code generated by coder agent. But it will be the reviewer's output.

This new behaviour is somewhat aligns with AutoGen studio 
![1706155331144](https://github.com/user-attachments/assets/355fb6f7-03cc-47f4-967e-5d9678ce65de)

seonghobae:
The problem with this approach is that when you send a POST request, it doesn't get the right text. I think this behaviour is quite misleading because the current Flowise native chat tool doesn't show all the Agent results either, and the Agent results are collapsed.

HenryHengZJ:
Understood, will see if we can create an optional field for supervisor, and provide summary only when user select so

HenryHengZJ:
@seonghobae changes has been pushed to 2.0.1, can you try and see?

seonghobae:
> @seonghobae changes has been pushed to 2.0.1, can you try and see?

I haven't reviewed the regression test in detail so I can't say for sure, but there doesn't seem to be anything out of the ordinary in our normal testing routine. If there is, I'll reopen the issue.

---

# Query
Find the weather in Seoul today specifically and tell me in English.

---

Response: 
[...]

PR #2848, opened 2024-07-23
[BUG] Agentflow Not Functioning Properly in Flowise 2.0.0 #2847

seonghobae:
🛠️ Supervisor.ts -> Removed summarization fields.

### DIFF
--- a/packages/components/nodes/multiagents/Supervisor/Supervisor.ts
+++ b/packages/components/nodes/multiagents/Supervisor/Supervisor.ts
@@ -30,7 +30,6 @@ Select strategically to minimize the number of steps taken.`
 
 const routerToolName = 'route'
 
-const defaultSummarization = 'Conversation finished'
 const defaultInstruction = 'Conversation finished'
 
 class Supervisor_MultiAgents implements INode {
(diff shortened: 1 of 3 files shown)


toi500:
While final summarization enhances contextual consistency between user queries and supervisor outputs, aiding task resumption with the introduction of the new Agent Memory node, it can also restrict the range of applications a Multi-Agent system can handle when a clean final output is essential.

So, can we make this optional?

seonghobae:
> While final summarization enhances contextual consistency between user queries and supervisor outputs, aiding task resumption with the introduction of the new Agent Memory node, it can also restrict the range of applications a Multi-Agent system can handle when a clean final output is essential.
> 
> So, can we make this optional?

The summarisation feature introduced in version 2.0.0 broke the behaviour of all the workers I created. I'm not a typescript expert, just a business data analyst, so this commit is to revert back to the previous state. In fact, it's not even clear from the git logs that I can revert this commit as a business data analyst, but I can confirm that removing the summary itself brought back the behaviour I intended. If the summary can be interpreted differently by different models and cause misbehaviour, I'd rather not have it. If you believe it should be an option, then I think it should be turned off by default.

HenryHengZJ:
Understood, will see if we can create an optional field for supervisor, and provide summary only when user select so

HenryHengZJ:
@seonghobae I've created a PR that brings back the old functionality, and move summarization as an optional field.

PR #2852, opened 2024-07-23
feat:  Supervisor allows customizing user prompt or empty to skip sending them

mokeyish:
I use Qwen2, which doesn't support user prompt, so need to skip sending user prompt.

-  allow customizing user prompt
-  ignore sending when user prompt is empty


HenryHengZJ:
Agentflow only have limited support on specific models, even if we discard the user prompt on supervisor, each result from worker is still a user message: 
image

PR #2884, opened 2024-07-25
Bugfix/Add streaming to chatfireworks

HenryHengZJ:
https://github.com/user-attachments/assets/7fbac41b-f562-459f-90eb-3148ac894ad5


davide445:
@HenryHengZJ not sure if better writing here or in the Closed issue so I'll do both: I'm experiencing the same bug using Flowise 2.0.2

Issue #2907, opened 2024-07-29
[BUG] AWSBedrockEmbedding throttling issue


RPKosec:
When using AWS Bedrock Embeddings on larger number of documents (larger Confluence space load for instance), AWS will return throttling error, which in turn will fail the upsert to vector store. 

**Steps to reproduce the behavior:**
1. Use Confluence document loader (should work with other too) and set it to data source with larger set of documents
2. Use AWS Bedrock Embeddings
3. Use In memory vector store (same behaviour was observed with Milvus and Postgres too)
4. Click Upsert
5. Error occurs

**Expected behavior**
When upserting larger number of documents using AWS Bedrock Embeddings, no error should be raised. Throttling error should be mitigated using batches (with batch size settable by user). If throttling happens due to to many requests, retry with timeout between retries (increasing with iterations, which should be limited and limit set by user).

**Flow Screenshots**
Basic example of flow, producing an error:
![AWSBedrockEmbeddingsBug](https://github.com/user-attachments/assets/59f1ece1-814c-4c62-afbb-1960c5bce8a8)

**Setup**

-   Installation: Local using VS Code dev containers, pnpm start
-   Flowise Version: 2.0.1
-   OS: macOS
-   Browser: Chrome

**Additional context**
Error trace of the exception:

```
2024-07-29 12:41:50 [INFO]: ⬆️ POST /api/v1/vector/internal-upsert/cf3dd83a-588e-4856-a53d-9865619492b6
@smithy/node-http-handler:WARN socket usage at capacity=50 and 5273 additional requests are enqueued. See https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-configuring-maxsockets.html or increase socketAcquisitionWarningTimeout=(millis) in the NodeHttpHandler config.
@smithy/node-http-handler:WARN socket usage at capacity=50 and 4261 additional requests are enqueued. See https://docs.aws.amazon.com/sdk-for-javascript/v3/developer-guide/node-configuring-maxsockets.html or increase socketAcquisitionWarningTimeout=(millis) in the NodeHttpHandler config.
2024-07-29 12:56:13 [ERROR]: ThrottlingException: Too many requests, please wait before trying again. You have sent too many requests.  Wait before trying again.
Error: ThrottlingException: Too many requests, please wait before trying again. You have sent too many requests.  Wait before trying again.
[...]

PR #2917, opened 2024-07-31
AWS Bedrock embeddings call retry extension

RPKosec:
AWS Bedrock embeddings retry for Titan models added.
Change sends requests to AWS Bedrock in batches, not all in one go and introduces timeout between calls if throttling exception is caught. After every retry, timeout is increased and stays at the value until the end of operation. Retry counter is reset when successful response is returned, but throws exception if there are to many throttling exceptions in a row.
Additional parameter, set by user are max retries and batch size. Current observations show, batch size 50 works well.


packages/components/nodes/embeddings/AWSBedrockEmbedding/AWSBedrockEmbedding.ts
```diff
@@ -144,7 +162,10 @@ class AWSBedrockEmbedding_Embeddings implements INode {
             if (iModel.startsWith('cohere')) {
                 return await embedTextCohere(documents, client, iModel, inputType)
             } else {
-                return Promise.all(documents.map((document) => embedTextTitan(document, client, iModel)))
+                // return Promise.all(documents.map((document) => embedTextTitan(document, client, iModel)))
```
HenryHengZJ, line 165: if we have proven the new approach works, we can go ahead and delete this

HenryHengZJ (approved):

packages/components/nodes/embeddings/AWSBedrockEmbedding/AWSBedrockEmbedding.ts
RPKosec: I did test on my setup using large Confluence space (2500+ pages before split) using flow below. Token Text Splitter is a must, otherwise you get to many tokens error, not related to this issue. It takes some time to process, but result is stored to vector database without errors, mentioned in opened issue #2907. I also noticed, I forgot to delete commented line, sorry for that.

AWSBedrockEmbeddingsFlow
AWSBedrockEmbeddingsFlow2

HenryHengZJ: yeah we can go ahead and remove the commented line

RPKosec: Done.

PR #2945, opened 2024-08-06
Chore/Move @langchain/core dependency to components

HenryHengZJ:


alwaysnolix:
I'm really enjoying learning more about AI capabilities with flowise. I'd like to contribute in some way if I can but I'm new to github. 

**I had a feature suggestion**: When you go to connect nodes to each other, any nodes that can be connected to the node you've clicked should display a green bubble or something like that to make it more clear what can be connected to what.

flowise node idea

How can I contribute to your documentation? I feel like as I fumble may way through flowise, I might be able to contribute some explanation for non-technical people to make flowise easier to use/understand. 

Cheers and keep up the great work!

PR #2956, opened 2024-08-07
OpenAI Model Updates

saatchi-david:
Hey @HenryHengZJ, trying my first PR  :D

Currently only "gpt4o" and "gpt4o-mini" models are available, which point to the latest version of the respective models.
I believe this is risky as updated model releases can introduce new and unexpected behavior.

I've updated models.json with the available gpt4o and gpt4o-mini versions, and updated labels with "(latest)" to help clarify which will always use the latest version.

This also gives everyone access to the new version of GPT4o just released today, which includes lower API costs. OpenAI hasn't yet updated the "gpt4o" model to point to this.

Hope that works / helps!


HenryHengZJ (approved):
awesome thanks @saatchi-david !

PR #2990, opened 2024-08-10
Bugfix/Tool Agent output message

HenryHengZJ:
When using Anthropic, the output has the following format:
```
[
  {
    index: 0,
    type: 'text',
    text: 'Hello! How can I assist you today?'
  }
]
```

This causes error when we try to regex.

Fix: check if its array, if yes take the text from the first item


takefy-dev:
The issue is still here https://github.com/FlowiseAI/Flowise/issues/3285

PR #3031, opened 2024-08-17
Feature/Add message history to agents

HenryHengZJ:
Adding few shot messages can greatly increase model performance, especially when it comes to function calling.

Reference: https://blog.langchain.dev/few-shot-prompting-to-improve-tool-calling-performance/


jazzandrock:
Hi guys! I was playing with your example called "Advanced Structured Output Parser" from the Marketplaces tab, and it didn't work. I think it's because of this line from this PR:
```js
if ((messageHistory && messageHistory === 'messageHistoryCode') || (selectedTab === 'messageHistoryCode' && messageHistoryCode)) {
```

Where `messageHistoryCode` can be `undefined`. 

With this line modified to check for `messageHistoryCode`, the example works. Maybe this is a bug. 

Anyway, thank you for the amazing product!

HenryHengZJ:
Hey @jazzandrock thanks for the heads-up! Would you be able to submit a PR fix? Happy to accept!

PR #3033, opened 2024-08-18
fix: ignore `allowedOrigins` config for undefined `origin` header to ensure correct CORS behavior

tanlethanh:
Hi Flowise team, 

I discovered that after configuring `allowedDomains` (as dynamic CORS) for a chatflow, It checks my origin header even if it is undefined. The request might be sent from a proxy server or postman and usually sends requests with undefined origin.

The error I got when requesting to `/api/v1/prediction` from postman or curl was:
```
{
    "statusCode": 500,
    "success": false,
    "message": "Invalid URL",
    "stack": {}
}
```


HenryHengZJ (approved):

HenryHengZJ:
thank you @tanlethanh !

PR #3042, opened 2024-08-20
Meilisearch - Added metadata as filterable attributes and polling on enqueued tasks

MohamedYasserOaf:
Hello there! back again with another improvement for Meilisearch component, I added metadata of the input documents to the metadata of meilisearch hits so no metadata information is lost, I also added all of the metadata content as filterable attributes to enable meilisearch filtering through the component. Moreover,some of meilisearch requests return an "enqueuedTask" , it creates an enqueuedTask depending on the request (adddocuments/update/delete/create-index etc...), so it was necessary to add proper polling on these created tasks to ensure correct upserting of large datasets and to avoid taking any action before these tasks are fully completed.

Example: loop on add documents enqueuedTask to ensure that the upsert function does not end before adding documents already happened on meilisearch cloud 

This is the shape of the updated component, it is fully tested and working on production environment, feel free to ask about any of the proposed changes. Thanks

meilisearch1-1


HenryHengZJ (approved):
thank you @MohamedYasserOaf !

MohamedYasserOaf:
you are welcome @HenryHengZJ ! thanks for your review 👍

PR #3059, opened 2024-08-23
Feature/add ability to upload file from chat

HenryHengZJ:
image

Embed PR: https://github.com/FlowiseAI/FlowiseChatEmbed/pull/230


Artemnikov:
That. Is a blasting feature
Thanks a lo <3

Can you add an example flow that handles the file read?

Thank you very much for this feature <333

Artemnikov:
Yeah, I added postgres vector store - tried both outputs.
(The documentation says postgres is supported)

It can only connect with auto-gpt which does not support file upload.

Thank you for your response!

PR #3068, opened 2024-08-23
added tls support to milvus

Eyalm321:


packages/components/nodes/vectorstores/Milvus/Milvus.ts
```diff
@@ -1,45 +1,45 @@
...
-        this.name = 'milvus'
-        this.version = 1.0
-        this.type = 'Milvus'
-        this.icon = 'milvus.svg'
-        this.category = 'Vector Stores'
-        this.description = `Upsert embedded data and perform similarity search upon query using Milvus, world's most advanced open-source vector database`
-        this.baseClasses = [this.type, 'VectorStoreRetriever', 'BaseRetriever']
+        this.label = 'Milvus';
+        this.name = 'milvus';
+        this.version = 1.0;
```
HenryHengZJ, line 30: can we increment the version to `2.0`?

HenryHengZJ:
`pnpm lint-fix` for linting

packages/components/nodes/vectorstores/Milvus/Milvus.ts
Eyalm321: You are using a newer version of Milvus SDK behind the scenes I don't see why not

HenryHengZJ: This is the version that is used to indicate the component version on Flowise, usually we update the version whenever there is new changes made to the component

Eyalm321:
> `pnpm lint-fix` for linting

done. updated to version with file upload aswell

packages/components/nodes/vectorstores/Milvus/Milvus.ts
```diff
@@ -136,10 +183,31 @@ class Milvus_VectorStores implements INode {
...
+            // partition
+            const partitionName = nodeData.inputs?.milvusPartition ?? '_default'
+
             // init MilvusLibArgs
             const milVusArgs: MilvusLibArgs = {
                 url: address,
-                collectionName: collectionName
+                collectionName: collectionName,
+                partitionName: partitionName,
+                clientConfig: {
```
HenryHengZJ, line 201: question: if user dont have secure SSL and the pem path, does it still works?

Eyalm321: it doesnt work for me but I made it secure in Milvus configurations, maybe with different configurations on Milvus config (i.e tls_mode: 1 instead of 2) it might work. Anyhow for my case its crucial.

PR #3102, opened 2024-08-28
Bugfix/prevent upsert when images are uploaded

HenryHengZJ:
Linked PR: https://github.com/FlowiseAI/FlowiseChatEmbed/pull/233


alper1969:
Same problem in my system.

PR #3075, opened 2024-08-25
Add $flow to Variable Resolution in Flow Building Process

raffareis:
## Changes Overview

This PR enhances the variable resolution process in the flow building mechanism, particularly focusing on the `buildFlow` and `getVariableValue` functions.

### Key Updates

1. **Flow Options Enhancement**
   - Added a new `options` object in the `buildFlow` function:
     ```typescript
     const options: ICommonObject = {
         chatflowid,
         chatId,
         sessionId,
         chatHistory
     }
     ```

2. **Variable Resolution Improvements**
   - Updated `resolveVariables` and `getVariableValue` functions to accept and utilize the new `options` parameter.

3. **Flow-specific Variable Support**
   - Implemented handling for variables starting with `$flow.` in the `getVariableValue` function:
     ```typescript
     if (variableFullPath.startsWith('$flow.') && options) {
         const variableValue = get(options, variableFullPath.replace('$flow.', ''))
         if (variableValue) {
             variableDict[`{{${variableFullPath}}}`] = variableValue
             returnVal = returnVal.split(`{{${variableFullPath}}}`).join(variableValue)
         }
     }
     ```

## Rationale

These changes allow for more flexible and context-aware variable resolution within the flow building process. By introducing the `options` object and supporting `$flow.` prefixed variables, we enable access to flow-specific data like `chatflowid`, `chatId`, `sessionId`, and `chatHistory` during variable resolution.

Also it can be used to enhance filtering and logic in any flow input:
image

## Testing

image
image
image

## Notes for Reviewers

- Pay special attention to the variable resolution logic in `getVariableValue`.
- Consider if there are any edge cases in flow-specific variable usage that should be addressed.

PR #3103, opened 2024-08-28
Passing state to tool so that we can use them in custom tools

Jrakru:
Allows the use of `$flow.state.var` inside the custom tool by passing the state in the ToolNode:

for example:
image

we have the output 
image

when we use the following function in the custom tool node:  ```return successfully called tool ${$flow.state.test}`;```

This is inlie with the behaviour mentionsed in the `How to Use Function`

image

Previously we were getting the following an error:
image

Ps: I am only passing the state without the messages since we are already using the toolNode to get the functions to calls. I dont see why we would need the messages in the tool itself but I could be shortsighted on this.


HenryHengZJ:
think we can close this one since this PR (https://github.com/FlowiseAI/Flowise/pull/3075) is merged and it fixed the issue

HenryHengZJ (approved):

HenryHengZJ:
thank you so much @Jrakru !

PR #3169, opened 2024-09-09
Feature: Custom Templates

vinodkiran:
### Ability to save flows as custom templates (for the market place)
image

### Assign Tags and Values
image

### New Tab in market place to show templates
image


HenryHengZJ:
can we also extend this to tool? so user can save as template as well:
image

HenryHengZJ (approved):

PR #3172, opened 2024-09-09
Bugfix: Check for relative path when saving file, to prevent unauthorised writes

vinodkiran:


HenryHengZJ:
can we use this library https://www.npmjs.com/package/sanitize-filename to sanitize it?

HenryHengZJ (approved):

PR #3179, opened 2024-09-12
Bugfix/chatOpenAICustom model not streaming

Wowso:
* Fixed issue preventing chatOpenAICustom model from streaming
* Updated streamAvailableLLMs to include chatOpenAICustom model


HenryHengZJ (approved):
awesome thanks!

PR #3216, opened 2024-09-18
Correct "as" casing

dillfrescott:
This was driving me crazy when building Flowise as it would give a warning that the casing was not consistent.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,5 +1,5 @@
 # Stage 1: Build stage
-FROM node:20-alpine as build
+FROM node:20-alpine AS build
 
 USER root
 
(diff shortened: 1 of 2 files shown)


dillfrescott:
Also removed "version" from docker compose because that is deprecated

HenryHengZJ:
> Also removed "version" from docker compose because that is deprecated. Any modern version of docker will also warn about version being deprecated.

if others are using the old docker engine, do they still need it?

HenryHengZJ:
cool, for now i will just change the casing first

dillfrescott:
Sounds good to me!

HenryHengZJ (approved):
thanks !

PR #3224, opened 2024-09-20
chore: update unstructured API url and doc reference

cragwolfe:
The non-deprecated serverless API URL for Unstructured is https://api.unstructuredapp.io/general/v0/general . Also update the doc reference.


HenryHengZJ (approved):
thanks!

PR #3233, opened 2024-09-22
LunaryAI automatic Thread and User tracking

vincelwt:
Hi :) 

This PR is a followup to #2667 making everything fit into handler.ts.

For reminder, this PR extends the Lunary integration to automatically report user conversations using the Threads feature (not just the LLM calls), as well as report the user data if lead capture is enabled.

This was requested by many of our mutual users. This is also fully supported by the OSS version of Lunary.

This conversation: 

image

Would show up like this in Lunary:

image

And can also be debugged with Traces:

image

-- 

We'd also love to support feedback tracking. However there seems to be some challenges of doing that from `handlers.ts`. We experimented using typeorm to listen to the database insertion events of ChatMessageFeedback, but this poses problem with reconciliating these to Lunary message IDs. Would love to discuss the best options here, such as adding a "onChatMessageFeedback" handler.

Thanks.


packages/components/src/handler.ts
```diff
@@ -487,7 +565,7 @@ export class AnalyticHandler {
                 await monitor.trackEvent('chain', 'start', {
                     runId,
                     name,
-                    userId: this.options.chatId,
+                    // userId: this.options.chatId,
```
HenryHengZJ, line 568: is it not possible to pass in `chatId` here for `trackEvent`? this `AnalyticHandler` is used when the module is not from langchain such as `OpenAIAssistant`

vincelwt: Thing is chatId is different than the userId, so this will have unintended consequences in Lunary (showing 1 new user for each conversation). 

Ideally we'd instead pull the current lead and pass the lead ID here.

HenryHengZJ: got it! in that case, I suppose we remove the commented line?

vincelwt: Done!

HenryHengZJ, line 568: last one here 😄

HenryHengZJ (approved):

PR #3261, opened 2024-09-26
feat: save a new Chatflow when the `ENTER` key is pressed

rhumbertgz:
This simple event handler improves the usability of the UI by avoiding having to use the mouse or having to tab twice and then hit enter to save a flow

Here you can see small example of the contribution of this PR

Before
saveflow-autofocus-without-keyboard-shortcut

After

saveflow-autofocus-with-keyboard-shortcut


HenryHengZJ (approved):
thanks @rhumbertgz !

HenryHengZJ (approved):
thank you @rhumbertgz !

PR #3263, opened 2024-09-26
Bugfix/Searxng tool not working

HenryHengZJ:
- fix searxng too error

TLDR: Make sure that you have `server.limiter` set to `false` and `json` is included in `search.formats`. These parameters can be defined in `searxng/settings.yml` :

```
server:
  limiter: false
general:
  debug: true
search:
  formats:
    - html
    - json
```
https://docs.flowiseai.com/integrations/langchain/tools/searxng


HenryHengZJ:
result: 
image

toi500:
It works pretty good now. Thanks for adding JSON as default format too. 

image

Just a minor issue with the link here:
image

PR #3264, opened 2024-09-26
feat: enable autofocus to the `edit chatflow title` field to improve UI usability

rhumbertgz:
The canvas header has only one input and it is the primary one, there is no need for an extra click to be able to edit the title

I think that in this case the eslint rule is properly disabled and it aligns with even [one](https://brucelawson.co.uk/2009/the-accessibility-of-html-5-autofocus/) of the refs listed in the jsx-a11y/no-autofocus

Here you can find two small gifs showing the contribution of this PR.

Before

edit-flow-name-autofocus-off

After

edit-flow-name-autofocus-on


HenryHengZJ (approved):
thank you @rhumbertgz !

PR #3265, opened 2024-09-26
feat: save Chatflow title when the `ENTER` key is pressed or discard upon `ESC` is pressed

rhumbertgz:
This simple event handler improves the usability of the UI by avoiding having to use the mouse to save or discard title changes

Here you can see small example of the contribution of this PR

Before

edit-flowname -without-keyboard-shortcut

After

edit-flowname -with-keyboard-shortcut


HenryHengZJ (approved):
thank you @rhumbertgz !

PR #3266, opened 2024-09-26
feat: highlight valid/invalid connection between nodes

rhumbertgz:
Change the inputs background to green/red to hint compatible connections, in addition to the `not-allowed` mouse cursor for incompatible connections.

Here you can find two small gifs showing the contribution of this PR.

Before
compatible-node-connections

After
highlight-compatible-node-connections


HenryHengZJ (approved):
thank you @rhumbertgz !

PR #3267, opened 2024-09-26
feat: add search keyboard shortcut based on the current platform

rhumbertgz:
This PR just add a well known keyboard shortcut to the search component of Chatflows and Agentflows

Here you can find two small gifs showing the contribution of this PR.

Before
search-flows-agents-views

After
keyboard-shortcut-search-flows-agents-views

PR #3270, opened 2024-09-26
fix: addEventListener error on views without search box introduced by commit #3267

rhumbertgz:
The error was caused  by [#3267](https://github.com/FlowiseAI/Flowise/pull/3267) when trying to add an event listener to an undefined search box in views without one like tools and assistants


HenryHengZJ:
thanks @rhumbertgz , I added the search to other views and also this changes as well in https://github.com/FlowiseAI/Flowise/pull/3271

HenryHengZJ:
Closing this

PR #3275, opened 2024-09-27
fix: warning when passing a boolean to border property of a Card

rhumbertgz:
By default MainCard wrappers like NodeCardWrapper and CardWrapper add a a solid border of 1px, but if the `MainCard.border` prop is used (`false`) the border prop was wrongly set to a boolean instead of string

Here is an example of the fixed warning

Screenshot 2024-09-27 at 09 58 47


HenryHengZJ (approved):
Thanks @rhumbertgz 🙏

PR #3276, opened 2024-09-27
fix: warning when a non-boolean values was used to set `checked` prop of a SwitchInput component

rhumbertgz:
The problem was that in the useEffect hook the plain value was used without validation like in useState

Screenshot 2024-09-27 at 10 46 44


HenryHengZJ (approved):
thanks @rhumbertgz !

PR #3280, opened 2024-09-27
Feature: Follow-up Prompts

0xi4o:
- Adds Follow-up Prompts tab to chatflow configuration dialog
Flowise-Low-code-LLM-apps-builder-09-30-2024_09_39_PM

- After receiving response from llm, generate follow-up prompts using the saved configuration and show them in the chat window
Flowise-Low-code-LLM-apps-builder-09-30-2024_08_50_PM

Related: https://github.com/FlowiseAI/FlowiseChatEmbed/pull/259


packages/ui/src/views/chatmessage/ChatMessage.jsx
```diff
@@ -664,6 +676,11 @@ export const ChatMessage = ({ open, chatflowid, isAgentCanvas, isDialog, preview
                 return allMessages
             })
         }
+
+        if (data.followUpPrompts) {
```
HenryHengZJ, line 680: When you refresh the page, follow up prompt is gone.

At Line 939, we need to also display the follow up prompt during initial useEffect. First check if chatflow has followup prompt status is true, and check if chatmessage has follow up prompt, if both fulfiled, display

0xi4o: Updated

HenryHengZJ (approved):

PR #3315, opened 2024-10-05
Added S3_FORCE_PATH_STYLE environment variable support for AWS S3 client configuration to ensure compatibility with custom S3 endpoints and path-style URLs. This change sets the forcePathStyle option based on the environment variable, defaulting to false if not provided.

thiagolealassis:


thiagolealassis:
Summary of Changes

This PR introduces support for the S3_FORCE_PATH_STYLE environment variable to configure the forcePathStyle option in the AWS S3 client. The forcePathStyle option ensures that the S3 client uses path-style URLs when interacting with S3 endpoints, which is crucial when working with custom S3-compatible services or self-hosted endpoints.

Reason for the Change

In certain scenarios, particularly when using custom S3-compatible storage solutions or services with non-standard domain structures, the default virtual-hosted style for S3 bucket URLs may not be compatible. Enabling the forcePathStyle option allows the use of path-style URLs, improving compatibility in these environments.

By adding this environment variable, users gain flexibility in configuring the client without modifying the codebase directly, defaulting to false if the variable is not set.

Importance of Acceptance

This change is backward-compatible and does not affect users who do not require path-style URLs. However, it provides a critical feature for those integrating with custom S3 endpoints, ensuring broader compatibility and adaptability of the Flowise project in diverse deployment scenarios.

HenryHengZJ (approved):
thanks @thiagolealassis !

PR #3332, opened 2024-10-10
Lunary: feedback tracking

vincelwt:
Hi! 

This PR enables mirroring of Flowise feedback into Lunary.

This is how feedback will show into Lunary:
image

This also removes a Lunary warning showing in the console in case no lead was captured.

Thanks!

### DIFF
--- a/packages/server/src/utils/buildChatflow.ts
+++ b/packages/server/src/utils/buildChatflow.ts
@@ -199,6 +199,8 @@ export const utilBuildChatflow = async (req: Request, isInternal: boolean = fals
         const nodes = parsedFlowData.nodes
         const edges = parsedFlowData.edges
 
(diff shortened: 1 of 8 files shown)


packages/server/src/utils/buildChatflow.ts
```diff
@@ -216,6 +218,7 @@ export const utilBuildChatflow = async (req: Request, isInternal: boolean = fals
                 chatId,
                 memoryType ?? '',
                 sessionId,
+                apiMessageId,
```
HenryHengZJ, line 221: the ordering of the params are incorrect for the function, should be:
```
chatId,
apiMessageId,
memoryType ?? '',
```

HenryHengZJ (approved):
thanks @vincelwt !

Issue #3340, opened 2024-10-11
[BUG] search nodes in Add Nodes doesn't work properly


mapachekurt:
**Describe the bug**
You want to add a node to your canvas so you type something like llm chain, but the search fails.

**To Reproduce**
Click on the add nodes button
Enter llm chain into the search box
Notice that when you type in the first part llm you see the llm chain
However when you enter theSearch term 'Llm chain' the search does not produce the result.

**Expected behavior**
Searching for llm chain should bring up that node only.

**Screenshots**
[Screen recording 2024-10-11 13.25.22.webm](https://github.com/user-attachments/assets/f8df2fca-e560-4be3-b487-85570d17e0b9)
Video Attached
screenshots too, as I'm not sure if that webm file is playable in github without using an external player.
![Screenshot 2024-10-11 13 28 59](https://github.com/user-attachments/assets/ea2f0768-0ba3-411d-a8b6-62455194a3be)
![Screenshot 2024-10-11 13 27 59](https://github.com/user-attachments/assets/cb48bfc4-e042-45ce-a62c-f6abf12f5aea)

PR #3348, opened 2024-10-13
fix: added search for node labels

prtkjakhar:
Fixes https://github.com/FlowiseAI/Flowise/issues/3340

Added support for search on node labels which was earlier not present.

Screenshot 2024-10-13 at 4 30 09 PM


HenryHengZJ (approved):
thanks @prtkjakhar !

mapachekurt:
This does not work in flowise cloud right now.
Screenshot 2024-10-30 11 58 49

PR #3355, opened 2024-10-14
[Feature] added Jina AI Embedding support

definitelynotchirag:
### [FEATURE] Jina AI Embedding support
- Added Jina AI embeddings Support using the jina-embeddings-v2-base-en
- Updated the models.json locally however changes need to be made in master file
- Added JinaApi Credential class
- Added JinaAI SVG Icon


HenryHengZJ (approved):
Thanks @definitelynotchirag !

Made a change on the model name to string type, because I think its better for users to specify the embedding model they want to use, instead of the selections provided in the `models.json` file

definitelynotchirag:
> Thanks @definitelynotchirag !
> 
> Made a change on the model name to string type, because I think its better for users to specify the embedding model they want to use, instead of the selections provided in the `models.json` file

Alright!! that would be great, Thank You for Approving!

PR #3359, opened 2024-10-15
Received tool input did not match expected schema resolution

DenisSmeon:
Resolved issue for Received tool input did not match expected schema

During the validation this.isXML was undefind so at tool.call was passed action.input directly without the structure but tool.call expected a structure as input.

with this new control the problem is resolved


DenisSmeon:
Opened issue instead this pr

Issue #3362, opened 2024-10-15
[BUG] Postgres VectorStore: TypeORMVectorStore issue with NULL characters and doesn't allow configuring content column name


JJK801:
**Describe the bug**
As Flowise Postgres VectorStore node uses TypeORMVectorStore, it doesn't deals with 0x00 characters from documents, leading to au postgres insert error.

It also enforce content column name to be `pageContent` (while PGVectorStore default is `text`, but allows to change it)

**To Reproduce**

1. Create a chatflow with Postgres as a vector store
2. try to load this document: https://sonoff.tech/wp-content/uploads/2021/03/%E8%AF%B4%E6%98%8E%E4%B9%A6-DUALR3-V1.0-20210305.pdf
3. PG doesn't insert due to error

**Expected behavior**
- Use PGVectorStore instead of TypeORMVectorStore (maybe make it configurable ?)
- Add additional configurations for customizing columns names

JJK801:
i can make a PR if needed

Issue #2198, opened 2024-04-16
[BUG] Postgres Vector Store inverted similarity calculation


vinibs:
**Describe the bug**
When using Postgres as the Vector Store for a flow and querying it with a specified minimum score, it's not returning anything. 
Searching the code, I noticed that the Postgres node calculates the distance between the vectors and returns them sorted ascending, but the VectorStore to Document node expects the number to be a similarity value, not the distance, resulting in it discarding the most relevant results for my query.
I'm making a local change to try changing this value returned by Postgres to be "1 - distance", which seems to be enough to fix this situation. 
If it doesn't bring other side effects, I can also make a PR for this little change.

**To Reproduce**

1. Create a flow
2. Configure a Postgres database server
3. Add the (Vector Store) Postgres node, with output set as "Postgres Vector Store" and set the connection up
4. Add the VectorStore to Document node and set its input as being the Postgres node's output
5. Pass "{{question}}" as the Query attribute of the VectorStore to Document node
6. Add a simple custom function as an ending node and set the VectorStore to Document node's output as an input variable for the function
7. Make the custom function only return the provided variable (for debugging purposes)
8. Use the Upsert API to insert data into the vector store for at least two different messages (questions)
9. Ask the chat one of the questions that were inserted into the Vector Store

**Expected behavior**
The output should bring the stored vectors ordered by the most similar to the question, but instead it brings them ordered by the less similar. If a minimum score is passed to the VectorStore to Document node, for the exact same question, no result is brought from the query. 

**Screenshots**
The results when querying without setting the minimum score (it brings all entries):
![Screenshot 2024-04-16 at 11 30 55](https://github.com/FlowiseAI/Flowise/assets/20691391/4254587b-c8d0-4954-b7a2-6ae37c18f978)

The results when querying with a minimum score of 80% (it doesn't bring any data):
![Screenshot 2024-04-16 at 11 31 28](https://github.com/FlowiseAI/Flowise/assets/20691391/0eca872b-7701-4d5f-9c23-8b5e50237987)

The VectorStore documents' log with the calculated similarity for this case (bringing the exact same question with a score of `0` while bringing non-related questions with greater scores - actually, greater distances):
![Screenshot 2024-04-16 at 11 33 30](https://github.com/FlowiseAI/Flowise/assets/20691391/f48ea675-618a-459a-959c-91dd22374aed)

**Flow**
[...]

PR #3367, opened 2024-10-16
[Prostgres Vector Store] Add PGVector Driver option + Fix null character issue w/ TypeORM Driver

JJK801:
Fix https://github.com/FlowiseAI/Flowise/issues/3362

Hi Flowise,

Here is a PR that improves PG Vector Store by:

- [x] Adding a `driver` config (which can be `typeorm` by default, or `pgvector`)
- [x] Adding an option to configure the `contentColumnName` (`pgvector` only), so that some can migrate easily from other tools
- [x] Fixing the 0x00 character issue (PG crash on insert) with `typeorm` driver

let me know if you need some more details or changes.


packages/components/nodes/vectorstores/Postgres/driver/TypeORM.ts
```diff
@@ -0,0 +1,154 @@
...
+            _filter = JSON.stringify(filter)
+        }
+
+        const queryString = `
+            SELECT *, embedding <=> $1 as "_distance"
+            FROM ${tableName}
+            WHERE metadata @> $2
+            ${notExists}
+            ORDER BY "_distance" ASC
+            LIMIT $3;`
```
JJK801, line 133: I'm dibutative about this logic... (i paste it as is from legacy code)

The WHERE clause sounds like: Where `metadata` includes `filter` OR `metadata` keys not includes `chatflowIdKey`
which seems to make the first part optional when second part is truthy, and this is clearly not the expected behavior

JJK801: In PGVector driver, i removed the chatflow part from filters and added a ` AND (<compareExistingValue> OR <keyNotExist>)`

HenryHengZJ: The original intention is to filter these conditions:
- metadata contains the JSON structure (filter)
OR
- metadata does not contains the key `flowise_chatId`

Say the following scenario:
1.) Owner of the chatflow upserted data to the postgres vector store through API, ex: `[{pageContent: 'abc', metadata: {}}]`
2.) User A interacts with the chatflow and upload some files. This will be upserted on the fly, resulting in `[{pageContent: 'this is userA', metadata: { flowise_chatId: userA-session }}]`
3.) User B interacts with the chatflow and upload some files. This will be upserted on the fly, resulting in `[{pageContent: 'this is userB', metadata: { flowise_chatId: userB-session }}]`

Now, when user A is asking questions, we want to only get 2 sources:
```
[
  {pageContent: 'abc', metadata: {}}
  {pageContent: 'this is userA', metadata: { flowise_chatId: userA-session }}
]
```
We do not want to get source with metadata of `flowise_chatId: userB-session`

Hence this results in the following filters:
metadata contains that specific chatId OR metadata that doesn't have the `flowise_chatId` key

JJK801: Tanks for your response @HenryHengZJ 🙏 

We're ok about the expected behavior 👍 

Now imagine a 4th scenario:
For whatever reason, i add a metadata filter on the vector store node, ex : `[{pageContent: 'this is userA', metadata: { flowise_chatId: userA-session, foo: 'bar' }}]`

With the actual or clause, the or clause will be: 

```sql
WHERE metadata @> { flowise_chatId: userA-session, foo: 'bar' }
OR NOT (metadata ? 'flowise_chatId')
```

This will return:
- Chatflow uploaded documents filtered with `foo=bar` 
- Any documents without `flowise_chatId`, not filtered with `foo=bar` 

If we want the right behavior, the where clause should be:

```sql
WHERE metadata @> { foo: 'bar' }
AND (metadata @> { flowise_chatId: userA-session } OR  NOT (metadata ? 'flowise_chatId'))
```

JJK801: as an alternative, we can also do:

```sql
WHERE metadata @> '{ foo: \'bar\' }'
OR metadata @> '{ flowise_chatId: \'userA-session\' }'
```

Which means:
- Any document with `foo=bar`
- Chatflow uploaded documents

HenryHengZJ: Long story short, you are absolutely right, we definitely want the 2nd query:
```sql
WHERE metadata @> '{ foo: \'bar\' }'
AND (metadata @> '{ flowise_chatId: \'userA-session\' }' OR NOT (metadata ? 'flowise_chatId'))
```

Would you be able to make the changes? for both typeorm and pgvector

packages/components/nodes/vectorstores/Postgres/driver/PGVector.ts
```diff
@@ -0,0 +1,107 @@
...
+                user: user,
+                password: password,
+                database: this.nodeData.inputs?.database as string
+            }
+        }
+
+        return this._postgresConnectionOptions
+    }
+
+    async getArgs(): Promise<PGVectorStoreArgs> {
```
HenryHengZJ, line 42: for `pgvector`, can we allow the option for choosing [distance strategy](https://github.com/langchain-ai/langchainjs/blob/main/libs/langchain-community/src/vectorstores/pgvector.ts#L40)? https://github.com/FlowiseAI/Flowise/issues/2198

packages/components/nodes/vectorstores/Postgres/driver/TypeORM.ts
JJK801: of course

packages/components/nodes/vectorstores/Postgres/driver/PGVector.ts
JJK801: probably, i don't know about it but it's a good learning opportunity ;)

packages/components/nodes/vectorstores/Postgres/driver/TypeORM.ts
JJK801: Done in https://github.com/FlowiseAI/Flowise/pull/3367/commits/2627b766ac32230f35066002af68b3ee873a08f8

packages/components/nodes/vectorstores/Postgres/driver/PGVector.ts
JJK801: @HenryHengZJ done in https://github.com/FlowiseAI/Flowise/pull/3367/commits/967f0bb53db4c514ed1d3b2756c431bf35257fd3

⚠️ not tested ⚠️

packages/components/nodes/vectorstores/Postgres/driver/TypeORM.ts
```diff
@@ -0,0 +1,173 @@
...
+
+        // Match chatflow uploaded file and keep filtering on other files:
+        // https://github.com/FlowiseAI/Flowise/pull/3367#discussion_r1804229295
+        if ($notexists) {
+            parameters.push({ [$notexists]: chatId })
+            notExists = `AND (metadata @> $${parameters.length} OR NOT (metadata ? '${$notexists}'))`
+        }
+
+        const queryString = `
+            SELECT *, embedding ${distanceOperator} $1 as "_distance"
```
HenryHengZJ, line 147: oh I didnt know we can even have different distance strategy in typeorm!

packages/components/nodes/vectorstores/Postgres/driver/PGVector.ts
HenryHengZJ: thanks! will test it out

packages/components/nodes/vectorstores/Postgres/driver/TypeORM.ts
JJK801: In fact TypeORM works on top of PGVector: https://github.com/langchain-ai/langchainjs/blob/4aee92415b7899045d806d83287ff7403229ff88/libs/langchain-community/src/vectorstores/typeorm.ts#L216

it just a some schema structure.

packages/components/nodes/vectorstores/Postgres/driver/TypeORM.ts
JJK801: @HenryHengZJ After more reflection, i think this query will match all the cases:

```sql
WHERE (metadata @> '{ foo: \'bar\' }' AND NOT (metadata ? 'flowise_chatId'))
OR metadata @> '{ flowise_chatId: \'userA-session\' }'
```

HenryHengZJ: Hmm I think its 3 clauses:

- Any documents with metadata contains `foo: bar`
- or any documents with metadata contains `foo: bar`, AND `flowise_chatId: <chatid>`
- or any documents with metadata contains `foo: bar`, AND without `flowise_chatId`

JJK801: IMO we don't care about `foo: bar` until there is a `flowise_chatId` set, because if it's been uploaded via the chatflow, it's always part of the chatflow context (and filters will not be added to the document's metadata).

Basically, the expectations are:
- Return any document uploaded via the chatflow
- Return filtered documents from the vector store
- Don't return the documents uploaded via other chatflows

JJK801: The part:
```sql
AND NOT (metadata ? 'flowise_chatId')
```
is just present to ensure that we don't return documents from other chatflows if there is no filter set:

```sql
WHERE (metadata @> '{}' AND NOT (metadata ? 'flowise_chatId'))
OR metadata @> '{ flowise_chatId: \'userA-session\' }'
```

HenryHengZJ: Okay so this will results if either of these two conditions is true:
- The metadata contains `foo': 'bar` and does not have `flowise_chatId`.
- The metadata contains `flowise_chatId': 'userA-session`, regardless of whether `foo': 'bar` is present or not.

Rewind back the scenario, and see if this fits:
1.) Owner upserted documents without metadata: `pageContent: 'this is doc 1', metadata: {}`
2.) Owner upserted documents WITH metadata: `pageContent: 'this is doc 2', metadata: { source: doc2 }`
Owner publish the chatflow to public
3.) User A uploaded a file from the chat window, results in `pageContent: 'this is user A', metadata: { flowise_chatId: userA }`
4.) User B uploaded a file from the chat window, results in `pageContent: 'this is user B', metadata: { flowise_chatId: userB }`

Now 4 docs are sitting in vector store:
1.) `pageContent: 'this is doc 1', metadata: {}`
2.) `pageContent: 'this is doc 2', metadata: { source: doc2 }`
3.) `pageContent: 'this is user A', metadata: { flowise_chatId: userA }`
4.) `pageContent: 'this is user B', metadata: { flowise_chatId: userB }`

Scenario 1: Owner doesnt set any filter on metadata. User A when chatting, should only get documents (1) (2) and (3)

Scenario 2: Owner pre set filter on metadata. User A when chatting, should only get documents (2) and (3)

Looks like the new clause you have there satisfy these 2 scenarios

JJK801: Fixed in https://github.com/FlowiseAI/Flowise/pull/3367/commits/0a39892ec56a34745285b3b66ec26004816a5ee5

packages/components/nodes/vectorstores/Postgres/driver/TypeORM.ts
JJK801: i just figured that we also have to handle this part for typeorm: https://github.com/langchain-ai/langchainjs/blob/7981aa950bd4042510704f8eb73221987aa8fd82/libs/langchain-community/src/vectorstores/pgvector.ts#L839

JJK801: Ok, it seems it's optional, so maybe we should handle this in another PR

https://js.langchain.com/docs/integrations/vectorstores/pgvector/#create-hnsw-index

packages/components/nodes/vectorstores/Postgres/driver/PGVector.ts
```diff
@@ -0,0 +1,119 @@
...
+            if (chatId) {
+                parameters.push({ [FLOWISE_CHATID]: chatId })
+
+                chatflowOr = `OR metadata @> $${parameters.length}`
+            }
+
+            if (queryString.match(whereClauseRegex)) {
+                queryString = queryString.replace(whereClauseRegex, `WHERE (($1) AND NOT (metadata ? '${FLOWISE_CHATID}')) ${chatflowOr}`)
+            } else {
+                const orderByClauseRegex = /ORDER BY (.*)/
```
HenryHengZJ, line 98: is this order by the distanceStrategy?

JJK801: it's just a matcher to tweak original query in cas there is no where clause (no filter), i match the order by clause (which is always present) and insert the WHERE clause before.

JJK801:
@HenryHengZJ Last commit (https://github.com/FlowiseAI/Flowise/pull/3367/commits/4e1f4c41c9a6756a9e473c77cdb5b1b8a2098da5) allows to set default Postgres credentials via env variables (for both vector store and record manager) and makes credentials optional if `user` and `password` are set.

HenryHengZJ:
> @HenryHengZJ Last commit ([4e1f4c4](https://github.com/FlowiseAI/Flowise/commit/4e1f4c41c9a6756a9e473c77cdb5b1b8a2098da5)) allows to set default Postgres credentials via env variables (for both vector store and record manager) and makes credentials optional if `user` and `password` are set.

thanks for that! I understand you prefer to read the value from an env file, but given that we have too many components, doing this will introduce too much env variables, and we wanted to limit env variables to only settings of the flowise apps, not for each and every components.

having said that, you can achieve the same using Variables: https://docs.flowiseai.com/using-flowise/variables#using-variables

For example, if I wanted to read Pinecone Index value, I can create a variable as such:
image
You can use a static value, or runtime value which can directly read from env file.

Then use it in the component:
image

JJK801:
> thanks for that! I understand you prefer to read the value from an env file, but given that we have too many components, doing this will introduce too much env variables, and we wanted to limit env variables to only settings of the flowise apps, not for each and every components.
This is a specific case because Langchain's tools for Postgres do not use env vars (that's why i documented it in Flowise) but almost all others do.

The reason why i'm doing this, is because of the `allow/deny all` strategy in flowise, i can't allow users to manage the chatflows without providing em' the ability to delete/modify credentials (which can break my whole production application in a mistake).

Also, i wan't to be able to deploy my stack all in once without having to manually setup credentials on each stage and put identifiers in client apps, or it will be an headache  to deploy. For this part, having the capacity to setup credentials as default  for each category would do the job, but it it doesn't solve the deletion issue.

packages/components/nodes/vectorstores/Postgres/Postgres.ts
```diff
@@ -58,28 +62,57 @@ class Postgres_VectorStores implements INode {
                 description: 'Keep track of the record to prevent duplication',
                 optional: true
             },
+            {
+                label: 'Driver',
+                name: 'driver',
+                type: 'options',
```
HenryHengZJ, line 68: `options`

packages/components/nodes/vectorstores/Postgres/driver/PGVector.ts
```diff
@@ -0,0 +1,118 @@
...
+    async fromDocuments(documents: Document[]) {
+        const instance = await this.instanciate()
+
+        await instance.addDocuments(documents)
+
+        return this.adaptInstance(instance)
+    }
+
+    protected async adaptInstance(instance: PGVectorStore, metadataFilters?: any): Promise<PGVectorStore> {
+        const { [FLOWISE_CHATID]: chatId, ...pgMetadataFilter } = metadataFilters
```
HenryHengZJ, line 65: this error when `metadatFilters` is undefined

```
const { [FLOWISE_CHATID]: chatId, ...pgMetadataFilter } = metadataFilters || {}
```

packages/components/nodes/vectorstores/Postgres/driver/PGVector.ts
```diff
@@ -0,0 +1,118 @@
...
+
+        const basePoolQueryFn = instance.pool.query.bind(instance.pool)
+
+        // @ts-ignore
+        instance.pool.query = async (queryString: string, parameters: any[]) => {
+            if (!instance.client) {
+                instance.client = await instance.pool.connect()
+            }
+
+            await instance.client.connect()
```
HenryHengZJ, line 81: I have this error: 
```
Error inserting: Client has already been connected. You cannot reuse a client.
```

HenryHengZJ:
> > thanks for that! I understand you prefer to read the value from an env file, but given that we have too many components, doing this will introduce too much env variables, and we wanted to limit env variables to only settings of the flowise apps, not for each and every components.
> > This is a specific case because Langchain's tools for Postgres do not use env vars (that's why i documented it in Flowise) but almost all others do.
> 
> The reason why i'm doing this, is because of the `allow/deny all` strategy in flowise, i can't allow users to manage the chatflows without providing em' the ability to delete/modify credentials (which can break my whole production application in a mistake).
> 
> Also, i wan't to be able to deploy my stack all in once without having to manually setup credentials on each stage and put identifiers in client apps, or it will be an headache to deploy. For this part, having the capacity to setup credentials as default for each category would do the job, but it it doesn't solve the deletion issue.

got it, in that case I'd suggest to add a README to the `packages/components/vectorstore/postgres` and same for record manager, for all the available env variables, and once these env variables are provided, it will override whatever you specified on the UI. I would remove the env variables from `CONTRIBUTING.md` and `.env.example` as those are meant for app settings. One effect of using env variables is that you will not have the flexbility to use different credentials if you have multiple flows

JJK801:
> got it, in that case I'd suggest to add a README to the `packages/components/vectorstore/postgres` and same for record manager, for all the available env variables, and once these env variables are provided, it will override whatever you specified on the UI. I would remove the env variables from `CONTRIBUTING.md` and `.env.example` as those are meant for app settings. One effect of using env variables is that you will not have the flexbility to use different credentials if you have multiple flows

LGTM 👍 

The expected behavior is that env vars act as default values and makes credentials optional, but it can be override by credentials if set.

packages/components/nodes/vectorstores/Postgres/driver/PGVector.ts
JJK801: hmm sorry, i implemented that because i got `Too many connections` error while using for a long period, but didn't test it since. I'll fix that.

packages/components/nodes/vectorstores/Postgres/driver/PGVector.ts
JJK801: i'll fix

packages/components/nodes/vectorstores/Postgres/Postgres.ts
JJK801: ?

packages/components/nodes/vectorstores/Postgres/driver/PGVector.ts
JJK801: Done in https://github.com/FlowiseAI/Flowise/pull/3367/commits/a7de41465d97704e2bd1aaeaabf88984b16e6e6c

I'm working on some issues of my project ATM, but i'll re-test all this stuff asap

packages/components/nodes/vectorstores/Postgres/driver/PGVector.ts
JJK801: Done in https://github.com/FlowiseAI/Flowise/pull/3367/commits/3527383899328dc339505d01771570b42848c987

packages/components/nodes/vectorstores/Postgres/Postgres.ts
HenryHengZJ: the type should be `options`, i think previously it was `option`, but good now, fixed

HenryHengZJ (approved):
sorry for late reply, thank you so much @JJK801 ! all looks good

JJK801:
my pleasure

PR #3368, opened 2024-10-16
Fix duplicate words in Airtable Agent description

robhitt:
* In multiple files the Airtable Agent description had the word "to" duplicated like this "Agent used to to answer queries on Airtable table". 
* Removed the duplicate


HenryHengZJ (approved):
thank you!

PR #3378, opened 2024-10-18
fix: unknown collate in AddCustomTemplate migration for MariaDB

karlobencic:
Fixes a problem when deploying with MariaDB database type.

`Migration "AddCustomTemplate1725629836652" failed, error: Unknown collation: 'utf8mb4_0900_ai_ci'`


HenryHengZJ (approved):
thanks @karlobencic !

PR #3409, opened 2024-10-24
fix: Add missing migration in MariaDB

karlobencic:
Add missing MariaDB migration for vector store configuration in document store


chungyau97 (approved):
The same script using in MySQL.

PR #3414, opened 2024-10-25
[Usntructured loaders] Make Unstructured API URL optional when environment variable is present

JJK801:
Hi Flowise,

When having an `UNSTRUCTURED_API_URL` set, we don't want this params as required in loaders.

This PR aims to:
- [x] make parameter required when env var is not set
- [x] make parameter optional when env var is set
- [x] allow parameter to override when env var is set

### DIFF
--- a/packages/components/nodes/documentloaders/S3File/S3File.ts
+++ b/packages/components/nodes/documentloaders/S3File/S3File.ts
@@ -70,7 +70,8 @@ class S3_DocumentLoaders implements INode {
                 description:
                     'Your Unstructured.io URL. Read <a target="_blank" href="https://unstructured-io.github.io/unstructured/introduction.html#getting-started">more</a> on how to get started',
                 type: 'string',
-                default: 'http://localhost:8000/general/v0/general'
+                placeholder: process.env.UNSTRUCTURED_API_URL || 'http://localhost:8000/general/v0/general',
+                optional: !!process.env.UNSTRUCTURED_API_URL
             },
             {
                 label: 'Unstructured API KEY',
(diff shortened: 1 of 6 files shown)


packages/components/nodes/documentloaders/S3File/S3File.ts
```diff
@@ -70,7 +70,8 @@ class S3_DocumentLoaders implements INode {
                 description:
                     'Your Unstructured.io URL. Read <a target="_blank" href="https://unstructured-io.github.io/unstructured/introduction.html#getting-started">more</a> on how to get started',
                 type: 'string',
-                default: 'http://localhost:8000/general/v0/general'
+                placeholder: process.env.UNSTRUCTURED_API_URL || 'http://localhost:8000/general/v0/general',
```
HenryHengZJ, line 73: normally we dont directly read the process env into component, same like we dont use process.env.OPENAI_APIKEY in ChatOpenAI node. I don't think URL is sensitive like Api Key right?

JJK801: no, URL is not as sensitive as the API Key, but id we want keep it secret we can just put `<From Environmen>` as a placeholder.

HenryHengZJ: @JJK801 can we also do the same thing as the postgres, where we add a README and get value

JJK801: of course

JJK801: Done in [ee2b1ac](https://github.com/FlowiseAI/Flowise/pull/3414/commits/ee2b1acf2d701c3efce301053a0ee7712eb12ed7)

HenryHengZJ (approved):

PR #3423, opened 2024-10-28
Support global-agent for central proxy configuration 

akoo24:
With central configuration of global agent via environment variables, the proxy agent is included in all node-fetch requests without any code changes and thus it is no longer necessary to implement the proxy URL in each ChatModel. Currently, the ProxyUrl is implemented in ChatOpenAI_ChatModels and as an example is still required in AzureChatOpenAI_ChatModels. The existing proxy mechanism would continue to work if required, because the global agent can be deactivated by a missing environment variable.

- Added global-agent as central Global HTTP/HTTPS proxy configurable sing environment variable
  -  https://www.npmjs.com/package/global-agent
- Server-Bootstrap on index.js 
  - import global-agent/bootstrap
- added dep in package.json 
  - global-agent: "^3.0.0

### DIFF
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -52,3 +52,9 @@ BLOB_STORAGE_PATH=/root/.flowise/storage
 
 # APIKEY_STORAGE_TYPE=json (json | db)
 # SHOW_COMMUNITY_NODES=true
+
+# Uncomment the following lines to enable global agent proxy
+# see https://www.npmjs.com/package/global-agent for more details
+# GLOBAL_AGENT_HTTP_PROXY=CorporateHttpProxyUrl
+# GLOBAL_AGENT_HTTPS_PROXY=CorporateHttpsProxyUrl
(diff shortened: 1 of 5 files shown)


HenryHengZJ:
thanks @akoo24 ! How do i test this though?

akoo24:
Hi Henry,

in our case we have an on-premises Kubernetes installation behind a corporate proxy.

This means that for testing purposes, outgoing connections to OpenAi / AzureOpenAI should only be possible via a proxy. In order to redirect outgoing requests centrally via a proxy, the following environment variables must be set at startup (see https://www.npmjs.com/package/global-agent)

- GLOBAL_AGENT_HTTP_PROXY: CorporateProxyUrl

- GLOBAL_AGENT_HTTPS_PROXY: CorporateProxyUrl

- GLOBAL_AGENT_NO_PROXY: Exception hosts to bypass proxy if needed


If the environment variables are not set, no centralized routing via proxy takes place,

Best Regards
Ali

HenryHengZJ (approved):
thank you @akoo24 !

PR #3440, opened 2024-11-01
Add Up and Down arrow functionality to chat messages

tgmerritt:
- Works like Linux shell
- History limited to 10 messages

Loom video showing functionality: https://www.loom.com/share/e7d557da0759454184f44ccf7fcd9c63?sid=7f563c6f-7bd4-45e5-9b80-fe269aabb867

### DIFF
--- a/packages/ui/src/views/chatflows/EmbedChat.jsx
+++ b/packages/ui/src/views/chatflows/EmbedChat.jsx
@@ -173,15 +173,6 @@ const chatwindowConfig = (isReact = false) => {
                         // sendSoundLocation: "send_message.mp3", // If this is not used, the default sound effect will be played if sendSoundMessage is true.
                         receiveMessageSound: true,
                         // receiveSoundLocation: "receive_message.mp3", // If this is not used, the default sound effect will be played if receiveSoundMessage is true.
-                        inputHistory: {
-                            enabled: true,
-                            maxHistory: 10, // Store last 10 messages
-                            persistHistory: true, // Optional: save to localStorage
-                        },
-                        keyBindings: {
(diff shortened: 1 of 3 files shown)


packages/ui/src/views/chatflows/EmbedChat.jsx
```diff
@@ -173,6 +173,15 @@ const chatwindowConfig = (isReact = false) => {
                         // sendSoundLocation: "send_message.mp3", // If this is not used, the default sound effect will be played if sendSoundMessage is true.
                         receiveMessageSound: true,
                         // receiveSoundLocation: "receive_message.mp3", // If this is not used, the default sound effect will be played if receiveSoundMessage is true.
+                        inputHistory: {
```
HenryHengZJ, line 176: do we need a follow up PR on Embed (https://github.com/FlowiseAI/FlowiseChatEmbed) for this functionality?

Currently it only works for Flowise UI right?

tgmerritt: I was actually trying to add this to the FlowiseChatEmbed _first_ but I couldn't get it clear in my head how to point the Flowise project at that Embed module locally on my machine.  I have much of that work in flight already, and if we can chat through it for 10 minutes and I can share my screen, much of the ChatHistory code is on a branch on my local and the same.

But yes, right now it's only the chat widget within the Flowise Agent flow that has this.

HenryHengZJ: since we are hardcoding it in embed, can we remove this config?
 
image

HenryHengZJ (approved):

PR #3452, opened 2024-11-04
Azure OpenAI Embeddings ignores env variable AZURE_OPENAI_BASE_PATH

akoo24:
* Set azureOpenAIBasePath to undefined if empty to enforce usage of env variable AZURE_OPENAI_BASE_PATH in @langchain+openai@0.0.30_encoding@0.1.13_langchain@0.2.11/node_modules/@langchain/openai/dist/embeddings.cjs

### DIFF
--- a/packages/components/nodes/embeddings/AzureOpenAIEmbedding/AzureOpenAIEmbedding.ts
+++ b/packages/components/nodes/embeddings/AzureOpenAIEmbedding/AzureOpenAIEmbedding.ts
@@ -71,7 +71,7 @@ class AzureOpenAIEmbedding_Embeddings implements INode {
             azureOpenAIApiInstanceName,
             azureOpenAIApiDeploymentName,
             azureOpenAIApiVersion,
-            azureOpenAIBasePath: basePath
+            azureOpenAIBasePath: basePath || undefined
         }
 
         if (batchSize) obj.batchSize = parseInt(batchSize, 10)


packages/components/nodes/embeddings/AzureOpenAIEmbedding/AzureOpenAIEmbedding.ts
```diff
@@ -71,7 +71,7 @@ class AzureOpenAIEmbedding_Embeddings implements INode {
             azureOpenAIApiInstanceName,
             azureOpenAIApiDeploymentName,
             azureOpenAIApiVersion,
-            azureOpenAIBasePath: basePath
+            azureOpenAIBasePath: basePath || undefined
```
HenryHengZJ, line 74: what happens if you dont have `basePath` specified ? it will automatically becomes undefined? or was it an empty string?

akoo24: exactly, `basePath` is an empty string by default, setting it to `undefined ` instead of empty ensures that env variable `AZURE_OPENAI_BASE_PATH` is used instead of empty in `@langchain+openai@0.0.30_encoding@0.1.13_langchain@0.2.11/node_modules/@langchain/openai/dist/embeddings.cjs`

HenryHengZJ (approved):

PR #3456, opened 2024-11-04
fix(proxy): restrict proxy to exact /api routes only

khakanali:
The proxy settings currently apply to all URLs that start with /api, including /apikey. This causes unintended routing where /apikey is proxied, resulting in a 404 error or incorrect loading behavior. The solution restricts proxy settings to exact /api routes or those that follow the /api/ pattern, allowing /apikey and similar routes to be handled by the frontend.


HenryHengZJ (approved):

HenryHengZJ:
thanks @khakanali !

PR #3480, opened 2024-11-08
feat: Add Custom Api Url Param for Self-hosted FireCrawl endpoint

ZacharyQin:
Minor change for supporting [self-hosted FireCrawl endpoint](https://docs.firecrawl.dev/contributing/self-host)
(By adding a missing param for `Api Url`)

Works fine for the  ver.0 API Scrape func.

### DIFF
--- a/packages/components/credentials/FireCrawlApi.credential.ts
+++ b/packages/components/credentials/FireCrawlApi.credential.ts
@@ -10,7 +10,7 @@ class FireCrawlApiCredential implements INodeCredential {
     constructor() {
         this.label = 'FireCrawl API'
         this.name = 'fireCrawlApi'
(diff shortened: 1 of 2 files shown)


packages/components/credentials/FireCrawlApi.credential.ts
```diff
@@ -18,6 +18,12 @@ class FireCrawlApiCredential implements INodeCredential {
                 label: 'FireCrawl API',
                 name: 'firecrawlApiToken',
                 type: 'password'
+            },
```
HenryHengZJ, line 21: we might want to increase the version above to 2.0

HenryHengZJ (approved):
thanks @ZacharyQin !

PR #3486, opened 2024-11-08
feat: Add BraveSearchAPI document loader

rynomster:
Adds a `BraveSearchAPI` document loader to use with Document Stores


HenryHengZJ (approved):
this is awesome, thanks @rynomster !

PR #3501, opened 2024-11-11
Feature/Add support for state-based metadata filter to Retriever Tool

serhiy-matoffo:
Extended Retriever Tool to support metadata filter based on the object stored in agent state.

### DIFF
--- a/packages/components/nodes/tools/RetrieverTool/RetrieverTool.ts
+++ b/packages/components/nodes/tools/RetrieverTool/RetrieverTool.ts
@@ -3,9 +3,22 @@ import { DynamicStructuredTool } from '@langchain/core/tools'
 import { CallbackManagerForToolRun } from '@langchain/core/callbacks/manager'
 import { DynamicTool } from '@langchain/core/tools'
 import { BaseRetriever } from '@langchain/core/retrievers'
+import { VectorStoreRetriever } from '@langchain/core/vectorstores'
 import { INode, INodeData, INodeParams } from '../../../src/Interface'
 import { getBaseClasses } from '../../../src/utils'
 import { SOURCE_DOCUMENTS_PREFIX } from '../../../src/agents'
+
+class FlowAwareDynamicStructuredTool<T extends z.ZodObject<any, any, any, any>> extends DynamicStructuredTool<T> {
+    private flowObj: any
+
+    setFlowObject(flow: any) {
+        this.flowObj = flow
+    }
+
+    getFlowObject() {
+        return this.flowObj
+    }
+}
 
 class Retriever_Tools implements INode {
     label: string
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
Interesting! do you have an example use case?

serhiy-matoffo:
Sure. We've found that our RAG flow often returns irrelevant documents. To be a bit more precise, result documents contain some content related to some words used by the user but are found on documents on a completely different subject. So, we've tasked one of our LLM Nodes to figure out the subject of the conversation and save that to state (at this point we can form a metadata filter object). Later down the line, we would like to take that subject and use it to narrow down the query to the vector store.

Is there a way to implement this flow using existing functionality? I couldn't find one and had to add this property to a Retriever Tool.

HenryHengZJ:
I was thinking of how to make it more generic..

How bout this? 
image

We can have an additional metadata filter, this adds to the existing filter from vector store if there is any. And users can filter using the existing flow variable such as `$flow.state`, `$flow.chatId` etc

serhiy-matoffo:
This would not allow complex/nested filters, right?

HenryHengZJ:
> This would not allow complex/nested filters, right?

Do you have an example of that? Under the hood it's just a simple JSON that adds on top of existing filter

serhiy-matoffo:
>Do you have an example of that?

Here is an example from pinecone documentation: `{"$and": [{"genre": "comedy"}, {"genre":"documentary"}]}`

Would the functionality you suggested support this kind of syntax?

HenryHengZJ:
yep definitely. If thats okay I can make the changes, you can then try it out

serhiy-matoffo:
Sounds good!

HenryHengZJ:
Updated, try and test see if it works @serhiy-matoffo

serhiy-matoffo:
**I can confirm that it works great, thanks!** 

Also, this trick allows using any dynamic object from the flow state:

image

I just prefer to form that object using a script instead of typing in all the properties through those boxes 😄

HenryHengZJ (approved):
merging

sirsimonson:
so excited to try this out

sirsimonson:
@HenryHengZJ Also the above filter mechanism (which had worked through "Test Retrieval Query" doesnt work for this new feature. Somehow the filter is not passed down to the postgres query

Sarahkunk:
Dynamic filtering is just what I was looking for. Has this been updated in cloud? Mine's showing older version still which doesn't have the filtering field: 

Screenshot 2024-11-26 at 11 59 38 AM


Screenshot 2024-11-26 at 12 00 59 PM

Trying it with pinecone but it doesn't pick up the dynamic values : 

Screenshot 2024-11-26 at 12 11 58 PM

Thankss

Abdulwahab-17:
Can I get an example chatflow, Which i can modify to test this?

PR #3543, opened 2024-11-21
Added new gpt4o model and set "gpt 4o (latest)" as default model for ChatOpenAI node.

saatchi-david:
OpenAI just released a new version of gpt4o: gpt-4o-2024-11-20
Added this to models.json.

Also updated the ChatOpenAI node to use "gpt4o (latest)" by default instead of the old and expensive 3.5 turbo.


HenryHengZJ:
Updated the default to gpt 4o mini

HenryHengZJ (approved):

PR #3578, opened 2024-11-26
fix: change data source lifecycle on agent memory mysql saver

joaopaulo-m:
Fix new Agent Memory database option MySQL creating multiple connections to database throughout chat messages


packages/components/nodes/memory/AgentMemory/mysqlSaver.ts
```diff
@@ -54,80 +50,64 @@ CREATE TABLE IF NOT EXISTS ${this.tableName} (
     }
 
     async getTuple(config: RunnableConfig): Promise<CheckpointTuple | undefined> {
-        await this.setup()
+        const dataSource = await this.getDataSource()
```
HenryHengZJ, line 53: thanks for the PR!

Wondering why are we initializing a new data source for every action? this will most likely cause memory leak

HenryHengZJ: ignore me, actually with current implementation every run will causes constructor to be initialized, hence its the same

HenryHengZJ (approved):
thank you @joaopaulo-m !

PR #3579, opened 2024-11-26
Feature/Extract Metadata Retriever

HenryHengZJ:
image

This retriever is designed to automatically extract keywords from query. The extracted JSON output is used as metadata filter for vector store.

For example, when we ask a question: "What is the profit for Apple", LLM will give an output of `{source: "apple"}`, and this will be passed to vectore store's metadata filter


toi500:
Outstanding.

saatchi-david:
Is my understanding correct that this is basically a more dynamic way to set a metadata filter rather than specifying one on the vector store node? Will it be added in addition to any specified on the vector store node?

jslicao:
But it doesn't work on Milvus.
the result is the same as ones which I don't give the metadata information.

PR #3580, opened 2024-11-26
Feature/add ability to show variable output

HenryHengZJ:
image

image


Hbmr9000:
maybe a newbie question, but could you explain the environment you used to have the input/output to the chatflow like this?

PR #3591, opened 2024-11-27
Bugfix/redis connection is closed

xmaiconx:
Refactored Redis Connection Management

This pull request refactors the Redis connection handling to remove the singleton pattern, which previously restricted the application to a single Redis client configuration. By opening and closing connections per operation, we ensure that multiple Redis clients can be configured and used independently within the application. This change improves flexibility and aligns with best practices for managing database connections.


HenryHengZJ (approved):
thank you @xmaiconx !

PR #3651, opened 2024-12-07
feat(models): add Llama 3.3 70B models to groqChat

jeanibarz:
Add Meta's latest Llama 3.3 70B models to groqChat options:
- llama-3.3-70b-versatile: New versatile model for general tasks
- llama-3.3-70b-specdec: Specialized model for specific tasks

These models offer comparable quality to Llama 3.1 405B at 1/5th the size, with improvements in:
- Reasoning and math capabilities
- General knowledge tasks
- Instruction following
- Tool use and JSON outputs
- Code generation and feedback

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -1,6 +1,6 @@
 {
     "name": "flowise-components",
-    "version": "2.2.1",
+    "version": "2.2.2",
     "description": "Flowiseai Components",
     "main": "dist/src/index",
     "types": "dist/src/index.d.ts",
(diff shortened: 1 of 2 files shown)


packages/components/package.json
```diff
@@ -1,6 +1,6 @@
 {
     "name": "flowise-components",
-    "version": "2.2.1",
+    "version": "2.2.2",
```
HenryHengZJ, line 3: We don't have to upgrade the dependency in this PR, will do it on separate one

jeanibarz: Thank you for your feedback. I restored the version to 2.2.1 then.

HenryHengZJ (approved):
thank you!

PR #3652, opened 2024-12-07
Bugfix/Prevent open connections on typeorm datasource

HenryHengZJ:


dkindlund:
@HenryHengZJ , I think this commit _might_ be why I'm running into issues here:
https://github.com/FlowiseAI/Flowise/issues/3998

PR #3653, opened 2024-12-08
feat: Add option to disable conversation history

jeanibarz:
- Add new `disableConversationHistory` boolean parameter in LLMNodes.ts and Agent.ts to optionally skip including conversation history in prompts
- Fix potential error in Agent.ts when messages array is empty by adding null safety checks
- Improve memory efficiency by allowing stateless interactions when history isn't needed

### DIFF
--- a/packages/components/nodes/sequentialagents/LLMNode/LLMNode.ts
+++ b/packages/components/nodes/sequentialagents/LLMNode/LLMNode.ts
@@ -188,6 +188,14 @@ class LLMNode_SeqAgents implements INode {
                 placeholder: 'LLM'
             },
             {
+                label: 'Disable Conversation History',
+                name: 'disableConversationHistory',
+                type: 'boolean',
+                optional: true,
+                description: `If set to true, the conversation history messages from the state will not be automatically included in the prompt`,
+                additionalParams: true
+            },
+            {
                 label: 'System Prompt',
                 name: 'systemMessagePrompt',
                 type: 'string',
@@ -535,7 +543,7 @@ async function agentNode(
         }
 
         // @ts-ignore
-        state.messages = restructureMessages(llm, state)
+        state.messages = nodeData.inputs?.disableConversationHistory | false ? [] : restructureMessages(llm, state)
 
         let result: AIMessageChunk | ICommonObject = await agent.invoke({ ...state, signal: abortControllerSignal.signal }, config)
 
--- a/packages/components/nodes/sequentialagents/Agent/Agent.ts
+++ b/packages/components/nodes/sequentialagents/Agent/Agent.ts
@@ -218,6 +218,14 @@ class Agent_SeqAgents implements INode {
                 default: examplePrompt
             },
             {
+                label: 'Disable Conversation History',
+                name: 'disableConversationHistory',
+                type: 'boolean',
+                optional: true,
+                description: `If set to true, the conversation history messages from the state will not be automatically included in the prompt`,
+                additionalParams: true
+            },
+            {
                 label: 'Human Prompt',
                 name: 'humanMessagePrompt',
                 type: 'string',
@@ -728,16 +736,16 @@ async function agentNode(
         }
 
         // @ts-ignore
-        state.messages = restructureMessages(llm, state)
+        state.messages = nodeData.inputs?.disableConversationHistory | false? [] : restructureMessages(llm, state)
 
         let result = await agent.invoke({ ...state, signal: abortControllerSignal.signal }, config)
 
         if (interrupt) {
             const messages = state.messages as unknown as BaseMessage[]
-            const lastMessage = messages[messages.length - 1]
+            const lastMessage = messages.length ? messages[messages.length - 1] : null
 
             // If the last message is a tool message and is an interrupted message, format output into standard agent output
-            if (lastMessage._getType() === 'tool' && lastMessage.additional_kwargs?.nodeId === nodeData.id) {
+            if (lastMessage && lastMessage._getType() === 'tool' && lastMessage.additional_kwargs?.nodeId === nodeData.id) {
                 let formattedAgentResult: {
                     output?: string
                     usedTools?: IUsedTool[]


jeanibarz:
When testing locally, I encountered an error with the agent tool when the "Require Approval" toggle is set to true. However, this issue occurs regardless of whether "Disable Conversation History" is set to false or true, so I suspect the problem is unrelated to the changes in this PR.
image

You can use the attached agent workflow, `Test flow Agents.json`, for verification.

Feel free to check with the joint "Test flow Agents.json" agent workflow.
[Test flow Agents.json](https://github.com/user-attachments/files/18050385/Test.flow.Agents.json)

Additionally, my existing LLM nodes/agent nodes are not updated automatically. I had to create new nodes for the "Disable Conversation History" parameter to appear under additional parameters. I'm unsure if this behavior is expected, but it happens regardless of whether I bump the component versions.

saatchi-david:
@jeanibarz 
This is great! So for example, if you only need to give a node the last message, you could disable including the full history, and manually pass just the last message right?

jeanibarz:
Yes, exactly! The feature disables the inclusion of historical messages in the prompt but still allows you to manually add specific messages—whether from the prompt or from the script that allow to create custom conversation messages. In theory, this means you could include the initial user query or even the last two messages from the state messages history, for example. Additionally, it’s worth noting that the node will still append its output to the state messages as usual.

saatchi-david:
Awesome, I had my fingers crossed that this would come soon. I think a lot of people don't realize the amount tokens they are burning behind the scenes. Hopefully just seeing the option will help increase awareness (on top of control of course).

jeanibarz:
I attempted to add a history message when the boolean disableConversationHistory is set to true. Here's what I observed:

- **Success with Human Prompt and Format Values:** Using a placeholder and the value from $flow.state.messages[-1].content, the placeholder was correctly replaced with the latest message. This approach works perfectly. Alternatively, the value can also be retrieved directly from the state or the output of another node.

- **Issue with Script Code in the "Messages History" Field:** (probably because I don't know how to use it properly though...) When I tried using state.messages[-1].content, flow.state.messages[-1].content, or $flow.state.messages[-1].content within the script code, I encountered an error stating that the key messages does not exist.
I also attempted to use a placeholder in the format prompt (similar to the human prompt), but the placeholder was not replaced with the expected value. You can see this issue in the screenshot below:
but probably because I don't know how to use it properly: when I use state.messages[-1].content or flow.state.messages[-1].content or $flow.state.messages[-1].content I get an error saying that key messages doesn't exist. I tried also using a placeholder and setting the value in format prompt, as for the human prompt, but the placeholder is not replaced with the value, see the screenshot below

image

packages/components/nodes/sequentialagents/LLMNode/LLMNode.ts
```diff
@@ -535,7 +543,7 @@ async function agentNode(
         }
 
         // @ts-ignore
-        state.messages = restructureMessages(llm, state)
+        state.messages = nodeData.inputs?.disableConversationHistory | false ? [] : restructureMessages(llm, state)
```
HenryHengZJ, line 546: I dont we can pass empty array, perhaps we need to take the last message: 
image

HenryHengZJ:
> When testing locally, I encountered an error with the agent tool when the "Require Approval" toggle is set to true. However, this issue occurs regardless of whether "Disable Conversation History" is set to false or true, so I suspect the problem is unrelated to the changes in this PR. image
> 
> Feel free to check with the joint "Test flow Agents.json" agent workflow. [Test flow Agents.json](https://github.com/user-attachments/files/18050385/Test.flow.Agents.json)
> 
> Additionally, my existing LLM nodes/agent nodes are not updated automatically. I had to create new nodes for the "Disable Conversation History" parameter to appear under additional parameters. I'm unsure if this behavior is expected, but it happens regardless of whether I bump the component versions.



> When testing locally, I encountered an error with the agent tool when the "Require Approval" toggle is set to true. However, this issue occurs regardless of whether "Disable Conversation History" is set to false or true, so I suspect the problem is unrelated to the changes in this PR. image
> 
> Feel free to check with the joint "Test flow Agents.json" agent workflow. [Test flow Agents.json](https://github.com/user-attachments/files/18050385/Test.flow.Agents.json)
> 
> Additionally, my existing LLM nodes/agent nodes are not updated automatically. I had to create new nodes for the "Disable Conversation History" parameter to appear under additional parameters. I'm unsure if this behavior is expected, but it happens regardless of whether I bump the component versions.

For approval tools, you need an agent memory attached

jeanibarz:
> > When testing locally, I encountered an error with the agent tool when the "Require Approval" toggle is set to true. However, this issue occurs regardless of whether "Disable Conversation History" is set to false or true, so I suspect the problem is unrelated to the changes in this PR. image
> > Feel free to check with the joint "Test flow Agents.json" agent workflow. [Test flow Agents.json](https://github.com/user-attachments/files/18050385/Test.flow.Agents.json)
> > Additionally, my existing LLM nodes/agent nodes are not updated automatically. I had to create new nodes for the "Disable Conversation History" parameter to appear under additional parameters. I'm unsure if this behavior is expected, but it happens regardless of whether I bump the component versions.
> 
> > When testing locally, I encountered an error with the agent tool when the "Require Approval" toggle is set to true. However, this issue occurs regardless of whether "Disable Conversation History" is set to false or true, so I suspect the problem is unrelated to the changes in this PR. image
> > Feel free to check with the joint "Test flow Agents.json" agent workflow. [Test flow Agents.json](https://github.com/user-attachments/files/18050385/Test.flow.Agents.json)
> > Additionally, my existing LLM nodes/agent nodes are not updated automatically. I had to create new nodes for the "Disable Conversation History" parameter to appear under additional parameters. I'm unsure if this behavior is expected, but it happens regardless of whether I bump the component versions.
> 
> For approval tools, you need an agent memory attached

Thank you. I tested with AgentMemory and it works when the required approval is true: thank you !

packages/components/nodes/sequentialagents/LLMNode/LLMNode.ts
jeanibarz: > I don't think we can pass an empty array; perhaps we need to take the last message  

Based on my perspective and local tests, it doesn’t seem necessary to always pass the last message. However, we do need to ensure that at least one of the following is included: the system prompt, human prompt, or message history code. Otherwise, the following error is raised:  

```
Error buildAgentGraph - Error: 400 {"error":{"message":"'messages' : minimum number of items is 1","type":"invalid_request_error"}}
```

Passing only the latest history message undermines the purpose of this feature, which is to provide the user full control over the messages passed to the node. Instead, I believe that we should allow users to explicitly define the messages to be sent.

HenryHengZJ: Right, Im thinking of whats the best way to cater to all 3 scenarios:

- taking the last message from previous node as input (i.e: `state.messages[-1]`)
- taking the user question as input
- taking all the conversation state messages (default)

So I think the best way to allow this is to have an Option field where user can select which one to be used.

jeanibarz: Thank you for the insightful suggestion! I've implemented the changes as you recommended and conducted some local tests, which indicate that everything is functioning as expected. Please review the current implementation and let me know if it meets your expectations or if there are any further adjustments you'd like to see.

jeanibarz: Here's a langsmith trace of one my workflow test: [https://smith.langchain.com/public/c04e60c1-0659-46a5-bf8f-eefb5ed312df/r](https://smith.langchain.com/public/c04e60c1-0659-46a5-bf8f-eefb5ed312df/r)

The workflow includes 4 LLM nodes: `jokeempty`, `jokeuserquestion`, `jokelastmessage`, and `jokefullhistory`. It also features an agent with a calculator, which receives only the user question.

HenryHengZJ:
rearranging order of the parameters to correctly reflect how sys prompt, history, human prompt are being sent:

image

HenryHengZJ (approved):
thank you so much @jeanibarz ! really good work

jeanibarz:
Thank you very much to you for your time and your patience ;-)

PR #3677, opened 2024-12-10
fix: followUpPrompts migration for MariaDB

karlobencic:
Fix wrong table name in AddFollowUpPrompts migration for MariaDB


HenryHengZJ (approved):
thanks @karlobencic !

PR #3680, opened 2024-12-10
Chore/Enhancement: Improve 'Require Approval' Description to Indicate MemoryAgent Dependency

jeanibarz:
This pull request updates the description of the 'Require Approval' trigger button in the Sequential Agent node settings to clarify its dependency on the MemoryAgent configuration. The updated description informs users that enabling the "Require Approval" option requires a defined MemoryAgent, helping prevent confusion and unnecessary troubleshooting.

**Changes:**
- Updated the description of the 'Require Approval' option from:
  > "Require approval before executing tools. Will proceed when tools are not called."
  
  To:
  > "Pause execution and request user approval before running tools.  
If enabled, the agent will prompt the user with customizable approve/reject options  
and will proceed only after approval. This requires a configured MemoryAgent to manage  
the state and handle approval requests.  
If no tools are invoked, the agent proceeds without interruption."

**Reason for Change:**
The original description did not provide sufficient information, leading to non-intuitive error messages when MemoryAgent was not configured. This update makes the requirement explicit, improving user experience and reducing potential errors.


HenryHengZJ (approved):
thank you @jeanibarz !

PR #3687, opened 2024-12-12
Feature: Autosave and Undo/Redo

0xi4o:
- Implements autosave in chatflow and agentflow canvas
- Implements undo/redo functionality in chatflow and agentflow canvas


HenryHengZJ:
spotted 2 issues:

1.) strange black bar at top when refresh
2.) when open a new flow, it always autosaved first:

https://github.com/user-attachments/assets/4b8c64ca-2fae-4d4b-adf5-c89243d106db

HenryHengZJ:
Also something we discussed, nodes feel pretty laggy when there are quite a lot

- [ ] memoization of the nodes and inputs

HenryHengZJ:
2 more bugs:

- [ ] as i type something and stop for seconds, it goes into autosave right, and it then unfocus the field, so I have to click the field again to make it focus, then i can start typing

https://github.com/user-attachments/assets/79356203-462d-46c4-ae21-b595cf8c746d

- [ ] open a canvas, modify some input field, wait for autosave to happens. now try pressing Ctrl + Z, it doesnt undo. However, it works when I click the Undo button

https://github.com/user-attachments/assets/73a8bc0a-0685-4f4e-9b77-296b23bea3f1

HenryHengZJ:
can play around with the background color of the toolbox in dark mode to see if white or some other dark colors suit better: 
image

0xi4o:
- [x] open a canvas, modify some input field, wait for autosave to happens. now try pressing Ctrl + Z, it doesnt undo. However, it works when I click the Undo button
- [x] canvas controls styling in dark mode
- [ ] as i type something and stop for seconds, it goes into autosave right, and it then unfocus the field, so I have to click the field again to make it focus, then i can start typing

HenryHengZJ:
1.) Delete node doesnt get autosaved:
https://github.com/user-attachments/assets/c6c20baf-bf0a-4ce6-a11d-89ac32583f57

2.) When click Save on expanded dialog, it doesnt get autosaved:
https://github.com/user-attachments/assets/eea128d1-15f8-48a8-89ae-af2cf0df7fd6

packages/ui/src/views/canvas/CredentialInputHandler.jsx
```diff
@@ -100,6 +104,7 @@ const CredentialInputHandler = ({ inputParam, data, onSelect, disabled = false }
                         <div key={reloadTimestamp} style={{ display: 'flex', flexDirection: 'row' }}>
                             <AsyncDropdown
                                 disabled={disabled}
+                                key={`${data.id}-${inputParam.name}-${data.inputs[credentialId]}`}
```
HenryHengZJ, line 107: why do we need this?

I got error when trying to create a custom assistant:
image

packages/ui/src/ui-component/dropdown/AsyncDropdown.jsx
```diff
@@ -152,6 +152,7 @@ export const AsyncDropdown = ({
                             )
                         }}
                         sx={{ height: '100%', '& .MuiInputBase-root': { height: '100%' } }}
+                        key={`${nodeData.id}-${name}-${nodeData.inputs[internalValue]}`}
```
HenryHengZJ, line 155: there are other places that uses AsyncDropdown, for example when creating custom assistant. In those cases, there will be no nodeData, then it will throw error:

image

mahmoudsorf:
This is extremely helpful!

saatchi-david:
A suggestion: it might be helpful to have a toggle to enable/disable autosave.

There are times when I make several changes and later realize I want to revert back to the previously saved version (will usually just refresh). Also times when I might demonstrate a change but not actually want it saved.

Having the ability to turn autosave off in advance would be useful in those kinds of situations.

Office's interface comes to mind:
image

HenryHengZJ:
> A suggestion: it might be helpful to have a toggle to enable/disable autosave.
> 
> There are times when I make several changes and later realize I want to revert back to the previously saved version (will usually just refresh). Also times when I might demonstrate a change but not actually want it saved.
> 
> Having the ability to turn autosave off in advance would be useful in those kinds of situations.
> 
> Office's interface comes to mind: image

in that case, does undo helps? i was thinking if you made some changes and it got autosaved, i would just Ctrl + Z to undo those changes

DigitasJapan:
Undo definitely helps! I'd still prefer a way to disable it so I can more easily jump back to the last saved version and not have to worry about forgetting to undo a change, but realize this might just be personal preference :)

saatchi-david:
Undo definitely helps! I'd still prefer a way to disable it so I can more easily jump back to the last saved version and not have to worry about forgetting to undo a change, but realize this might just be personal preference :)

PR #3702, opened 2024-12-13
[Feature] Add prompt to RetreivalQAChain

JJK801:
Hi Flowise,

This PR adds the ability to link a prompt template to `RetrievalQAChain` (it's also achievable for MultiRetrievalQAChain but this one targets `PromptTemplate` instead of `BasePromptTemplate`, so it would implies a lot on refacto)

@HenryHengZJ BTW it would be great to update `RetrievalQAChain` logic as `RetrievalQAChain` is deprecated by langchain.

### DIFF
--- a/packages/components/nodes/chains/RetrievalQAChain/RetrievalQAChain.ts
+++ b/packages/components/nodes/chains/RetrievalQAChain/RetrievalQAChain.ts
@@ -1,6 +1,7 @@
 import { BaseRetriever } from '@langchain/core/retrievers'
 import { BaseLanguageModel } from '@langchain/core/language_models/base'
 import { RetrievalQAChain } from 'langchain/chains'
+import { BasePromptTemplate } from '@langchain/core/prompts'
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
> Hi Flowise,
> 
> This PR adds the ability to link a prompt template to `RetrievalQAChain` (it's also achievable for `MultiRetrievalQAChain` but this one targets `PromptTemplate` instead of `BasePromptTemplate`, so it would implies a lot on refacto)
> 
> * [x]  Add `prompt` input (optional to prevent breaking changes and fit langchain behavior)
> * [x]  Allow return source documents for `RetrievalQAChain`
> 
> @HenryHengZJ BTW it would be great to update `RetrievalQAChain` logic as `RetrievalQAChain` is deprecated by langchain. (will be removed on 1.0.0)

would you like to take an attempt to do it in this PR? using the new way via `createRetrievalQAChain`: 
image

JJK801:
@HenryHengZJ sure i can, i'll do it by the end of week.

Issue #3707, opened 2024-12-14
[BUG] Enabling "Show Agent Reasoning" disables file upload


Tokail:
**Describe the bug**
When enabling "Show Agent Reasoning", it immediately disables file upload.

**To Reproduce**

1. Create an agent/assistant/chatflow
2. Enable file upload from the configurations
3. Save
4. Go to share chatbot
5. Make public
6. Enable "Show Agent Reasoning"
7- Visit public link
8- File upload is disabled

**Screenshots**
![image](https://github.com/user-attachments/assets/cdc121a5-5bab-46db-8803-cc5f029ba221)

**Setup**

-   Installation [cloud]

HenryHengZJ:
fix merged! https://github.com/FlowiseAI/Flowise/pull/3713

PR #3713, opened 2024-12-16
fix: mapping full file upload configuration

daniloferreira:
Fixing the bug reported at #3707


HenryHengZJ (approved):
thanks @daniloferreira !

PR #3718, opened 2024-12-16
Feature/Add Azure Cognitive speech-to-text functionality

ghondar:
Azure Cognitive Services integration for speech-to-text functionality

- Introduced a new credential class for Azure Cognitive Services.
- Updated speech-to-text processing to support Azure Cognitive Services as a provider.
- Enhanced UI components to include Azure Cognitive Services options and inputs for configuration.
- Added necessary imports and error handling for Azure API requests.
Screenshot 2024-12-16 at 5 19 08 PM


packages/components/src/speechToText.ts
```diff
@@ -3,12 +3,15 @@ import { getCredentialData } from './utils'
 import { type ClientOptions, OpenAIClient, toFile } from '@langchain/openai'
 import { AssemblyAI } from 'assemblyai'
 import { getFileFromStorage } from './storageUtils'
+import axios from 'axios'
+import FormData from 'form-data'
```
HenryHengZJ, line 7: i think this is probably redundant, I've always just directly use the built in `FormData`, can you try to remove this and see if it works fine?

ghondar: Sure, but when I remove that import, I get this error:
Screenshot 2024-12-17 at 7 57 07 PM

ghondar: fixed

HenryHengZJ (approved):
thank you so much!

PR #3728, opened 2024-12-18
Bugfix/Disabling of chatbotConfig.chatFeedback when share chatbot updated

richersoon:
This PR is similar to #3707, but this time it is in the Chatflow Configuration's Enable chat feedback instead of File Upload.

Screenshot 2024-12-18 at 9 18 23 AM


HenryHengZJ:
thanks @richersoon !

After a closer look, I've realized it will override any config saved, not just feedback, and file uploads. Here's the correct fix: https://github.com/FlowiseAI/Flowise/pull/3742

Closing for now

PR #3729, opened 2024-12-18
Bugfix/Hide agent response for chatflow

richersoon:
`Show Agent Reasoning` is not saving for chatflows, based on code line number 150:

```
if (isAgentCanvas) {
            // if showAgentMessages is undefined, default to true
            if (showAgentMessages === undefined || showAgentMessages === null) {
                obj.showAgentMessages = true
            } else {
                obj.showAgentMessages = showAgentMessages
            }
        }
```

it is only applicable for Agentflows, so maybe better to hide it for Chatflows.


HenryHengZJ (approved):
thanks @richersoon

PR #3747, opened 2024-12-19
[BUGFIX] Fix value.startsWith is not a function error

simsketch:
There is an issue in 2.2.1 where using AnthropicChat in an agent flow with a tool was causing an error because it wasn't setup to handle an array.

### DIFF
--- a/packages/components/nodes/sequentialagents/commonUtils.ts
+++ b/packages/components/nodes/sequentialagents/commonUtils.ts
@@ -104,6 +104,9 @@ export const transformObjectPropertyToFunction = (obj: ICommonObject, state: ISe
                     if (typeof messageOutput.content === 'string') value = messageOutput.content
                     // if messageOutput.content is an array
                     else if (Array.isArray(messageOutput.content)) {
+                        if (messageOutput.content.length === 0) {
+                            throw new Error(`Message output content is an empty array for node ${parsedValue.id}`)
+                        }
                         // Get the first element of the array
                         const messageOutputContentFirstElement: any = messageOutput.content[0]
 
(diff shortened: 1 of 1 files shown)


packages/components/nodes/sequentialagents/commonUtils.ts
```diff
@@ -99,7 +99,22 @@ export const transformObjectPropertyToFunction = (obj: ICommonObject, state: ISe
                     (message) => message.additional_kwargs && message.additional_kwargs?.nodeId === parsedValue.id
                 )
                 const messageOutput = messageOutputs[messageOutputs.length - 1]
-                if (messageOutput) value = messageOutput.content
+                if (messageOutput) {
+                    // if messageOutput.content is a string, set value to the content
+                    if (typeof messageOutput.content === 'string') value = messageOutput.content
+                    // if messageOutput.content is an array
+                    else if (Array.isArray(messageOutput.content)) {
```
HenryHengZJ, line 106: should we also check if the array is not empty?

simsketch: yeah that makes sense... standby

simsketch: Hmm... should we throw an error if it's empty or...?

simsketch: that's what i added in there

HenryHengZJ: yeah we can throw an error

HenryHengZJ (approved):
thannks @simsketch !

PR #3761, opened 2024-12-23
fix: getting docId, storeId parameters

hardviper:
The solution that was in the release caused an error in vectostoreconfig if the flowise runned not in the root URL, but in a subpath. I propose an option with the correct receipt of parameters that are provided by the route.


HenryHengZJ (approved):
thanks @hardviper !

PR #3771, opened 2024-12-26
feat(components/storage): simplify file mangment

ahmedrowaihi:
this file is huge, I simplifed with [flydrive](https://flydrive.dev/docs/introduction)

I need your help in confirming functionality that it's working correct for all cases


HenryHengZJ:
interesting, the library seems to be quite new, and is it actively maintained?

ryanhalliday:
Maybe not until this gets more real world tested? Seems like no major gain to be an early adopter here. 

Most of those storage providers already have an S3 compatible interface right? So no action needed to actually support them.

PR #3780, opened 2024-12-28
Feat/new session btn

khiemfle:
Feature: support creating new internal chat session via the chat box instead of clearing the chat history.
Purpose: we may like to test with new flow configuration but still like to see the history of previous configs for the comparison.
 
Screenshot 2024-12-28 at 10 09 11


HenryHengZJ:
thanks @khiemfle ! Though I think it would be better to be consistent with the UI, we can have another button beside clear message:
image

When expanded, it should be shown beside Clear Messages button:
image

PR #3787, opened 2025-01-01
Chore: Update @langchain/community

PylotLight:
- Updated @langchain/community and @langchain/core pkgs in components dir
- Fix breaking change for Jina which now references apikey directly in the interface


HenryHengZJ (approved):
thanks @PylotLight !

PR #3789, opened 2025-01-01
bugfix: Add nil check in escapeAllStrings to prevent TypeError when upsert vectors

nguyenhung10012003:
There is an issue when upsert vectors using redis vector store. I encountered this problem when doing document upsert from some PDF files. Not all PDF files are affected, just some

image

This pull request includes changes to the `packages/components/nodes/vectorstores/Redis/utils.ts` file to improve error handling and ensure that null or undefined objects are properly managed.


HenryHengZJ (approved):
thanks @nguyenhung10012003 !

Issue #3832, opened 2025-01-08
[FEATURE] Automatically delete messages from "View Messages" after a timeout


dentroai:
**Describe the feature you'd like**
I don't want to store all messages by user on my flowise instance.
Allow me to set a timeout (or set a fixed timeout) that deletes messages after e.g. 2 days.

**Additional context**
At "View Messages" I can see all messages that were sent to my Flow:
![image](https://github.com/user-attachments/assets/a0735b86-27a9-4ee4-b603-f592e6239ecc)

I want that all those messages are deleted from flowise automatically. 

The reason is data security. The less data is stored, the less data can be leaked #Germany

ryanhalliday:
You could set this up with a cron job script to access and clear the DB. Would be nice to include in Flowise core though. 

I also have a PR open right now to entirely prevent Flowise from storing message logs if that is the configuration you need: https://github.com/FlowiseAI/Flowise/pull/3837

PR #3837, opened 2025-01-09
Disable message saving to DB via an environment variable

ryanhalliday:
Add an environment variable `DISABLE_MESSAGE_SAVING` which disables messages being saved to the database. This is useful for scenarios where you are avoiding saving potential PII to your Flowise instance. 

This disables the buffer memory nodes. If they were used in a Flow it will no longer work and will instead show an error on run and in the UI. Because this is a very serious environment variable to enable, so I think this is an okay result. 
image

Analytics integrations continue to work if enabled per flow.

Feedback will no longer work but I have not added error messages for that. I could add a validation to prevent Feedback being toggled on? 

The messages UI will show a different error message:
image

Future upgrades to this may shift this from an environment variable to a per-flow setting but this is the simpler approach for now. 

Please let me know if you would like any changes to this or there are any other potential areas impacted!

cc @tgmerritt @HenryHengZJ


dentroai:
oh great if this PR is doing it's work it would solve https://github.com/FlowiseAI/Flowise/issues/3832 !

dentroai:
ok I see. 
For the use cases I have in my mind this should be ok!

For use cases where citations or tools have to be displayed on the chat, one would need to add custom logic to delete a message older than X. Via a cronjob for example.

PR #3864, opened 2025-01-13
Bugfix/Handling empty strings when resolving variables

supraface:
There are some issues. 

1. Not considering empty string when LLMChain process the given promptValues. It makes the 'seen' stack do unexpected things

2. The value of the variable is the intended empty string, but it doesn't resolve.

- Expected: Sending a request using a prompt where the `emptystring` variable (the output of a custom function) is replaced with an empty string.
스크린샷 2025-01-14 오전 1 33 01

- Current behavior: The value of `question` is repeatedly assigned to a variable that should be assigned an empty string
스크린샷 2025-01-14 오전 1 32 47

### DIFF
--- a/packages/components/nodes/chains/LLMChain/LLMChain.ts
+++ b/packages/components/nodes/chains/LLMChain/LLMChain.ts
@@ -249,7 +249,7 @@ const runPrediction = async (
 
         for (const variable of inputVariables) {
             seen.push(variable)
-            if (promptValues[variable]) {
+            if (promptValues[variable] != null) {
                 seen.pop()
             }
         }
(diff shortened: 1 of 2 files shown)


HenryHengZJ:
in what scenarios you would want an empty string to be passed in as variable value?

supraface:
> in what scenarios you would want an empty string to be passed in as variable value?

Optional promptValues.

Some promptValues should be optional, for example, hints could be added to the prompt that are not always present but would help with generation if added.
Of course, you can undefine the field when triggering the chatflow, but it's obviously weird to have an explicit blank replaced by the value of another field.

HenryHengZJ (approved):
thanks for clarifying @supraface !

PR #3867, opened 2025-01-14
Get/Delete ChatMessage based on startDateTime / endDateTime

ryanhalliday:
supplements existing startDate / endDate fields

### DIFF
--- a/packages/server/src/controllers/chat-messages/index.ts
+++ b/packages/server/src/controllers/chat-messages/index.ts
@@ -71,6 +71,8 @@ const getAllChatMessages = async (req: Request, res: Response, next: NextFunctio
         const messageId = req.query?.messageId as string | undefined
         const startDate = req.query?.startDate as string | undefined
         const endDate = req.query?.endDate as string | undefined
+        const startDateTime = req.query?.startDateTime as string | undefined
+        const endDateTime = req.query?.endDateTime as string | undefined
         const feedback = req.query?.feedback as boolean | undefined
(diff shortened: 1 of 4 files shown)


ryanhalliday:
Also added some handling for zero results on the delete query - behaviour is generally the same just http status changes from 500 with error message to 200 with `affected: 0`

packages/server/src/controllers/chat-messages/index.ts
```diff
@@ -158,6 +166,8 @@ const removeAllChatMessages = async (req: Request, res: Response, next: NextFunc
         const _chatType = req.query?.chatType as string | undefined
         const startDate = req.query?.startDate as string | undefined
         const endDate = req.query?.endDate as string | undefined
+        const startDateTime = req.query?.startDateTime as string | undefined
```
HenryHengZJ, line 169: why do we need `startDateTime` and `endDateTime` since we already have `startDate` and `endDate`? wouldnt that be confusing which to use?

HenryHengZJ: I see, the new ones include time. I think for consistency, we can just use `startDate`and `endDate` but with precision down to time, instead of just dates

ryanhalliday: Yeah I like that, I didn't want to change the API much in case others are using it but that would be much better. 

We're looking for querying down to the minute level so whatever gets us there is perfect!

HenryHengZJ:
Update PR to enable using `startDate` and `endDate` including time, hours, minutes, milliseconds:

image

PR #3875, opened 2025-01-15
Bugfix/add hidden properties to custom tool

HenryHengZJ:
image

This is done so users can enable overrideconfig:
image


heslanx:
Thank you @HenryHengZJ for the quick resolution! 👏 I'll test the fix in our next deployment and confirm if it resolves the override configuration issue. Appreciate the detailed work on addressing this bug.

PR #3891, opened 2025-01-17
Mssql integration

apollo4040:
Here I add mssql integration with Init migration. This integration was tested and works as expected. Please review!


HenryHengZJ:
thanks! we also probably need to write all of the migration scripts [here](https://github.com/FlowiseAI/Flowise/tree/main/packages/server/src/database/migrations/mysql)

for now we prob wont be supporting another database as we are working on some changes to some of the tables

HenryHengZJ:
we can keep this PR open and revisit soon!

Issue #3899, opened 2025-01-20
[BUG] S3LoggerStream config fails when using any S3 provider other than amazonaws.


touatiosema:
**Describe the bug**
In the logger config `/packages/server/src/utils/logger.ts`, the logger configuration is not using the custom Url as specified in the .env file (the env variable process.env.S3_ENDPOINT_URL)

**To Reproduce**

1. Go to `.env` file and set up an S3 provider other than amazon
2. try to run the backend server
3. the backend logger defaults to amazonaws as the custom url S3_ENDPOINT_URL is not configured.
4. server crashes when it tries to resolve the dns as in the error messsage:
```log
2025-01-20 14:30:40 [ERROR]: unhandledRejection: getaddrinfo ENOTFOUND <bucket>.s3.<region>.amazonaws.com
Error: getaddrinfo ENOTFOUND <bucket>.s3.<region>.amazonaws.com
    at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:120:26)
    at GetAddrInfoReqWrap.callbackTrampoline (node:internal/async_hooks:130:17)
```

**Expected behavior**
The server is expected to start with the S3 store configured as a store for log file.

**Setup**

-   configure custom S3 in .env (other than aws's S3).
-   `cd packages/server && pnpm run dev`
-   Flowise Version 2.2.4
-   OS: Linux

PR #3900, opened 2025-01-20
fix: Fixing logger for custom S3 stream

touatiosema:
This PR solves the issue mentionned on #3899, The issue is that when trying to plug any S3 provider other than amazon aws, the server startup fails in dns request (check the issue for more info).

This pr solves the issue by using the custom s3 endpoint present in the env variables.


HenryHengZJ (approved):
thanks @touatiosema !

PR #3903, opened 2025-01-21
fix: Correct logic error in tool input assignment for ChatflowTool

kirisame-wang:
According to the description of `customInput`, leaving it empty allows the LLM to decide the input. Therefore, if `customInput` is not an empty string, it should be passed to the chatflow.
The previous code incorrectly assigned `toolInput` to an empty string when `customInput` was non-empty. This has been corrected to ensure that `toolInput` is assigned `customInput` when `useQuestionFromChat` is false and `customInput` is non-empty.


HenryHengZJ (approved):
thanks for the fix!

kirisame-wang:
Thank you as well! I'm thrilled to have made my first contribution to an open source project.

PR #3909, opened 2025-01-22
add DeepSeek-R1

toi500:
image


toi500:
Just tested and it works as expected. What I don't see is showing the CoT itself, just the final response, so not sure if this it is a issue or a current display limitation.

image

HenryHengZJ:
thanks @toi500 ! merged https://github.com/FlowiseAI/Flowise/pull/3911

PR #3957, opened 2025-01-31
feat: preserve UI state memory on API override

grapestore:
- When the state memory is overridden by an API call, the existing UI-based settings get lost. 
This commit ensures that any UI-defined state persists even if an API override occurs.
- The reason for adding this feature is that, in addition to the input values provided by the API, there are agent-specific states required for maintaining context. We believe having this functionality in place would be highly beneficial.


HenryHengZJ (approved):
thank you !

PR #3959, opened 2025-01-31
Fix: Ensure normalizeURL retains port number

jeme95:
## Issue
The normalizeURL function was incorrectly removing the port from URLs.

## Fix
- Explicitly include the port if it's present (`urlObj.port`)
- Prevents unintended behavior where ports (e.g., `:3000`) were stripped

## Example
Before:
normalizeURL("http://example.com:3000/path?query=1") → "example.com/path?query=1" (port missing)

After:
normalizeURL("http://example.com:3000/path?query=1") → "example.com:3000/path?query=1" (port retained)


HenryHengZJ (approved):
thank you !

PR #3989, opened 2025-02-05
Several features for OpenAPI toolkit and OpenAI Assistants

owengo:
The PR proposes to:

1. Allow OpenAI assistant to call several tools in the same run. This is quite common when the assistant decides to call a second tool after the first one
2. The OpenAPI Toolkit now supports a "x-strict": true  attribute for an action. This creates a function with "strict mode".

The "strict mode" is described here: https://openai.com/index/introducing-structured-outputs-in-the-api/ , https://platform.openai.com/docs/guides/structured-outputs 
Basically it allows to drastically reduce hallucination problems while generating tool calls.

The strict mode has the problem that only a subset of the json spec is supported.
One of the caveats is that all fields must be required. So the workaround is to use an "anyOf" schema with the type null.

For example if we want to make the "limit" parameter optional we can specify it like this:
```
{
            "name": "limit",
            "in": "query",
            "required": true,
            "schema": {
              "anyOf": [
                {
                  "type": "integer"
                },
                {
                  "type": "null"
                }
              ],
              "description": "Maximum number of records to return (1-100)",
              "title": "Limit"
            },
            "description": "Maximum number of records to return (1-100)"
          },
```
But then the model will call the api with "limit=null" .
So the PR adds the option to remove null parameters from the api call, so that your api backend won't choke on the null parameter.

### DIFF
--- a/packages/components/nodes/agents/OpenAIAssistant/OpenAIAssistant.ts
+++ b/packages/components/nodes/agents/OpenAIAssistant/OpenAIAssistant.ts
@@ -924,14 +924,12 @@ async function handleToolSubmission(params: ToolSubmissionParams): Promise<ToolS
         chatId,
         options,
         input,
-        usedTools,
-        text,
-        isStreamingStarted
+        usedTools
     } = params
 
     let updatedText = params.text
     let updatedIsStreamingStarted = params.isStreamingStarted
-    
+
     const stream = openai.beta.threads.runs.submitToolOutputsStream(threadId, runThreadId, {
         tool_outputs: submitToolOutputs
     })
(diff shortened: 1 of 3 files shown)


owengo:
Sorry but I can't reproduce the error, the cypress tests do pass, it's the nodejs cleanup action which seems to have failed

HenryHengZJ:
can you pull the latest changes? it should be fixed

AsharibAli:
Amazing work, guys!

0xi4o (approved):

0xi4o (approved):

HenryHengZJ (approved):
works like a charm!
image

truly appreciate the work you put in @owengo !

PR #3999, opened 2025-02-06
Update README.md

mikz0rz:
Fixed deployment section links, /configuration/ part was missing from the URLs


HenryHengZJ (approved):
thanks!

PR #4009, opened 2025-02-08
o3-mini Adjustments.

saatchi-david:
Several small adjustments for o3-mini

**Update ChatOpenAI.ts - corrected reasoning description and default**
- Description for reasoning effort only mentioned o1. Added o3.
- Changed reasoning effort default to medium as this is OpenAI's default / what users will most likely expect (https://platform.openai.com/docs/guides/reasoning)

**Update AzureChatOpenAI.ts - corrected reasoning description and default**
- Description for reasoning effort only mentioned o1. Added o3.
- Changed reasoning effort default to medium as this is OpenAI's default / what users will most likely expect (https://platform.openai.com/docs/guides/reasoning)

**Update models.json - add specific model ID for o3-mini**
- Added o3-mini-2025-01-31
- Updated "o3-mini" label to "o3-mini (latest)"
This allows the user to choose a specific model ID and avoid the risk of unexpected behavior if the "o3-mini" alias is updated.

### DIFF
--- a/packages/components/models.json
+++ b/packages/components/models.json
@@ -613,8 +613,12 @@
                     "name": "gpt-4o-2024-05-13"
                 },
                 {
-                    "label": "o3-mini",
+                    "label": "o3-mini (latest)",
                     "name": "o3-mini"
+                },
+                {
+                    "label": "o3-mini-2025-01-31",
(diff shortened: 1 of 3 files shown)


HenryHengZJ (approved):
thanks man!

PR #4061, opened 2025-02-22
Feat: Support Google Cloud Storage

AllenFang:
hi team, I would like to make Flowise to support GCS as one of storage type so please help to me review this PR. I've been tested my work a lots, if you need a demo as the screenshot or screen recording, please let me know. 

furthermore, if i'm lucky that this MR could be merged. I will help to refactor the code a bit especially the storageUtil.ts.


0xi4o (approved):

0xi4o:
Ah ok. I will wait for your updates @AllenFang.

AllenFang:
hi @0xi4o 
I just notice that your conflict fix seems mess up the lock file
it's my previous change: https://github.com/FlowiseAI/Flowise/pull/4061/files#diff-32824c984905bb02bc7ffcef96a77addd1f1602cff71a11fbbfdd7f53ee026bb

But after your fix, it seems change a lots (not sure whether it is cause by indent space?

AllenFang:
never mind, I get it back and fix the logger part. @0xi4o , please take a look. thank you

HenryHengZJ:
The paths for GCS are not normalized, causing the uploaded file become a single path file with backslash, ideally it should be forward slash and stored in folders:
image

I've included the fix

HenryHengZJ (approved):

PR #4068, opened 2025-02-24
Fix: ignore skipInferTableTypes on document storage process

sunnymodi21:


HenryHengZJ:
I changed it to check if the `fileNames` has item, because if we already have a regex check above to see if the content is base 64 data.

@sunnymodi21 can you check to see if this new approach works?

sunnymodi21:
Yes, this logic is better and works. Thanks!

Issue #3997, opened 2025-02-06
[BUG] AWS S3 credentials issue in 2.2.5


danieldabate:
**Describe the bug**
Flowise crashes after updating to 2.2.5 with the following error:

> Error: Resolved credential object is not valid                                                                                                                                                                                           
> at SignatureV4S3Express.validateResolvedCredentials (/usr/local/lib/node_modules/flowise/node_modules/@smithy/signature-v4/dist-cjs/index.js:553:13)
> at SignatureV4S3Express.signRequest (/usr/local/lib/node_modules/flowise/node_modules/@smithy/signature-v4/dist-cjs/index.js:478:10)                                                                                                  at process.processTicksAndRejections (node:internal/process/task_queues:95:5)                                                                                                                                                         at async AwsSdkSigV4Signer.sign (/usr/local/lib/node_modules/flowise/node_modules/@aws-sdk/core/dist-cjs/submodules/httpAuthSchemes/index.js:106:27) 
> at async /usr/local/lib/node_modules/flowise/node_modules/@aws-sdk/middleware-sdk-s3/dist-cjs/index.js:481:15                                                                                                                         at async /usr/local/lib/node_modules/flowise/node_modules/@smithy/middleware-retry/dist-cjs/index.js:321:38                                                                                                                           at async /usr/local/lib/node_modules/flowise/node_modules/@aws-sdk/middleware-flexible-checksums/dist-cjs/index.js:315:18                                                                                                             at async /usr/local/lib/node_modules/flowise/node_modules/@aws-sdk/middleware-sdk-s3/dist-cjs/index.js:109:22
> at async /usr/local/lib/node_modules/flowise/node_modules/@aws-sdk/middleware-sdk-s3/dist-cjs/index.js:136:14                                                                                                                         at async /usr/local/lib/node_modules/flowise/node_modules/@aws-sdk/middleware-logger/dist-cjs/index.js:33:22

Reverting back to 2.2.4 solves the issue.   

We are currently setting up the [credentials using service accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html).

I haven't had time to investigate the issue further. But was there any changes in 2.2.5 that could affect the AWS SDK or the S3 logic?

**Setup**

-   Installation [K8S/Docker]
-   Flowise Version [2.2.5]
-   OS: [Linux]
-   Browser [-]

HenryHengZJ:
[...]

PR #4114, opened 2025-03-03
[Fix] Change S3 logger to accept more authorization methods for AWS

danieldabate:
This PR solves issue #3997 by making environment credentials optional for S3 logger, so they can be taken from different alternative sources like K8S service accounts.

### DIFF
--- a/packages/server/src/utils/logger.ts
+++ b/packages/server/src/utils/logger.ts
@@ -32,7 +32,7 @@ if (process.env.STORAGE_TYPE === 's3') {
             secretAccessKey
         }
     }
-    
+
     const s3Config: S3ClientConfig = {
         region: region,
         endpoint: customURL,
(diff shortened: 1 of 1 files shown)


saqib-ahmed:
@danieldabate @HenryHengZJ 

Can we please add unit/integration tests to prevent this sort of regression in future?

HenryHengZJ:
However, others still have the same problem even though with this fix, are you guys still experiencing the problem? I was suspecting because it doesnt accept `credentials: undefined`?

I opened a PR (https://github.com/FlowiseAI/Flowise/pull/4166) to only set `credentials` when both keys are found

danieldabate:
How have you tested it? It is not yet included in the latest release, right?

HenryHengZJ:
Yep, its not in open source yet, but I've tried to push this in a separate private docker image for enteprise, same problem

danieldabate:
I tested locally in a rush and probably did something wrong, sorry for the trouble 🙏

HenryHengZJ:
nono i was just wondering thats all, because I was not able to test it out, so asking here and see if others still faces same problem

PR #4128, opened 2025-03-05
fix metadata filtering on Weaviate

gururise:
Per this [discussion](https://github.com/FlowiseAI/Flowise/discussions/4118) the metadata filtering in Weaviate node was broken.  

I've fixed it to default to an Equal operator given the {key:value} pair in the UI.

In order to allow for more complex metadata filtering in Weaviate, the UI may need to be updated.


HenryHengZJ:
hey thanks for the PR!

one thing that could be a concern is because if we update everything to have Equal operator, that means user can't really do any other operator.

our initial intention is to let user specify the filter by themselve, but you'll have to refer to the vector database provider for the correct syntax, as each vectordb filtering syntax could be different

PR #4130, opened 2025-03-06
feat: Add SSL configuration option for Postgres connections

mikemanyoung:
Add SSL option to vector store and record manager to be consistent with the postgres database driver. Most commercial postgres implementation will require us to use SSL.


tgabi333:
👍

HenryHengZJ (approved):
thanks @mikemanyoung !

PR #4162, opened 2025-03-12
enhance json parses

iEgit:
reopened from https://github.com/FlowiseAI/Flowise/pull/4125

This PR improves the JSON parsing logic in StructuredOutputParser.ts to handle malformed JSON responses from LLMs more effectively.

Changes:
• Fixed JSON extraction logic:
• Previously, JSON extraction relied on splitting text using code block markers (`json).
• Now, it uses a regex (/(?:^[^{[])|(?:[^}\]]$)/g) to remove any leading or trailing non-JSON content, ensuring a more robust parsing approach.

Why this change?
• The previous approach failed in some edge cases where JSON responses did not follow the expected format.
• The new regex-based approach ensures that only valid JSON content is passed to jsonrepair, reducing the likelihood of parsing errors.

Testing:
• Manually tested with different malformed JSON outputs to verify correct parsing.
• Ensured existing functionality remains intact.

### DIFF
--- a/packages/components/nodes/outputparsers/StructuredOutputParser/StructuredOutputParser.ts
+++ b/packages/components/nodes/outputparsers/StructuredOutputParser/StructuredOutputParser.ts
@@ -79,7 +79,7 @@ class StructuredOutputParser implements INode {
 
             // Fix broken JSON from LLM
             structuredOutputParser.parse = (text) => {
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
Do you have some examples where the old approach is not able to parse, but solvable with the newer approach?

PR #4163, opened 2025-03-12
Fix Agent Chat History Retrieval

iEgit:
fixes the retrieval of agent chat history from the API. 

The issue was caused by referencing the incorrect input key (memory instead of agentMemory) in getChatHistory. This update ensures that agent memory is correctly accessed, preventing chat history from being lost or incorrectly fetched.

Changes:
	•	Updated the input reference from memory to agentMemory in buildChatflow.ts.
	•	Ensured that when agentMemory is not present, the function returns prependMessages instead of an empty array.


HenryHengZJ (approved):
thank you!

PR #4169, opened 2025-03-13
Add an Ollama baseUrl in chatflow configuration for Follow-up prompts

SargeJoseph:
When using Ollama for the follow-up prompts, you sometimes need to specify a different baseURL instead of http://127.0.0.1:11434

This is especially usefull when:

Ollama is not used in the Chatflow but you still want to use it for follow-up prompts.
Flowise is running in Docker


HenryHengZJ (approved):
thank you @SargeJoseph

PR #4190, opened 2025-03-17
docs: modify Simplified Chinese characters

dushaobindoudou:
The correct way to write Simplified Chinese is 简体中文


HenryHengZJ:
thanks, already merged: https://github.com/FlowiseAI/Flowise/pull/4229

Issue #4107, opened 2025-03-02
[BUG] PDF Document Loader cannot upload files with a comma in the file name


damirtresnjo:
I noticed an issue when you are running flowise locally on a windows 11 machine. You cannot upload a PDF file with commas in the file name into a PDF loader under Document Store.

**To Reproduce**

1. Go to Document Stores
2. Add a store
3. Add a PDF loader
4. Click on upload a file
5. Select a file with a , in the file name (e.g. test,1.pdf)
6. No error, just no filename is selected

**Setup**

-   Installation: npx flowise start
-   Flowise Version: 2.2.6-hotfix.1
-   OS: Windows 11
-   Browser Chrome

Hetsavani:
Hey can anyone review my pull request and tell me the feedback please?

PR #4197, opened 2025-03-19
Fixed: Document Loader Cannot Upload File with comma

Hetsavani:
fixes: #4107 

Screenshot 2025-03-19 135625


HenryHengZJ:
thanks for the fix! we probably also need to handle the case for multiple file uploads

but files with comma name is not generally a common case, its better to rename the file without comma

Hetsavani:
> thanks for the fix! we probably also need to handle the case for multiple file uploads
> 
> but files with comma name is not generally a common case, its better to rename the file without comma

I couldn't understand the issue
can you pls elaborate?

Hetsavani:
I have fixed issue happening on multiple file upload. Kindly check the last commit and tell me if there's any changes or feedback

PR #4213, opened 2025-03-21
Adds Mem0 Memory Node

whysosaket:
Adds [Mem0](https://github.com/mem0ai/mem0) Memory Node on Top of Buffer Memory

- Mem0 is a self-improving memory layer for LLM applications and AI Agents
- Implementation of Mem0 differs from that of other implementation, while other memory providers treat Chat History as Memory, Mem0 acts as in independent memory layer on top of Buffer Memory to provide facts about the entity.

Screenshot 2025-03-21 at 7 59 44 PM


toi500:
Very needed. Thank you for this.

whysosaket:
> Very needed. Thank you for this.

Yup! It should be available soon

whysosaket:
Hey @jimjimovich @matthias @appleboy, we're almost done with the integration.

Just needed some clarity for better consumer experience. Is there a way to get the hold of inputMessage inside of `getChatMessages` ? This would allow us to perform the `search` operation from Mem0 which fetches better and more relevant memories. Right now we return all memories that the user has, which is not the best solution.

Wanted to know if there is a way to get this without changing the base class.

Thanks

edvinPL:
> Hey @chungyau97 @toi500, we're almost done with the integration.

You could potentially use $flow.input ?
Screenshot_20250328_120131_com android chrome~2

chungyau97:
Hi @whysosaket,

Everything is working as expected, as shown in the screenshots below.
image
image

Could you add Mem0's documentation (including description and setup process) to [FlowiseAI's documentation under memory](https://github.com/FlowiseAI/FlowiseDocs/tree/main/en/integrations/langchain/memory)?

chungyau97 (approved):

whysosaket:
Hi @chungyau97,

I have added the docs. See here: https://github.com/FlowiseAI/FlowiseDocs/pull/159

Let me know if any other changes are required.

KaliLinuxPentest:
Can you write appropriate Use cases for this tool? I don't really understand how I should use it yet....

edvinPL:
@whysosaket any way to modify the size of context within mem0 node?? E.g. Having last K-exchanges as-is for better agentic recollection?
image

bengoetzinger:
Hello,

I (or rather, Claude has...) have made a custom node to connect to my self-hosted mem0 memory storage. I tried it out in a locally deployed flowise on my MacBook by adding it in /Flowise/packages/components/nodes/memory/Mem0/Mem0Custom.ts, and it does show in the side menu.
Now I tried to use it in my production server which runs various flowise instances in Docker containers. I tried to do the same there, but have gotten nowhere. I even tried to substitute the existing mem0-node with my code, but it stays the same that connects to standard mem0-service.

Any help highly appreciated!

Best regards
Ben

[Mem0Custom.ts](https://github.com/user-attachments/files/23900829/Mem0Custom.ts)

PR #4224, opened 2025-03-27
FIXED Document path error & simplified Chinese font is incorrect

codeagles:
[fix] Document path error & simplified Chinese font is incorrect


HenryHengZJ (approved):
thanks!

PR #4257, opened 2025-04-04
feat(Mem0): Add option to use Flowise Chat ID

toi500:
**Description:**

This PR enhances the Mem0 memory node by adding a new toggle option: "Use Flowise Chat ID".

*   **When Enabled:** The node will ignore the hardcoded "User ID" input field and use the internal Flowise `chatId` (session ID) as the `user_id` for all Mem0 operations (saving context, loading variables, clearing memory).
*   **When Disabled (Default):** The node behaves as before, using the value provided in the "User ID" input field for Mem0 operations.

**Changes:**

*   Added a "Use Flowise Chat ID" boolean input toggle to the Mem0 node UI.
*   Updated internal logic (`getEffectiveUserId`) to select the correct ID (Flowise `chatId` or the input field value) based on the toggle state.
*   The "User ID" input field is now only strictly required if the "Use Flowise Chat ID" toggle is disabled.

This provides users with more flexibility, allowing them to easily tie Mem0 memory directly to Flowise chat sessions without manual configuration if desired.

image

image

### DIFF
--- a/packages/components/nodes/memory/Mem0/Mem0.ts
+++ b/packages/components/nodes/memory/Mem0/Mem0.ts
@@ -49,9 +49,17 @@ class Mem0_Memory implements INode {
                 label: 'User ID',
                 name: 'user_id',
                 type: 'string',
-                description: 'Unique identifier for the user',
+                description: 'Unique identifier for the user. Required only if "Use Flowise Chat ID" is OFF.',
                 default: 'flowise-default-user',
-                optional: false
+                optional: true
+            },
+            {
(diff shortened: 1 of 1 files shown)


HenryHengZJ (approved):
Thanks @toi500 , looks good!

toi500:
I have tested it on prpduction too, and it works as expected. This modification allows us to segregate groups of memories, as per example, an email AI agent that wants to access user-annotated memories per email address or by user name if the customer is logged in, etc.

PR #4260, opened 2025-04-06
feat(feedback): add validation for feedback creation and update

MehdiCHEBBAH:
- Introduced validation functions for feedback creation and update in the new validation service. 
- Updated feedback controller to utilize these validation functions before processing requests.


HenryHengZJ (approved):
thanks @MehdiCHEBBAH !

PR #4271, opened 2025-04-09
Update how credential data is saved in db when using aws secrets manager

0xi4o:
Also accounts for existing users using AWS Secrets Manager to store credentials data.


packages/server/src/utils/index.ts
```diff
@@ -1394,6 +1394,29 @@ export const getEncryptionKey = async (): Promise<string> => {
     if (process.env.FLOWISE_SECRETKEY_OVERWRITE !== undefined && process.env.FLOWISE_SECRETKEY_OVERWRITE !== '') {
         return process.env.FLOWISE_SECRETKEY_OVERWRITE
     }
+    if (USE_AWS_SECRETS_MANAGER && secretsManagerClient) {
```
HenryHengZJ, line 1397: lint error, 'PutSecretValueCommand' is defined but never used

0xi4o: Fixed

HenryHengZJ (approved):

PR #4273, opened 2025-04-09
feat:Adds new Param in getChatMessages for better context handling

whysosaket:
This is a proposal to introduce a new parameter in the `getChatMessages` function that allows us to retrieve the latest user message, enabling more accurate memory retrieval. Currently, Mem0 Memory uses `get_all` to fetch all user memories, but this change would allow us to perform a semantic search, which is more efficient for handling context.

I'd like to understand if there's a better approach to achieve this, and whether it makes sense to apply the same logic to other components as well. Thanks!

### Changelog

- Enhanced `getChatMessages` Method: Introduced a new parameter `currentMessages` to allow for dynamic message retrieval based on the latest user input, improving the context handling during chat sessions.
- Improved Message Handling: Updated logic to prepend messages from `currentMessages` to the chat history, ensuring that the latest user input is considered when fetching the memories.

### DIFF
--- a/packages/components/nodes/chains/ConversationChain/ConversationChain.ts
+++ b/packages/components/nodes/chains/ConversationChain/ConversationChain.ts
@@ -24,7 +24,8 @@ import {
     INodeData,
     INodeParams,
     MessageContentImageUrl,
-    IServerSideEventStreamer
+    IServerSideEventStreamer,
+    IMessage
 } from '../../../src/Interface'
 import { ConsoleCallbackHandler, CustomChainHandler, additionalCallbacks } from '../../../src/handler'
 import { getBaseClasses, handleEscapeCharacters, transformBracesWithColon } from '../../../src/utils'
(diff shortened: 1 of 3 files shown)


whysosaket:
Hi @chungyau97,
Could you please review this and confirm that this method works or is there a better way to handle this use-case?

whysosaket:
Hey @chungyau97,
Can you please help here?

whysosaket:
Hey @HenryHengZJ , can you help here?

HenryHengZJ:
hey sorry for late reply. I think a better approach is to use `getSessionChatHistory` and pass in the `incomingInput`. Then in the `getSessionChatHistory` function, pass the `incomingInput` to `const initializedInstance: FlowiseMemory = await newNodeInstance.init(memoryNode.data, incomingInput, {`

Finally, in the mem0 you can then take the input from init function `async init(nodeData: INodeData, input: string, options: ICommonObject): Promise<any> {`

whysosaket:
> hey sorry for late reply. I think a better approach is to use `getSessionChatHistory` and pass in the `incomingInput`. Then in the `getSessionChatHistory` function, pass the `incomingInput` to `const initializedInstance: FlowiseMemory = await newNodeInstance.init(memoryNode.data, incomingInput, {`
> 
> Finally, in the mem0 you can then take the input from init function `async init(nodeData: INodeData, input: string, options: ICommonObject): Promise<any> {`

This really helped a lot, was just a few lines of change, super thanks!!

image

HenryHengZJ (approved):
thanks!

Issue #4243, opened 2025-04-01
[BUG] Document Store [Save Config] button is not working as desired


almirb:
**Describe the bug**
When clicking on [Save Config] button on the "Configure Embeddings, Vector Store and Record Manager" screen, doesn't save things.

**To Reproduce**

1. Go to 'Document Store' screen
2. Create new Document Store
3. Create a basic loader (can be Text File Document Loader) with some random text
4. Click on "Process".
5. Go to 'Upsert All Chunks' screen
6. Configure Embeddings, Vector Store and Record Manager data
7. Click on 'Save Config'
8. Go back to previous screen
9. Go to 'Upsert All Chunks' screen again

**Expected behavior**
Al the settings would be there, but the config screen is empty.

**Setup**

-   Installation: docker
-   Flowise Version: 2.2.7-patch.1.
-   OS: Ubuntu 24.04
-   Browser: Brave (but the API call was also tested with Insomnia)

**Additional context**
- I'm sure the browser is firing the `/api/v1/document-store/vectorstore/save` is being fired.
- The config is only written when using the 'Upsert' button.
- After the first 'Upsert', you can change something and 'Save Config'.  The changes won't be there when returning to the screen.

One very weird thing that happened twice.....
After upserting the first Document Store, created another one.... went to the steps 1 to 9 above.... when return to 'Upsert All Chunks' I saw all the configs (embedding, vector store....) from my other Document Store! 😢  That makes no sense, but It happened twice.

PR #4277, opened 2025-04-09
fix : data persistence issue while saving config in documentStore

aditya162002:
> This PR fixes the issue #4243  of data persistence in document storage 

To Reproduce

Go to 'Document Store' screen
Create new Document Store
Create a basic loader (can be Text File Document Loader) with some random text
Click on "Process".
Go to 'Upsert All Chunks' screen
Configure Embeddings, Vector Store and Record Manager data
Click on 'Save Config'
Go back to previous screen
Go to 'Upsert All Chunks' screen again
Expected behavior
Al the settings would be there, but the config screen is empty.

Below is the image before fix
Screenshot 2025-04-09 at 11 53 40 PM

after the changes now we can see that the configuration data in persistant, please see it in attachment below
Screenshot 2025-04-09 at 11 59 04 PM

Root Cause:
1)Frontend sent `storeId` but backend expected` id`
2) The configuration objects weren't formatted correctly
3)JSON parsing failures weren't handled gracefully
4)The system couldn't handle both string and object configurations

Solution:
1)Fixed parameter names to ensure proper data transfer between frontend and backend
2)Restructured the configuration data format for consistent handling
3)Improved type handling to work with both string and object configuration formats

### DIFF
--- a/packages/server/src/Interface.DocumentStore.ts
+++ b/packages/server/src/Interface.DocumentStore.ts
@@ -249,31 +249,68 @@ export class DocumentStoreDTO {
         documentStoreDTO.totalChunks = 0
 
         if (entity.whereUsed) {
-            documentStoreDTO.whereUsed = JSON.parse(entity.whereUsed)
+            try {
+                documentStoreDTO.whereUsed = JSON.parse(entity.whereUsed)
+            } catch (e) {
+                documentStoreDTO.whereUsed = []
+            }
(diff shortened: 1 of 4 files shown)


HenryHengZJ:
its actually a one line fix, opened a bugfix here https://github.com/FlowiseAI/Flowise/pull/4299 thansk anyway!

PR #4285, opened 2025-04-11
Fix(FlowiseChatGoogleGenerativeAI): Prevent "parts must not be empty" API error in Seq Agents

toi500:
**Problem:**

The `ChatGoogleGenerativeAI` node can fail with the error `Error: Google Generative AI requires alternate messages between authors` or potentially `contents.parts must not be empty`. This typically occurs in sequential execution flows (e.g., loops, agents) where the message history passed to the node might contain two consecutive messages with the same role (often `user` -> `user`) or where the final message added by Flowise has empty content parts.

image

**Root Cause:**

The Google Generative AI API strictly enforces alternating `user` and `model` roles and requires message parts to have content. Flowise's execution model often appends a final `user` message (derived from the node's "Human Prompt" input field) to the existing history. If the last message *before* this append was also `user`, or if the "Human Prompt" was empty, the API requirements could be violated. Previous code might throw errors or fail the API call.

**Solution:**

This PR modifies the `convertBaseMessagesToContent` helper function within the `ChatGoogleGenerativeAI` node's code:

1.  **Added Post-Processing Merge:** After the initial message list is constructed, a post-processing step now iterates through it.
2.  **Merge Logic:** If two consecutive messages have the *same role* AND both contain only simple text parts, the text content of the second message is merged (with a newline separator) into the text part of the first message. The second message is *not* added separately to the final list sent to the API. This resolves consecutive role violations and implicitly removes messages that might only contain empty text parts added by Flowise.
3.  **Final Safeguard:** After merging, a final check ensures the list doesn't end with a `model` role. If it does, a minimal `user` message (`{ role: 'user', parts: [{ text: '...' }] }`) is appended solely to ensure API compliance.

**Why this Solution:**

This approach avoids halting the node's execution by handling consecutive identical roles through merging. This merging implicitly fixes the "empty parts" issue often caused by Flowise's placeholder and explicitly fixes the "alternate messages" violation. The final safeguard ensures API compliance regarding the final message role.

The bug in Seq Agents can be reproduce using this flow:
[gemini-seq-agent-bug Agents.json](https://github.com/user-attachments/files/19717393/gemini-seq-agent-bug.Agents.json)

--

PD: This fix also works with multi agents setups

image

### DIFF
--- a/packages/components/nodes/chatmodels/ChatGoogleGenerativeAI/FlowiseChatGoogleGenerativeAI.ts
+++ b/packages/components/nodes/chatmodels/ChatGoogleGenerativeAI/FlowiseChatGoogleGenerativeAI.ts
@@ -249,7 +249,8 @@ class LangchainChatGoogleGenerativeAI
         runManager?: CallbackManagerForLLMRun
     ): Promise<ChatResult> {
         let prompt = convertBaseMessagesToContent(messages, this._isMultimodalModel)
-        prompt = checkIfEmptyContentAndSameRole(prompt)
+        // Removed the potentially problematic checkIfEmptyContentAndSameRole call here
+        // The logic is now handled within the updated convertBaseMessagesToContent
 
         // Handle streaming
         if (this.streaming) {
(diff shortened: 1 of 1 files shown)


toi500:
Merged here https://github.com/FlowiseAI/Flowise/pull/4292

Issue #4276, opened 2025-04-09
[BUG] AWS ChatBedrock image upload error


nicklhw:
**Describe the bug**
The AWS ChatBedrock node throws an error for image upload. Text input works but image upload seems to be calling the wrong inference profile and always defaults to anthropic.claude-3-haiku-20240307-v1:0.

2025-04-09 14:47:01 [ERROR]: [server]: Error: Invocation of model ID anthropic.claude-3-haiku-20240307-v1:0 with on-demand throughput isn’t supported. Retry your request with the ID or ARN of an inference profile that contains this model.
ValidationException: Invocation of model ID anthropic.claude-3-haiku-20240307-v1:0 with on-demand throughput isn’t supported. Retry your request with the ID or ARN of an inference profile that contains this model.
    at de_ValidationExceptionRes (/Users/nick.wong/.nvm/versions/node/v20.17.0/lib/node_modules/flowise/node_modules/@langchain/aws/node_modules/@aws-sdk/client-bedrock-runtime/dist-cjs/index.js:1607:21)
    at de_CommandError (/Users/nick.wong/.nvm/versions/node/v20.17.0/lib/node_modules/flowise/node_modules/@langchain/aws/node_modules/@aws-sdk/client-bedrock-runtime/dist-cjs/index.js:1424:19)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async /Users/nick.wong/.nvm/versions/node/v20.17.0/lib/node_modules/flowise/node_modules/@smithy/middleware-serde/dist-cjs/index.js:35:20
    at async /Users/nick.wong/.nvm/versions/node/v20.17.0/lib/node_modules/flowise/node_modules/@smithy/core/dist-cjs/index.js:167:18
    at async /Users/nick.wong/.nvm/versions/node/v20.17.0/lib/node_modules/flowise/node_modules/@smithy/middleware-retry/dist-cjs/index.js:320:38
    at async /Users/nick.wong/.nvm/versions/node/v20.17.0/lib/node_modules/flowise/node_modules/@aws-sdk/middleware-logger/dist-cjs/index.js:33:22
    at async BedrockChat._streamResponseChunks (/Users/nick.wong/.nvm/versions/node/v20.17.0/lib/node_modules/flowise/node_modules/@langchain/aws/dist/chat_models.cjs:695:26)
[...]

PR #4309, opened 2025-04-16
Fix: Default vision model overriding custom model name for AWS Chat Bedrock.

nicklhw:
This PR fixes this [bug](https://github.com/FlowiseAI/Flowise/issues/4276).

Bedrock requires the use of an [inference profile](https://docs.aws.amazon.com/bedrock/latest/userguide/cross-region-inference.html?icmpid=docs_bedrock_help_panel_cross_region_inference) when making on-demand API requests to Claude sonnet. The inference profile is in the form `<region>.<model name and version>`, i.e. `us.anthropic.claude-3-7-sonnet-20250219-v1:0`. The inference profile ID needs to be set as the `Custom Model Name` attribute for the AWS ChatBedrock node in order to override the default model name. 

image

The existing code incorrectly sets the vision model to default to `anthropic.claude-3-haiku-20240307-v1:0` when an inference profile ID is set as the `Custom Model Name`. For example, setting custom model name `us.anthropic.claude-3-7-sonnet-20250219-v1:0` will cause the vision model to default to haiku because the custom model name does not start with `claude-3`. 

The proposed fix checks that the model name includes the string `claude-3`, which would allow the custom model name to be used as the vision model.


HenryHengZJ (approved):
thanks!

PR #4314, opened 2025-04-18
changes & fixes for various issues in chatflow config

vinodkiran:
1. Fix: Configuration Not saved properly - Rate Limit & Override Config will overwrite each other
2. Fix: Post Processing Incorrectly escaping newlines and other chars 
3. Fix: Follow on Prompts error while rendering in non-streaming mode
4. NEW: Ability to select filetypes for full file uploads

### DIFF
--- a/packages/ui/src/ui-component/extended/OverrideConfig.jsx
+++ b/packages/ui/src/ui-component/extended/OverrideConfig.jsx
@@ -123,10 +123,15 @@ const OverrideConfig = ({ dialogProps }) => {
 
         let overrideConfig = { status: overrideConfigStatus }
         if (overrideConfigStatus) {
+            const filteredNodeOverrides = {}
+            for (const key in nodeOverrides) {
+                filteredNodeOverrides[key] = nodeOverrides[key].filter((node) => node.enabled)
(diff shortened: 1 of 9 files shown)


packages/ui/src/ui-component/extended/OverrideConfig.jsx
```diff
@@ -116,25 +116,22 @@ const OverrideConfig = ({ dialogProps }) => {
...
-                filteredNodeOverrides[key] = nodeOverrides[key].filter((node) => node.enabled)
-            }
-
-            obj.overrideConfig = {
-                ...obj.overrideConfig,
-                nodes: filteredNodeOverrides,
-                variables: variableOverrides.filter((node) => node.enabled)
+            overrideConfig = {
+                ...overrideConfig,
+                nodes: nodeOverrides,
```
HenryHengZJ, line 128: we dont have to filter out the enabled ones?

vinodkiran: Fixed in the new commit

HenryHengZJ (approved):

PR #4327, opened 2025-04-21
Feature/added litellm component

patrickreinan:


HenryHengZJ:
we can already use Custom Chat OpenAI node for LiteLLM no?

patrickreinan:
Hi,

Yes, and I was using it. But, I think that a dedicated component is better due more flexibility and if something changes in LiteLLM is faster to fix.

HenryHengZJ (approved):
thanks!

PR #4332, opened 2025-04-23
fix(document-stores/vector-databases): upserting error invalid metadata keys

ahmedrowaihi:
many vector dbs ( mainly weaviate ) do not allow metadata keys that has special chars like ":"
so it fails upserting, and most files ( especially PDFs ) includes this type of keys

this utility fixed all our problems internally

### DIFF
--- a/packages/server/src/queue/PredictionQueue.ts
+++ b/packages/server/src/queue/PredictionQueue.ts
@@ -10,6 +10,7 @@ import { RedisOptions } from 'bullmq'
 import logger from '../utils/logger'
 import { generateAgentflowv2 as generateAgentflowv2_json } from 'flowise-components'
 import { databaseEntities } from '../utils'
+import { executeCustomNodeFunction } from '../utils/executeCustomNodeFunction'
 
 interface PredictionQueueOptions {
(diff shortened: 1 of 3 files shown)


HenryHengZJ:
so this replace non-alphanumeric characters with an underscore

for example:
```
{
    'user-name': 'John',
    'contact.info': {
        'email@domain.com': 'john@example.com',
        'phone-number': '123456789'
    },
    tags: ['tag1', {'nested-key!': 'value'}]
}
```

to
```
{
    'user_name': 'John',
    'contact_info': {
        'email_domain_com': 'john@example.com',
        'phone_number': '123456789'
    },
    tags: ['tag1', {'nested_key_': 'value'}]
}
```

this is quite dangerous as im not sure will this cause any existing users that have special characters in their metadata filter keys. For exampe chroma that supports `$` - https://docs.trychroma.com/docs/querying-collections/metadata-filtering

```
{
  "metadata_field": {
    "$in": ["value1", "value2", "value3"]
  }
}
```

HenryHengZJ:
I think we can have a generic function, but only used it for weaviate for now

PR #4359, opened 2025-04-29
fix(nodes): Correct path and case sensitivity for Cheerio Web Scraper icon

toi500:
Resolves issues preventing the icon for the Cheerio Web Scraper node from displaying correctly in deployed environments.

**Problem:** The icon was missing after deployment, although it worked locally.

- Updates icon file reference to match exact filename case.
- Change node directory path (`CheerioWebScraper` instead of `FreeWebScraper`).


HenryHengZJ:
Think we can rename the tool name, because it is now the same as Cheerio Web Scraper in Document Loader, causing this bug to happen: https://github.com/FlowiseAI/Flowise/issues/4363

Perhaps just `Web Scraper Tool` ?

toi500:
> Think we can rename the tool name, because it is now the same as Cheerio Web Scraper in Document Loader, causing this bug to happen: #4363
> 
> Perhaps just `Web Scraper Tool` ?

Like this?

image
image

HenryHengZJ:
👍 yep

HenryHengZJ (approved):

PR #4371, opened 2025-05-02
fix: update correct label for InputLabel component

anoopw3bdev:
Noticed minor UI issue while exploring Flowise.

Issue:
On the Marketplace page, the label provided to the OutlinedInput component is incorrect, which causes the "Framework" label to overlap with the input border. Additionally, for other Select fields like "Tag" and "Type", there is extra spacing between the label and the border. (Please see the attached screenshot for reference.)

**Before**:
Screenshot 2025-05-02 at 7 59 31 AM

**After**:
image


anoopw3bdev:
I directly created the PR since its a very minor issue, please do let me know if I should create an issue and link it with this PR.
Thank you!

HenryHengZJ (approved):
thanks @anoopw3bdev

PR #4381, opened 2025-05-05
Fix(PostgresRecordManager): Handle time correctly and ensure schema creation

### LINKED ISSUE #4339: [BUG] PostgresRecordManager Inserts NaN into updated_at on Azure PostgreSQL, Breaking Cleanup Modes
**Describe the bug**
When using the Postgres Record Manager node with an Azure PostgreSQL Flexible Server database, the updated_at column (double precision type) in the specified record manager table (e.g., upsertion_records) gets populated with NaN instead of a valid epoch timestamp number.
This prevents the Incremental and Full cleanup modes from working correctly because the timestamps cannot be compared to determine stale records. As a result, all vectors in the target vector store (e.g., Pinecone) are deleted immediately after the upsert operation finishes, effectively preventing persistent storage when cleanup is enabled.
Setting the Cleanup mode to None allows vectors to be added and persist, but the underlying NaN insertion into the updated_at column still occurs.

**To Reproduce**
1. Configure Flowise to use Azure PostgreSQL Flexible Server as the main database (ensure pgcrypto and uuid-ossp extensions are enabled and created in the database).
2. Create a Document Store upsert process (e.g., using a simple file loader and Pinecone as the Vector Store).
3. Add the Postgres Record Manager node to the Document Store configuration.
4. Configure the Postgres Record Manager node:
- Point it to the same Azure PostgreSQL instance used by Flowise (correct host, port, database name 'production', user, password).
- Specify a table name (e.g., upsertion_records).
- Ensure SSL is enabled for the connection (using the POSTGRES_RECORDMANAGER_SSL=true environment variable, as there is no SSL toggle in the node UI).
- Set the Cleanup mode to Incremental or Full.
- Set the Source Id Key (e.g., source).
5. Run the upsert operation with a document that has the specified sourceIdKey in its metadata.
6. Observe the process: vectors are added to the target vector store (Pinecone) but are immediately deleted.
7. (Optional) Connect to the PostgreSQL database via psql and run SELECT key, updated_at FROM upsertion_records;. Observe that the updated_at column contains NaN.

**Expected behavior**
The updated_at column in the record manager table (upsertion_records) should contain a valid numeric epoch timestamp (double precision). The Incremental or Full cleanup modes should correctly compare these timestamps against subsequent upsert runs and only delete vectors corresponding to stale or removed source documents, not all vectors on every run.

**Screenshots**
Here's the NaNs in my upsertion_records database
Image

**Setup**
Installation: Docker deployed on Azure App Service (1 Main App + 5 Worker Apps, initially tested queue mode, later confirmed error persists in main mode on main app with workers stopped)
Flowise Version: [2.2.7-patch.1]
OS: Linux (Azure App Service Standard Plan)
Browser: Chrome (irrelevant)
Database: Azure PostgreSQL Flexible Server
Vector Store: Pinecone (Irrelevant: the error occurs during the Record Manager's interaction with PostgreSQL, before/after Pinecone interaction)

**Additional context**
Extensive troubleshooting confirmed the following:
Database is correctly configured: Direct psql tests confirm:
pgcrypto and uuid-ossp extensions are installed and functions (gen_random_uuid, uuid_generate_v4) are available and executable by the Flowise DB user (db_flowisechatbot_p).
The record manager table (upsertion_records) exists with the correct schema (updated_at double precision, uuid UUID PRIMARY KEY DEFAULT gen_random_uuid()).
Direct SQL command SELECT EXTRACT(EPOCH FROM CURRENT_TIMESTAMP) returns a valid number.
Direct SQL INSERT and UPDATE statements setting updated_at = EXTRACT(EPOCH FROM CURRENT_TIMESTAMP) work correctly and store a valid number (NOT NaN).
The Flowise DB user has SELECT, INSERT, UPDATE, DELETE permissions on the upsertion_records table.
The DB user's search_path includes public.
Environment Variables are correct: Confirmed via Azure CLI appsettings list and printenv inside the container that POSTGRES_RECORDMANAGER_SSL=true and other DB connection variables are correctly set for the Flowise apps.
Likely Code Bug Identified: The issue appears to stem from the getTime() function within packages/server/src/nodes/recordmanagers/PostgresRecordManager/PostgresRecordManager.ts.
The code executes: const res = await queryRunner.manager.query('SELECT EXTRACT(EPOCH FROM CURRENT_TIMESTAMP)');
It then attempts to parse the result using: return Number.parseFloat(res[0].extract);
However, the standard PostgreSQL output column name for EXTRACT(...) is date_part, not extract.
Therefore, res[0].extract is likely undefined.
Number.parseFloat(undefined) correctly returns NaN in JavaScript.
This NaN value is then passed to the update() function and inserted into the updated_at column, causing the downstream cleanup logic to fail.
Suggested Fix: Modify the getTime() function in PostgresRecordManager.ts to correctly access the result column, for example:
Change return Number.parseFloat(res[0].extract); to return Number.parseFloat(res[0].date_part);
OR alias the column in the SQL query: query('SELECT EXTRACT(EPOCH FROM CURRENT_TIMESTAMP) AS current_epoch') and use res[0].current_epoch.
Workaround: Setting the Cleanup mode to None prevents data loss but does not fix the underlying NaN insertion.

nikitas-novatix:
**Problem:**

The `PostgresRecordManager` node experienced two main issues:

1.  **Incorrect Time Handling:** The `getTime()` method used the query `SELECT EXTRACT(EPOCH FROM CURRENT_TIMESTAMP)` without aliasing the result column. This caused `res[0].now` to be `undefined` in the subsequent code, leading to a`NaN` value in the database. This `NaN` value was then stored in the `updated_at` column, causing incorrect record tracking, cleanup behavior, and indexing issues.
2.  **Schema Creation Failure on Older PostgreSQL:** The `createSchema()` method uses `gen_random_uuid()` in its `CREATE TABLE` statement. On older PostgreSQL versions (prior to v13), this function is only available via the `pgcrypto` extension. The node did not ensure this extension was enabled, leading to failures when initializing the node against such database versions.

This issue is tracked in #4339 

**Solution:**

This PR addresses the identified problems with the following changes to `packages/components/nodes/recordmanager/PostgresRecordManager/PostgresRecordManager.ts`:

1.  **Correct Time Handling:** Added the `AS now` alias to the `EXTRACT(EPOCH FROM CURRENT_TIMESTAMP)` query within the `getTime()` method. This ensures the result column is correctly named, `res[0].now` is populated with the epoch value, and `Number.parseFloat()` returns a valid number instead of `NaN`.
2.  **Ensure `pgcrypto` Extension:** Added the `CREATE EXTENSION IF NOT EXISTS pgcrypto;` command within the `createSchema()` method, executed just before the `CREATE TABLE` statement. This ensures the `gen_random_uuid()` function is available, providing backward compatibility for PostgreSQL versions prior to v13 where the function is not built-in. The `IF NOT EXISTS` clause makes this command safe – it causes no errors or side effects on newer PostgreSQL versions (>=13, where the function is built-in) or if the extension is already enabled for other reasons.

**Testing:**

-   Successfully reproduced the `NaN` value being generated for `updated_at` when the `AS now` alias was missing. Verified that adding the alias correctly produces a numeric timestamp.
-   Confirmed that adding `CREATE EXTENSION IF NOT EXISTS pgcrypto;` allows the node to initialize and create its schema correctly. Tested against PostgreSQL v17.4 where the command executes harmlessly due to `IF NOT EXISTS`. The change ensures compatibility for users on older PostgreSQL versions requiring the extension.

**Closes:**

Fixes #4339
Potentially other ones too.

### DIFF
--- a/packages/components/nodes/recordmanager/PostgresRecordManager/PostgresRecordManager.ts
+++ b/packages/components/nodes/recordmanager/PostgresRecordManager/PostgresRecordManager.ts
@@ -227,6 +227,8 @@ class PostgresRecordManager implements RecordManagerInterface {
             const queryRunner = dataSource.createQueryRunner()
             const tableName = this.sanitizeTableName(this.tableName)
 
+            await queryRunner.query('CREATE EXTENSION IF NOT EXISTS pgcrypto;')
+
             await queryRunner.manager.query(`
   CREATE TABLE IF NOT EXISTS "${tableName}" (
     uuid UUID PRIMARY KEY DEFAULT gen_random_uuid(),
@@ -258,9 +260,9 @@ class PostgresRecordManager implements RecordManagerInterface {
         const dataSource = await this.getDataSource()
         try {
             const queryRunner = dataSource.createQueryRunner()
-            const res = await queryRunner.manager.query('SELECT EXTRACT(EPOCH FROM CURRENT_TIMESTAMP)')
+            const res = await queryRunner.manager.query('SELECT EXTRACT(EPOCH FROM CURRENT_TIMESTAMP) AS now')
             await queryRunner.release()
-            return Number.parseFloat(res[0].extract)
+            return Number.parseFloat(res[0].now)
         } catch (error) {
             console.error('Error getting time in PostgresRecordManager:')
             throw error


HenryHengZJ:
How come I didnt run into any error when using PostgresQL Record Manager?

image

image

image

nikitas-novatix:
> How come I didnt run into any error when using PostgresQL Record Manager?
> 
> image image image

@HenryHengZJ You need to use either "Incremental" or "Full" cleanup mode. Did you do that?

I re-tested locally and the NaN didn't appear this time with the current code (which uses res[0].extract). This is because my local Postgres setup happens to name the result of SELECT EXTRACT(EPOCH FROM CURRENT_TIMESTAMP) as extract.
However, this default naming isn't guaranteed across all Postgres versions or environments (like my Azure setup, where the bug persists). The unaliased column might be named date_part (this is the case on my Azure setup) or something else, causing res[0].extract to be undefined and leading to NaN.

The PR makes this robust by:
- Aliasing the query: ... AS now
- Using the alias in code: res[0].now

This explicit aliasing is safer and ensures consistent behavior everywhere. It won't change any other behavior or break anything. What do you think?

HenryHengZJ (approved):

HenryHengZJ:
tested and works fine! thank you

Issue #4322, opened 2025-04-20
[BUG] ChatGoogleGenerativeAI forces to add a custom and existing model name.


IvanDrag0:
**Describe the bug**
When creating a new assistant and selecting the ChatGoogleGenerativeAI as the model provider, even though I selected the model I want to use from the drop-down menu, the form forces me to add a custom model name since it's a required field.

**To Reproduce**

1. Go to 'Assistants'
2. Click on 'Custom Assistant'
3. Add a new custom assistant with any name.
4. Fill in the assistant information.
5. Select ChatGoogleGenerativeAI as the model provider.
6. Select a model from the "Model Name" drop down and leave the "Custom Model Name" field blank.
7. Click Save Assistant
8. Receive error message about filling out all mandatory fields.
9. Clear the model name from the "Model Name" drop-down menu.
10. Add a model name in the "Custom Model Name" field.
11. Click Save Assistant
12. Receive error message about filling out all mandatory fields.
13. Fill in both fields.
14. Click Save Assistant
15. Receive success message that the assistant has been saved successfully. 

**Setup**

-   Installation: npx flowise start
-   Flowise Version: 2.2.7-patch.1
-   OS: Windows 11
-   Browser: Edge

PR #4414, opened 2025-05-11
bugfix: set Custom Model Name field as optional

k-santos:
This PR sets the Custom Model Name field as optional when using ChatGoogleGenerativeAI in the Custom Assistant feature.

Before this PR:
image
After this PR:
image

Closes https://github.com/FlowiseAI/Flowise/issues/4322


HenryHengZJ:
thx! it was fixed here: https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/chatmodels/ChatGoogleGenerativeAI/ChatGoogleGenerativeAI.ts#L66

will be pushing to new release soon

PR #4446, opened 2025-05-16
bugfix: change field to be optional

k-santos:
Today, the customModelName field is not optional, despite the message indicating that it should be. This PR makes this field optional.

How it is currently before this PR:
image


HenryHengZJ:
try refresh, redrag the node, it should be upadted in latest version

PR #4457, opened 2025-05-19
fix: allow non-UUID sessionId by changing column type to varchar

eriksonssilva:
This change updates the sessionId column in the Execution entity from uuid to varchar, allowing strings (e.g. phone number, email, custom IDs) to be used as session identifiers without errors.

While this behavior already works when using a Chatflow, Agentflow relies on the execution table — which previously enforced a UUID type. As a result, sending a string would cause the following error:

QueryFailedError: invalid input syntax for type uuid: "some_string"

This change fixes that inconsistency without breaking any existing functionality.

### DIFF
--- a/packages/server/src/database/entities/Execution.ts
+++ b/packages/server/src/database/entities/Execution.ts
@@ -18,7 +18,7 @@ export class Execution implements IExecution {
     agentflowId: string
 
     @Index()
-    @Column({ type: 'uuid' })
+    @Column({ type: 'varchar' })
     sessionId: string
 
     @Column({ nullable: true, type: 'text' })


HenryHengZJ:
we also need to add additional migration script to alter the column type to varchar for all databases

chungyau97:
Hi everyone,

It should be `chat_message.id`, not `chat_message.chatId` or `chat_message.sessionId`.

Each execution corresponds to a single `chat_message.id`, not a full chat session.

`chat_message.chatId` can be associated with multiple `chat_message.id` records.

So if you're linking from the `execution` table using `chatId`, it won't make sense—because you’re pointing to a conversation, not a specific message. 

Need to modify the `sessionId` column to `chat_message.id` in the execution table to properly reference the correct message.

PR #4584, opened 2025-06-05
Fix stripe-related issues

0xi4o:


packages/server/src/enterprise/webhooks/stripe.ts
```diff
@@ -0,0 +1,239 @@
...
+        logger.warn(`No subscription ID found in invoice: ${invoice.id}`)
+        return
+    }
+
+    const subscriptionId = typeof invoice.subscription === 'string' ? invoice.subscription : invoice.subscription.id
+    const queryRunner = getRunningExpressApp().AppDataSource.createQueryRunner()
+    await queryRunner.connect()
+
+    try {
+        await queryRunner.startTransaction()
```
chungyau97: This should start at line:107 after `const now = new Date().toISOString()`

packages/server/src/enterprise/webhooks/stripe.ts
```diff
@@ -0,0 +1,239 @@
...
+            }
+        } catch (stripeError) {
+            logger.error('Failed to update Stripe subscription to free plan', {
+                subscriptionId,
+                error: stripeError instanceof Error ? stripeError.message : stripeError,
+                stack: stripeError instanceof Error ? stripeError.stack : undefined
+            })
+        }
+
+        await queryRunner.commitTransaction()
```
chungyau97: "handleInvoiceMarkedUncollectible" function should only have 1 `await queryRunner.commitTransaction()`, which is at line:231 before `} catch (stripeError) {`.

packages/server/src/enterprise/webhooks/stripe.ts
```diff
@@ -0,0 +1,239 @@
...
+
+async function getStripe(): Promise<Stripe> {
+    if (!stripe) {
+        const stripeManager = await StripeManager.getInstance()
+        stripe = stripeManager.getStripe()
+    }
+    return stripe
+}
+
+export async function handleStripeWebhook(req: Request, res: Response) {
```
chungyau97: This should be following [workspace.controller.ts#L181](https://github.com/FlowiseAI/Flowise/blob/main/packages/server/src/enterprise/controllers/workspace.controller.ts#L181) in handling queryRunner and try catch.

packages/server/src/index.ts
```diff
@@ -153,6 +154,9 @@ export class App {
     }
 
     async config() {
+        // Add Stripe webhook route BEFORE global JSON middleware to preserve raw body
+        this.app.post('/api/v1/webhooks/stripe', express.raw({ type: 'application/json' }), handleStripeWebhook)
```
chungyau97: What will happen if move this into [./packages/server/src/routes/index.ts](https://github.com/FlowiseAI/Flowise/blob/main/packages/server/src/routes/index.ts)?

packages/server/src/enterprise/webhooks/stripe.ts
```diff
@@ -0,0 +1,239 @@
...
+import Stripe from 'stripe'
+import { StripeManager } from '../../StripeManager'
+import logger from '../../utils/logger'
+import { getRunningExpressApp } from '../../utils/getRunningExpressApp'
+import { Organization } from '../database/entities/organization.entity'
+import { WorkspaceUser } from '../database/entities/workspace-user.entity'
+import { OrganizationUserService } from '../services/organization-user.service'
+import { UsageCacheManager } from '../../UsageCacheManager'
+
+let stripe: Stripe
```
chungyau97: `let stripe: Stripe`, "getStripe", and "handleInvoiceMarkedUncollectible" can move to enterprise's services folder.

chungyau97 (changes requested):

packages/server/src/enterprise/webhooks/stripe.ts
```diff
@@ -0,0 +1,239 @@
...
+                userCount: userIds.length
+            })}`
+        )
+
+        const now = new Date().toISOString()
+        const result = await queryRunner.manager
+            .createQueryBuilder()
+            .update(WorkspaceUser)
+            .set({ lastLogin: now })
+            .where('userId IN (:...userIds)', { userIds })
```
chungyau97: This code lead to updating all workspaceId based on array of userId given.

**Before:**
image

**After:**
image

**Expected Result:**
Each user should have their own "Default Organization" and  "Default Workspace" which are created during the registration process. These "lastLogin" column should be update.

packages/server/src/enterprise/webhooks/stripe.ts
0xi4o: Updated

packages/server/src/enterprise/webhooks/stripe.ts
0xi4o: > Updated

packages/server/src/enterprise/webhooks/stripe.ts
0xi4o: Updated

packages/server/src/enterprise/webhooks/stripe.ts
0xi4o: Updated

packages/server/src/enterprise/webhooks/stripe.ts
0xi4o: Updated

packages/server/src/index.ts
0xi4o: This one can't be changed since Stripe expects the request body to be raw but we have to use `json` middleware for our other endpoints.

PR #4587, opened 2025-06-05
feat: Add configurable system prompt to Condition Agent

toi500:
-> PR Update Here: https://github.com/FlowiseAI/Flowise/pull/4587#issuecomment-2954235630

image

chrome_x59A4FcSmu

### DIFF
--- a/packages/components/nodes/agentflow/ConditionAgent/ConditionAgent.ts
+++ b/packages/components/nodes/agentflow/ConditionAgent/ConditionAgent.ts
@@ -27,7 +27,7 @@ class ConditionAgent_Agentflow implements INode {
     constructor() {
         this.label = 'Condition Agent'
         this.name = 'conditionAgentAgentflow'
-        this.version = 1.2
+        this.version = 1.1
         this.type = 'ConditionAgent'
         this.category = 'Agent Flows'
         this.description = `Utilize an agent to split flows based on dynamic conditions`


toi500:
## NOTE

I have put a bit of thought into this default scenario thing, and the more I think about it, the less sense it makes.

There are a lot of use cases where you might not need a default value or fallback, when you need the LLM to select one or another, and in the event you need it, it should be the user who implements such fallback scenario for their particular use case. There is nothing wrong about adding such fallback when needed, but it shouldn't be imposed by a hardcoded prompt that, honestly, non-technical users will never read. Much less if the node itself doesn't have any dedicated defult output for it (thankfully, because with it, it would be much less flexible than it can be).

So, I humbly propose to get rid of it in the node system prompt to help those non-technical users who will never read the prompt, can get errors without any idea why this is happening, and just expect that "it works."

Here is the prompt with that little tweak, leaving the examples for {{question}} as it is probably the most used input, but now with this edit field, we can, as developers, customize it for our special use case and adapt it to any input we want.

```text
You are part of a multi-agent system designed to make agent coordination and execution easy. Your task is to analyze the given input and select one matching scenario from a provided set of scenarios."

Input: A string representing the user's query, message or data.
Scenarios: A list of predefined scenarios that relate to the input.
Instruction: Determine which of the provided scenarios is the best fit for the input.

Steps
1. Read the input string and the list of scenarios.
2. Analyze the content of the input to identify its main topic or intention.
3. Compare the input with each scenario: Evaluate how well the input's topic or intention aligns with each of the provided scenarios and select the one that is the best fit.
4. Output the result: Return the selected scenario in the specified JSON format.

Output Format
Output should be a JSON object that names the selected scenario, like this: {"output": "<selected_scenario_name>"}. No explanation is needed.

Examples
1.
Input: {"input": "Hello", "scenarios": ["user is asking about AI", "user is making a general greeting"], "instruction": "Your task is to check if the user is asking about AI or just saying hello."}
Output: {"output": "user is making a general greeting"}

2.
Input: {"input": "What is AIGC?", "scenarios": ["user is asking about AI", "user is asking about the weather"], "instruction": "Your task is to check and see if the user is asking a topic about AI."}
Output: {"output": "user is asking about AI"}

3.
Input: {"input": "Can you explain deep learning?", "scenarios": ["user is interested in AI topics", "user wants to order food"], "instruction": "Determine if the user is interested in learning about AI."}
Output: {"output": "user is interested in AI topics"}

Note
- Ensure that the input scenarios align well with potential user queries for accurate matching.
- DO NOT include anything other than the JSON in your response.
```

toi500:
## Few Considerations

I have put a bit of thought into this default scenario thing, and the more I think about it, the less sense it makes.

There are a lot of use cases where you might not need a default value or fallback, when you need the LLM to select one or another, and in the event you need it, it should be the user who implements such fallback scenario for their particular use case. There is nothing wrong about adding such fallback when needed, but it shouldn't be imposed by a hardcoded prompt that, honestly, non-technical users will never read. Much less if the node itself doesn't have any dedicated defult output for it (**thankfully, because with it, it would be much less flexible than it can be**).

So, I humbly propose to get rid of it in the node system prompt to help those non-technical users who will never read the prompt, can get errors without any idea why this is happening, and just expect that "it works."

**EDIT:** I have tested this prompt variation, A LOT, and it works just fine since, as I said, the default scenario is, per se, just another scenario.

**EDIT 2:** This prompt variation is also retro-compatible because, in the extremely rare event that a user has, after reading the default prompt, added a default scenario to prevent errors, the scenario was manually set up for the user, so the LLM will just treat that default scenario as any other one.

**EDIT 3:** Is this really hardcode stuff needed? I have removed it in my testing, and all works as expected since there are 3 explicit examples in the node system prompt. If we still need it, It would work better just with 1 role-based example.

image

toi500:
## PR Update

Small overhaul of the `ConditionAgent` node to ensure predictable and expected behavior, aligning its logic with its dynamic nature.

### Changes

*   **Added System Prompt Override:** A new optional "Node System Prompt" field has been introduced. This allows advanced users to tailor the agent's core instructions for specific use cases, such as referencing different input variables (e.g., `{{chat_history}}` instead of `{{question}}`) or optimizing for a particular model's behavior.

*   **Forced-Choice Logic:** The `default` fallback logic has been entirely removed. This is critical because the node’s output branches are dynamically generated from user-defined scenarios, meaning **there is no dedicated 'default' branch for a fallback to route to.** The system now enforces a strict forced-choice mechanism, requiring the AI to select the best-fitting scenario from the provided list.

*   **Synchronized Examples:** The hardcoded few-shot example in the node code is now perfectly aligned with the system prompt, ensuring consistent instruction for the AI (reduced from 3 to 1 to reduce the bias towards {{question}} as the required input, since all original examples used it).

*   **Hardened Error Handling:** The node will now **throw an explicit error** if the LLM fails to return a valid JSON structure. This prevents silent failures where the workflow would previously attempt to route to a non-existent path.

@HenryHengZJ @saatchi-david, I would like to read your thoughts on this. It could be a placebo effect, but the decisions made by the node are faster than before.

alt text

This issue was brought up for this user: https://discord.com/channels/1087698854775881778/1087698855342125118/1379834384118714428

packages/components/nodes/agentflow/prompt.ts
```diff
@@ -39,37 +39,45 @@ export const DEFAULT_HUMAN_INPUT_DESCRIPTION_HTML = `<p>Summarize the conversati
...
-   **Output**: "\`\`\`json\n{"output": "user is asking about AI"}\`\`\`"
-
-3. **Input**: {"input": "Can you explain deep learning?", "scenarios": ["user is interested in AI topics", "default"], "instruction": "Determine if the user is interested in learning about AI"}  
-   **Output**: "\`\`\`json\n{"output": "user is interested in AI topics"}\`\`\`"
-
-## Note
-- Ensure that the input scenarios align well with potential user queries for accurate matching
-- DO NOT include anything other than the JSON in your response.
+export const CONDITION_AGENT_SYSTEM_PROMPT = `
+<p>You are part of a multi-agent system designed to make agent coordination and execution easy. Your task is to analyze the given input and select one matching scenario from a provided set of scenarios.</p>
```
HenryHengZJ, line 43: can we change to markdown?

packages/components/nodes/agentflow/ConditionAgent/ConditionAgent.ts
```diff
@@ -59,6 +59,16 @@ class ConditionAgent_Agentflow implements INode {
                 acceptVariable: true,
                 default: '<p><span class="variable" data-type="mention" data-id="question" data-label="question">{{ question }}</span> </p>'
             },
+            {
```
HenryHengZJ: I'd prefer not to have this option as changing this alter not just the node behaviour, but also the flow logic in `buildAgentflow`

toi500: Removing the new Node System Prompt option will undermine the whole purpose of this PR.

As I explained in the PR description (apparently unsuccessfully), this node accepts all sorts of inputs (which is a good thing), not just {{questions}}, which are the only example types in the system prompt and role-based example. Advanced users could want to use any other type of check against, like chat_history, function outputs, or a tool output, and it is critical to ensure versatility and flexibility for ALL use cases.

Here is a real example, check again 3 factors at the same time:

- ### Instructions
Based on the provided data, determine the correct agent to route the user to. To qualify for the upsell agent, three conditions must be met: 1. The 'currentTime' must be before 5:00 PM. 2. The 'userTone' must be 'friendly'. 3. The 'hasOrderedBefore' value must be true. If any of these conditions are not met, route to the standard order agent.

- ### Inputs
 currentTime: {{tool_output_time}}
 userTone: {{question}}",
 hasOrderedBefore: {{chat_history}}

- ### Scenario Logic

   **Scenario 1: `Go to the upsell agent`**
    *   **Condition:** If the current time is before 5 PM **AND** the user's tone is "friendly" **AND** the user has ordered before.

   **Scenario 2: `Go to the standard order agent`**
    *   **Condition:** This is the path for all other cases (Example: it's after 5 PM, the tone is not friendly, or it's a new customer).

@HenryHengZJ this new feature is optional and a real necessity for advanced users and developers who need to instruct the LLM with non-{{question}} inputs.

How could adding or removing examples for the LLM alter the logic in buildAgentFlow? Isn't this the same as what we had with the system prompt of the Supervisor in v1? I mean, I personally had to change the prompt of the supervisor to make it work with older models or Gemini, as that prompt was very pro-OpenAI models. This feature is targeted at advanced users and developers.

saatchi-david:
Just chiming in as I was tagged. :)

First of all always appreciate the extra thinking and all the work @toi500!

Completely agree with removing the 'default' fallback and allowing the user to specify a scenario.

As for overriding the node's prompt, I'm torn. I generally would rather have the extra control, but question how much I would use it, and whether it's worth complicating the UI/UX. 

At the moment, the main use case I can imagine is optimizing it (i.e. removing/updating irrelevant examples) to keep token cost down.

@toi500 you mentioned this is necessary for cases that are "non-{{question}} inputs", but couldn't we just add the relevant/necessary inputs in the inputs field?

toi500:
@saatchi-david, thanks, man, I appreciate your kind words here.

Did you have the oportunbity to read this update? https://github.com/FlowiseAI/Flowise/pull/4587?notification_referrer_id=NT_kwDOCD7j27UxNjgyMTkwNjI0MjoxMzgzMzkyOTE#discussion_r2135021569

In relation to your question? You are correct, the input field is meant to provide the data the LLM needs to factor in, but it is not the same sort of data as a user question or a tool output. If we only have hardcoded, non-editable examples for the LLM, we have a problem here.

There isnt a perfect solution in terms of what examples to add. @HenryHengZJ added the question ones, as it is probably the most used, but without an editable prompt, this node becomes just a showcase, not ready for production, since you could not factor several inputs at the same time. The LLM probably will get confused since it will never see that complex scenario. Giving us the opportunity to add the examples we need for our use cases is very important and the main reason for this PR.

chrome_riOoFUBPyx

saatchi-david:
@toi500 
I'm seeing the value of what you're saying. I guess it's a question of how well the LLM would follow the user's instructions, assuming the instructions were clear in the first place.

To be honest, looking at the prompt, I feel like it's structured well and that the LLM would still handle complex instructions / multiple inputs.
But again, I'm also down for control/transparency.

Maybe there's a happy middle ground? Some quick thoughts.

1. Instead of giving control to override the full prompt, maybe giving control over updating only the examples? For example, below the "Scenarios" section, add another section titled "Reference Examples (optional)" with a "(+) Add Example" button. When clicked, the user can populate 2 fields: 1) the input, 2) select an output via dropdown populated with existing scenarios. This would allow the format of the example to be controlled and added dynamically to the prompt, replacing the default prompts.

2. Allowing for updating the full prompt, but making it an option that doesn't interrupt/complicate the current flow, which will be sufficient for most. n8n's option drop down comes to mind. It sits at the bottom of the node settings, and only says "Add Options". Once clicked the user can select an option, and configure from there. It's nice because it's out of the way, doesn't add any choices for the users, but easy enough to find for users looking for more.
image

Just my 2 cents! 

*Edit: corrected my first suggestion.

toi500:
@saatchi-david I am truly happy you engaged in this conversation, the more thought put into getting this done, the better.

When I tweaked the prompt, I tried to make it as neutral as possible with things like "A string representing the user's query, message, or data."

Anyway, I guess we could do your option 1: I mean, it's better than having hardcoded, non-editable examples imposed, although I don't see the truly advantage of this, having the supervisor system prompt editable has saved me more than once.

In relation to your point 2, unfortunately, Flowise doesnt have such a thing for node UIs. I tried to add a toggle option so it would show the edit field, but to do that, other files need to be modified, and its better for me not to mess with the node UI that has it own structure.

saatchi-david:
Happy to hear it! All just food for thought, but glad if productive.

1: True, but there's usually a tradeoff between control and user friendliness. Assuming the team is leaning a bit more towards ease-of-use with V2.

2: I see.. not to go too far down the rabbit hole, but maybe it's worth starting with that issue? Could imagine other nodes would benefit from such flexibility and that this kind of discussion around how much detail/control to give to the user will pop up again.

Will leave it to you and big boss Henry to ~~duke it out~~ negotiate on the best course of action 😁

toi500:
Yes, let's wait before proceeding.

Just let me add an open final rhetorical thought.

What is the alternative for me or anyone else who wants to implement complex scenarios to factor in several inputs and do a truly smart agentic application using Flowise? Do we rely on hardcoded examples for the LLM that are irrelevant for our own case? Do we need to only use very expensive chatModels that can get it right even if the examples make no sense for the data that they will find? Do we need to wish for the best trust in our own good luck?🤞

HenryHengZJ:
After testing several times I think we should be okay to proceed. To minimize the user chance of messing around with the prompt, I suggest adding a toggle button, which will only show the system prompt field when its enabled:

image

So:
```
{
      label: 'Scenarios',
      name: 'conditionAgentScenarios',
      description: 'Define the scenarios that will be used as the conditions to split the flow',
      type: 'array',
      array: [
          {
              label: 'Scenario',
              name: 'scenario',
              type: 'string',
              placeholder: 'User is asking for a pizza'
          }
      ],
      default: [
          {
              scenario: ''
          },
          {
              scenario: ''
          }
      ]
  },
  {
      label: 'Override System Prompt',
      name: 'conditionAgentOverrideSystemPrompt',
      type: 'boolean',
      description: 'Override initial system prompt for Condition Agent',
      optional: true
  },
  {
      label: 'Node System Prompt',
      name: 'conditionAgentSystemPrompt',
      type: 'string',
      rows: 4,
      optional: true,
      acceptVariable: true,
      default: CONDITION_AGENT_SYSTEM_PROMPT,
      description: 'Expert use only. Modifying this can significantly alter agent behavior. Leave default if unsure',
      show: {
          conditionAgentOverrideSystemPrompt: true
      }
  }
```

Even if user try to change the system prompt to something it isnt meant to do, we will throw the error which prevent the logic from further processing nodes. Also we can remove the `// *** CORRECTED LOGIC ***`

toi500:
@HenryHengZJ, thank you for testing this out. This makes total sense to me. Since you have already made this implementation, would it be okay for you to finish this out? You have write permission in the PR.

Edit: Also, it should be version 1.1 since we are bumping from 1.0.

PR #4589, opened 2025-06-05
Evaluations for Agentflows v2 & Assistants

vinodkiran:
This PR brings evaluations to the New Agentflows (v2) and Custom Assistants.

When creating your evaluation, you can now select a combination of Chatflow/Agentflow/Assistants to evaluate
image

After the evaluation completes, view and compare the results side by side.
image

The drill down features shows even more details (like the data/tokens/costs at a node level for the Agentflows)
image


packages/server/src/utils/buildChatflow.ts
```diff
@@ -724,7 +724,9 @@ export const executeFlow = async ({
             fileUploads: uploads ? JSON.stringify(fileUploads) : undefined,
             leadEmail: incomingInput.leadEmail
         }
-        await utilAddChatMessage(userMessage, appDataSource)
+        if (!isEvaluation) {
```
HenryHengZJ: I think previously we decided to save it to chat message because if not, the treating evaluation as same session wouldnt work. So thats why we have the `chatType` = `ChatType.EVALUATION`. So I think instead of not storing to chat message, we can use the same `chatType: isEvaluation ? ChatType.EVALUATION : isInternal ? ChatType.INTERNAL : ChatType.EXTERNAL` for buildAgentflow

packages/server/src/utils/buildChatflow.ts
```diff
@@ -816,7 +816,8 @@ export const executeFlow = async ({
         /*** Prepare response ***/
         result.question = incomingInput.question // return the question in the response, this is used when input text is empty but question is in audio format
         result.chatId = chatId
-        result.chatMessageId = chatMessage?.id
+        result.chatMessageId = chatMessage?.id || uuidv4()
+        result.chatflowid = chatflowid
```
HenryHengZJ: why do we need to return this?

packages/server/src/utils/buildAgentflow.ts
```diff
@@ -1868,7 +1868,7 @@ export const executeAgentFlow = async ({
     result.question = incomingInput.question // return the question in the response, this is used when input text is empty but question is in audio format
     result.form = form
     result.chatId = chatId
-    result.chatMessageId = chatMessage?.id
+    result.chatMessageId = chatMessage?.id || uuidv4()
```
HenryHengZJ: we shouldnt generate a random uuid as this is the response we returned

packages/server/src/utils/buildChatflow.ts
vinodkiran: Nope, not needed

Issue #4600, opened 2025-06-07
[BUG] Iteration Node is not parsing the Flow variable, it remains as a string.


jtguzman:
Describe the bug
The Iteration node is not properly parsing variables to JSON format. Instead of interpreting the variable as a JSON object/array, it treats the variable as a string, preventing proper iteration over the data structure.
To Reproduce

Go to Agentflow
Create a flow with an Iteration node
Pass a variable that should be parsed as JSON to the Iteration node
Configure the node to iterate over the JSON data
Execute the flow
Observe that the variable is treated as a string instead of being parsed as JSON
See error - iteration fails because it's trying to iterate over a string rather than a JSON structure

Expected behavior
The Iteration node should automatically parse variables containing JSON data into proper JSON objects/arrays, allowing for correct iteration over the data structure elements.

Screenshots

![Image](https://github.com/user-attachments/assets/fd36c2a5-3157-4536-9442-e0071c327a25)
![Image](https://github.com/user-attachments/assets/c7e470b7-053e-4b68-ae3b-5e26aa340948)

Setup
Installation: Docker Flowise@3.0.1

Additional context

The variable contains valid JSON data
The issue appears to be with the parsing mechanism in the Iteration node
This prevents proper iteration over JSON arrays/objects
The node should automatically detect and parse JSON formatted strings
Similar functionality works correctly in other nodes that handle JSON data

HenryHengZJ:
not sure is this fixed on latest version, but here's what I tried and its working:

![Image](https://github.com/user-attachments/assets/168c26e5-9863-4202-8fa1-6630810fc76b)

In custom function, I returned a simple array `[1, 2, 3]`:

![Image](https://github.com/user-attachments/assets/c54889c0-7320-4b1a-a5ab-2d723aba35be)

Iteration node picks this up:

![Image](https://github.com/user-attachments/assets/54eee463-e513-4c42-9281-aa2ba3942742)

And correctly processed 3 times:

![Image](https://github.com/user-attachments/assets/1f9e070a-aa0a-453b-9321-c7b9ca81a768)

PR #4603, opened 2025-06-07
fix: robust JSON parsing in Iteration node

hackice20:
This PR fixes a bug in the Iteration node where JSON input variables were not being parsed correctly. The node now robustly parses stringified JSON (including those with escape characters), allowing proper iteration over arrays and objects. This prevents errors when passing variables or JSON strings as input to the Iteration node.
How to test:
Pass a variable or stringified JSON array/object to the Iteration node’s input.
The node should correctly iterate over each element without JSON parsing errors.


HenryHengZJ:
can you give me a demo agentflow to replicate the error?

hackice20:
[BUG] Iteration Node is not parsing the Flow variable, it remains as a string. #4600

hackice20:
somebody aprove this sh*t

Issue #4563, opened 2025-06-03
[BUG] docker compose TypeError: JwtStrategy requires a secret or key


Mitty-ZH:
**Describe the bug**
An Error happened when running "docker compose up -d", and can't open the webpage at "https://localhost:3001"

**To Reproduce**
1. cd docker
2. cp .env.example .env and modify PORT=3001
3. docker compose up -d
4. docker compose logs -f

**Expected behavior**
Service started successfully.

**Screenshots**
flowise-1  | 2025-06-03 08:05:21 [INFO]: Starting Flowise...
flowise-1  | 2025-06-03 08:05:21 [INFO]: 📦 [server]: Data Source is initializing...
flowise-1  | 2025-06-03 08:05:24 [INFO]: 📦 [server]: Data Source has been initialized!
flowise-1  | 2025-06-03 08:05:24 [ERROR]: TypeError: JwtStrategy requires a secret or key

**Flow**
If applicable, add exported flow in order to help replicating the problem.

**Setup**

-   Installation: docker
-   Flowise Version: flowiseai/flowise:3.0.1
-   OS: Linux
-   Browser: chrome

**Additional context**

ilog2000:
Not really necessary to update the port. It fails on 3000 with the same error, too.

jasonmcmunn:
3.0.1 had a breaking change.. see the docs:
https://github.com/FlowiseAI/Flowise/releases/tag/flowise%403.0.1

TL;DR comment these lines out and update appropriately:

 JWT_AUTH_TOKEN_SECRET='AABBCCDDAABBCCDDAABBCCDDAABBCCDDAABBCCDD'
 JWT_REFRESH_TOKEN_SECRET='AABBCCDDAABBCCDDAABBCCDDAABBCCDDAABBCCDD'
 JWT_ISSUER='ISSUER'
 JWT_AUDIENCE='AUDIENCE'
 JWT_TOKEN_EXPIRY_IN_MINUTES=360
 JWT_REFRESH_TOKEN_EXPIRY_IN_MINUTES=43200

ilog2000:
> 3.0.1 had a breaking change.. see the docs: https://github.com/FlowiseAI/Flowise/releases/tag/flowise%403.0.1
> 
> TL;DR comment these lines out and update appropriately:
> 
[...]

PR #4606, opened 2025-06-07
fix(docker): uncomment required env variables

saishreyakumar:
Uncommented env variables to be in line with https://github.com/FlowiseAI/Flowise/releases/tag/flowise%403.0.1

Related to #4563


saishreyakumar:
@HenryHengZJ Please check if this is right. Thank You!

HenryHengZJ:
also here: https://github.com/FlowiseAI/Flowise/blob/main/packages/server/.env.example

saishreyakumar:
Thanks for the feedback! I have now uncommented the required JWT variables in both docker/.env.example and packages/server/.env.example.

HenryHengZJ (approved):
thanks!

Issue #4388, opened 2025-05-06
[FEATURE] Improve role assignment


jlistemann-ish:
**Describe the feature you'd like**
Changing the role of a user can be easier.

Currently, I need to:
1. Unassign the role from the user, confirm
2. Assign the new role to the user, confirm

I just want to select another role and save.

Just so you know, you should think that logoff/on is required by the affected user too.
Maybe the users' session could be updated behind the scenes, too.

HenryHengZJ:
yep we have that in-work!

PR #4616, opened 2025-06-09
New Feature: Ability to change role for a workspace user.

vinodkiran:
**Feature: Change role of a workspace user**
*Edit on active users would allow changing of role, without having to unlink and reassign.*
image
*Change Role Dialog*
image

Fix for #4388

***Note: the user has to logout and login again for the new role to take effect. The ability to refresh in the same session will come as a separate PR.***

### DIFF
--- a/packages/server/src/enterprise/services/workspace-user.service.ts
+++ b/packages/server/src/enterprise/services/workspace-user.service.ts
@@ -327,30 +327,31 @@ export class WorkspaceUserService {
         return newWorkspace
     }
 
-    public async updateWorkspaceUser(newWorkspaserUser: Partial<WorkspaceUser>) {
+    public async updateWorkspaceUser(newUserDetails: Partial<WorkspaceUser>) {
         const queryRunner = this.dataSource.createQueryRunner()
         await queryRunner.connect()
 
         const { workspaceUser } = await this.readWorkspaceUserByWorkspaceIdUserId(
-            newWorkspaserUser.workspaceId,
-            newWorkspaserUser.userId,
+            newUserDetails.workspaceId,
+            newUserDetails.userId,
             queryRunner
         )
         if (!workspaceUser) throw new InternalFlowiseError(StatusCodes.NOT_FOUND, WorkspaceUserErrorMessage.WORKSPACE_USER_NOT_FOUND)
-        if (newWorkspaserUser.roleId) {
-            const role = await this.roleService.readRoleById(newWorkspaserUser.roleId, queryRunner)
+        if (newUserDetails.roleId) {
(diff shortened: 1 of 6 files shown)


HenryHengZJ:
@chungyau97 need your help in testing this

packages/server/src/enterprise/services/workspace-user.service.ts
chungyau97, line 1: This file can be revert back to what it used to since there are nothing change beside renaming.
Renaming it to `newUserDetails` might be confusing as it is an object of `WorkspaceUser`.

The only justifiable changes required for this file are:
1. `queryRunner` move into `workspace-user.controller.ts` and better `queryRunner` handling.
Code Example:
```typescript
let queryRunner: QueryRunner | undefined
try {
    queryRunner = getRunningExpressApp().AppDataSource.createQueryRunner()
    await queryRunner.connect()
    
} catch (error) {
    if (queryRunner && queryRunner.isTransactionActive) await queryRunner.rollbackTransaction()
    next(error)
} finally {
    if (queryRunner && !queryRunner.isReleased) await queryRunner.release()
}
```
2. Removing `startTransaction()`, and `commitTransaction()`, due to it's just updating 1 row of data instead of `n` row of data,
3. Add verification for `roleId` pass in should reference the correct organization's of the worksapceUser tie to, as the role is tie to organization. Trangresion of not doing this is people can just pass in other roleId of other organziation via API.

chungyau97 (changes requested):

packages/server/src/enterprise/services/workspace-user.service.ts
vinodkiran: I have reverted the changes and added the check for the organization. I am not addressing the other changes (transaction in service and controller), hopefully we will take this up at a later phase.

chungyau97:
## Test Step
**UI:**
1. Register 2 user.
2. Create 2 roles in each user's own organization.
3. During role modification, it will only show current organization's roles to choose from.
4. Modification of role will be successful.

**API:**
1. Update workspaceUser' role with `roleId` belongs to that organzation, will return status code 200 and workpsaceUser object.
5. Update workpsaceUser' role with `roleId` does not belong to organization, will return status code 404 and error message "Role Not Found".

## Result
**UI:**
|yau's organization before|yau's organization after|
| -------- | ------- |
|image|image|

|hello's organization before|hello's organization after|
| -------- | ------- |
|image|image|

**API:**
|roleId belong to organization|roleId does not belong to organization|
| -------- | ------- |
|image|image|

HenryHengZJ (approved):

PR #4620, opened 2025-06-10
feature/single-organization-limit

chungyau97:


HenryHengZJ:
Malicious user cannot register again if the organization has already existed:
image

PR #4626, opened 2025-06-10
Fix: Correct formatting of Condition Agent prompt

toi500:
The HTML for the `CONDITION_AGENT_SYSTEM_PROMPT` constant contained extra whitespace, which caused improper rendering in the UI (Expandable debugging view only). This PR adjusts the HTML formatting to ensure it displays correctly as intended.

**Before:**
capture_250610_201354

**After:**
capture_250610_201619

### DIFF
--- a/packages/components/nodes/agentflow/prompt.ts
+++ b/packages/components/nodes/agentflow/prompt.ts
@@ -39,15 +39,12 @@ export const DEFAULT_HUMAN_INPUT_DESCRIPTION_HTML = `<p>Summarize the conversati
 </ul>
 `
 
-export const CONDITION_AGENT_SYSTEM_PROMPT = `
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
This is the Before for me, looks like the same as After
image

toi500:
I have checked in a production environment and a local deployment, both updated to the last commit, and I definitely see the UI bug. I just followed the prompt format above for the fix.

image

Before, in my testing deployment (for the original PR), it was working just fine, but it was 5 days old, so I dont know, maybe it is now conflicting with a recent commit. I cant say.

Edit: maybe this one? https://github.com/FlowiseAI/Flowise/commit/2cd8db0c534e4ffeaa6ab56bbfd0e1d847ce56c9

HenryHengZJ:
yeah not sure too, thanks for the fix!

PR #4642, opened 2025-06-12
fix: Modify the style of the AsyncDropdown component

ejrtks1020:
### Changes
- Adjusted the `Autocomplete` component's width to `90%` when `multiple` is true to prevent clipping of the refresh button.
- Added a tooltip on hover for selected action items to display the **full action name**, especially when it overflows.

### Reason
- Long action names were causing UI overflow, making the refresh button partially hidden and selected text unreadable.
- Improved UX by making the layout more responsive and enhancing the visibility of selected items.

### Notes
- This change affects only the `Autocomplete` UI within the affected node and has no side effects on other components.

image


HenryHengZJ (approved):
thank you!

PR #4654, opened 2025-06-13
doc: Add links to translated README versions

dowithless:
Allowing users around the world to read the documentation in their native languages — including German, Spanish, French, Portuguese, and Russian. 

This helps improve accessibility and understanding of the project for a global audience.

The updated links can be previewed in my forked repository: https://github.com/dowithless/Flowise/tree/patch-1

> Keep these links. Translations will automatically update with the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,12 @@
 [![GitHub star chart](https://img.shields.io/github/stars/FlowiseAI/Flowise?style=social)](https://star-history.com/#FlowiseAI/Flowise)
 [![GitHub fork](https://img.shields.io/github/forks/FlowiseAI/Flowise?style=social)](https://github.com/FlowiseAI/Flowise/fork)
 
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
thanks! do you mind add it to https://github.com/FlowiseAI/Flowise/tree/main/i18n

dowithless:
> thanks! do you mind add it to https://github.com/FlowiseAI/Flowise/tree/main/i18n

Got it — do you mean putting the translated README files directly into the project and updating the links to use internal paths?

That's definitely doable! The readme-i18n tool supports downloading any translated README, so we can place them directly in the repo.

The only thing to keep in mind is that with this approach, whenever the original README is updated, the translations would need to be manually re-synced.

In contrast, using external links allows the translations to stay automatically in sync with the latest README updates — no extra effort needed.

That said, if you'd prefer to go with internal links, I'm happy to help get that set up!

PR #4734, opened 2025-06-25
Bugfix/Allow OverrideConfig For Multiple Nodes In AgentflowV2

HenryHengZJ:
Improve the handling of input configurations by merging existing values with overrides instead of complete replacement. This includes support for merging objects and parsing JSON strings when necessary.

### DIFF
--- a/packages/server/src/utils/index.ts
+++ b/packages/server/src/utils/index.ts
@@ -1149,7 +1149,33 @@ export const replaceInputsWithConfig = (
                 if (nodeIds.includes(flowNodeData.id)) {
                     // Check if this parameter is enabled
                     if (isParameterEnabled(flowNodeData.label, config)) {
-                        inputsObj[config] = overrideConfig[config][flowNodeData.id]
+                        const existingValue = inputsObj[config]
+                        const overrideValue = overrideConfig[config][flowNodeData.id]
+
+                        // Merge objects instead of completely overriding
+                        if (
+                            typeof existingValue === 'object' &&
+                            typeof overrideValue === 'object' &&
+                            !Array.isArray(existingValue) &&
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
@toi500 can you try this PR?

toi500:
> @toi500 can you try this PR?

Yes, let me test this out.

toi500:
1. The error is definitely gone, but I don't think it's actually overriding anything.

image

2. About `"credential": "string"`, how does this need to be filled?

HenryHengZJ:
My testing:

```
{
    "question": "hey",
    "overrideConfig": {
        "agentModelConfig": {
            "agentAgentflow_0": {
                "temperature": "0.2"
            },
            "agentAgentflow_1": {
                "openAIApiKey": "abc"
            }
        }
    }
}
```

image

Agent 0 has correct overriden temperature of 0.2:
image

And Agent 1 has the invalid Open API Key:
image

These shows the overriden values are correctly reflected

toi500:
I have redeploy again, locally, and I confirm this now working as expected. 

image

## Feedback

To get the correct `openAIApiKey` name, I had to explicitly create a chatflow a add chatOpenAI Node (same apply for Gemini, etc), so maybe it would be good to let the user know what `providerApikey` is used in V2 (or document how `credential` work, if I am not missing something here). 

image

toi500:
@HenryHengZJ Also, Agent Configuration is fetching the same JSON (openAI by default) even when nodes use different chat models. This will surely confuse users who will get the keys from there, as happened to me yesterday.

Nothing important, but maybe you want to be aware of this in case users state that overrides are not working properly, as they could be using the wrong data from there.

chrome_kZsAsJfJZU

Issue #4749, opened 2025-06-28
Loading local mcp server failed silently


rickliujh:
### Describe the bug

When some of tools have incorrect schema in MCP server, Flowise failed to load that MCP server with zero error indication. This cause two problems:

1. remaining tools that available in that MCP server is not accessible because of this 
2. making it hard to debug for MCP users

### To Reproduce

1. install [Smarsheet MCP](https://github.com/smartsheet-platform/smar-mcp) in the local following the instruction in the readme 
2.  open chatflow
3. drag a Custom MCP
4. use the config below
```json
{
   "command": "/usr/local/bin/node",
   "args": [
      "-r",
      "/root/smar-mcp/node_modules/dotenv/config",
      "/root/smar-mcp/build/index.js"
   ],
   "env": {
      "SMARTSHEET_API_KEY":"123",
      "SMARTSHEET_ENDPOINT": "https://api.smartsheet.eu/2.0"
   }
}
```
5. click the "refresh" button in the right side of available actions dropdown list
6. check the server logs

### Expected behavior

all other tools provided by SmartSheet MCP should be loaded besides two user related tools that have the schema problem

### Screenshots

![Image](https://github.com/user-attachments/assets/fc0865f6-2a19-422a-995b-35f78bfe8586)

![Image](https://github.com/user-attachments/assets/ea87809f-0e75-4c9f-9edd-d745641e66ac)

### Use Method

Docker

### Flowise Version

3.0.1

### Operating System

Linux

### Browser

Other

### Additional context

The error of user related tools in SmartSheet MCP failed the loading 
```
Error: Invalid schema type or missing properties
    at createSchemaModel (/usr/src/packages/components/dist/nodes/tools/MCP/core.js:139:15)
    at /usr/src/packages/components/dist/nodes/tools/MCP/core.js:86:29
    at Array.map (<anonymous>)
    at MCPToolkit.get_tools (/usr/src/packages/components/dist/nodes/tools/MCP/core.js:77:49)
    at MCPToolkit.initialize (/usr/src/packages/components/dist/nodes/tools/MCP/core.js:66:37)
```

PR #4750, opened 2025-06-28
fix(components): promise.all failed silently when loading MCP tools

rickliujh:
When some of tools in custom MCP server have incorrect schema definition, Flowise failed to load that MCP server with zero error indication.

Linked to Issue #4749 

The root cause is when there is any rejected promise the `Promise.all()` will abort all remaining result and fall silently.

This PR replace `Promise.all()` with `Promise.allSettled()`, this change provides two benefits:

1. even some tools failed to resolve, remaining tools that available in that MCP server is still accessible
2. print errors making it easier to debug for Custom MCP users


HenryHengZJ:
`pnpm lint-fix` to get rid of the linting issues

PR #4777, opened 2025-07-02
adding performance config option to the AWS ChatBedrock component

cerashdan:
Some models in Amazon Bedrock are latency optimized according to [Optimize model inference for latency](https://docs.aws.amazon.com/bedrock/latest/userguide/latency-optimized-inference.html) and @Langchain/aws has introduced the performance config parameter in version 0.1.11. 

This PR updates the AWS Chatbedrock component to include the Latency optimized option.


HenryHengZJ (approved):
thank you!

PR #4804, opened 2025-07-06
feat: adding github action to run all tests

nqodube:


HenryHengZJ:
is this meant to be your forked changes?

nqodube:
Yes they are, tested the action in our forked repo.

nqodube:
@HenryHengZJ are you asking because you may not want to run due to Github action minutes?

They are only triggered on demand and not as part of every PR.

Also, It may be beneficial for anyone that forks to also run these on their forked version too.

PR #4808, opened 2025-07-07
Fix issue 3570 and increasing db connections for pg vector and pg record manager

lingsongzeng:
## Description:
1. uncomment and use pgvector so that key column from 'upsertion_records' table will match the id column from 'documents' table.
2. However, above will result in increasing db connections. Release connections in proper places.


packages/components/nodes/vectorstores/Postgres/Postgres.ts
```diff
@@ -247,20 +203,28 @@ class Postgres_VectorStores implements INode {
 
             try {
                 if (recordManager) {
-                    const vectorStore = await vectorStoreDriver.instanciate()
+                    const vectorStore: any = await vectorStoreDriver.instanciate()
```
iEgit, line 206: this call creates new connection every time  


maybe we could cache it by tableName?

PR #4839, opened 2025-07-10
Added region support to chatGoogleVertexAi

Stono:
This PR adds support for `region` in `chatGoogleVertexAi`.

It also adds the known regions to `models.json`.

This has been added as an optional parameter so as to be backwards compatible.


packages/components/nodes/chatmodels/ChatGoogleVertexAI/ChatGoogleVertexAI.ts
```diff
@@ -223,6 +235,7 @@ class GoogleVertexAI_ChatModels implements INode {
         if (cache) obj.cache = cache
         if (topK) obj.topK = parseFloat(topK)
         if (thinkingBudget) obj.thinkingBudget = parseInt(thinkingBudget, 10)
+        if (region) obj.location = region
```
HenryHengZJ, line 238: what happens if you dont have the location? is it a must now? or what benefits does it add for specifying the region additionally, i dont' use vertex so kinda curious

Stono: It defaults to `us-central` which is buried away in the google libraries.  
Being able to decide where your data is processed is critical for any commercial use.

Stono:
Two more commits to add region support to the embeddings too, as well as removing models that no longer exist, and adding new ones

packages/components/models.json
```diff
@@ -673,6 +673,50 @@
                     "input_cost": 2.5e-7,
                     "output_cost": 1.25e-6
                 }
+            ],
+            "regions": [
```
HenryHengZJ, line 677: can we refractor the region into an object that can be re-used?

Stono: Yes, no problem.  Do you have any preferences as to where in the code base?

HenryHengZJ: hmm i think we are good for now, because otherwise we can't make it into one single file

HenryHengZJ (approved):
thank you!!

PR #4681, opened 2025-06-18
Add PostgreSQL schema configuration support via DATABASE_SCHEMA

stuart-pebble:
This update enables users to specify a custom PostgreSQL schema for Flowise to use, improving database organization and multi-tenant deployments.

Changes:
- Add schema property to PostgreSQL DataSource configuration using DATABASE_SCHEMA env var
- Configure search_path to include custom schema, public, and extensions schemas
- Ensure UUID functions remain accessible from extensions schema
- Maintain backward compatibility with default 'public' schema when DATABASE_SCHEMA is not set

PR #4840, opened 2025-07-10
Added support for postgres schema

Stono:
This PR adds support for the postgres schema to be set on the postgres data source


HenryHengZJ:
duplicated: https://github.com/FlowiseAI/Flowise/pull/4681

Stono:
Interestingly we've got different implementations though!

Issue #4841, opened 2025-07-10
Chat freezes after submitting the message


artoniaz:
### Describe the bug

I simply add a message to the chat. I get not response. It simplyt freezes. However when I close the chat window and reopen it it shows messages as if notheng ever happened. In LangSmith I see the LLM getting messages and returning output. It seems there is something wrong with frontend. 

I get this errors in the console:
anifest: Line: 1, column: 1, Syntax error.Understand this error
8VM304 vendor.js:151 Uncaught (in promise) Error: A listener indicated an asynchronous response by returning true, but the message channel closed before a response was received

### To Reproduce

This is test message that USUALLY doesn't work. Sometimes everything works fine:
martyna, Ze strefy na śląską płyta k003 - 15szt, blat 990#28 mat - 2szt, okleina s5501 22x0,8 - 300mb

### Expected behavior

The UI shoud provide a response from LLM

### Flow

[test agent Agents (1).json](https://github.com/user-attachments/files/21170930/test.agent.Agents.1.json)

### Use Method

None

### Flowise Version

cloud, agentflows v2

### Operating System

None

### Browser

None

nehaaprasad:
### Fix
- Fixed EventSource error handling that was crashing the chat
- Added safe error property access to prevent undefined errors
- Protected JSON parsing from malformed responses

- Chat now works reliably without freezing.

HenryHengZJ:
is this from cloud or self hosted service?

naymintun800:
> is this from cloud or self hosted service?

Self hosted one! I'm also getting the same problem. Requests to "flowise:3000" is failing. Especially this end point "/api/v1/internal-prediction/" .. I think.. Is there any workaround or fix for now? Should I use different version?

nehaaprasad:
> is this from cloud or self hosted service?

self hosted

stefanofalasca:
Same problem for me self hosted

artoniaz:
I still experience the issue. I happens on both local env and cloud, more often on cloud from my experience.

pit-crew:
I too am experiencing this issue suddenly.  Local env and cloud as well.  This is critical.  We've had to remove the chatbot from our web app!

warlocc:
This is still a problem- the entire server freezes up randomly and needs to be force restarted.

PR #4843, opened 2025-07-11
fix: Fix chat freezing when messages contain special characters or network errors occur

naaa760:
### 🐛 Problem
Users reported that the chat interface would freeze after submitting messages, especially with messages containing special characters (like Polish text). The UI would become unresponsive even though the backend was processing correctly.

#4841

###  Solution
Fixed EventSource error handling that was causing the chat to freeze:

- Removed error throwing** in EventSource `onerror` handler that was breaking the UI
- Added safe error property access** to prevent crashes when error objects don't have expected structure  
- Enhanced connection validation** to handle failed EventSource connections gracefully
- Protected JSON parsing** to prevent single malformed messages from breaking the entire chat
- Added connection-level error handling** with user-friendly fallback messages

### ✅ Testing
- Chat no longer freezes with special character messages
- Error scenarios are handled gracefully with proper user feedback
- Build passes without errors
- Existing functionality preserved

###  Impact
- Better user experience with stable chat interface
- Improved error handling and debugging
- More resilient to network issues and server errors

### DIFF
--- a/packages/ui/src/views/chatmessage/ChatMessage.jsx
+++ b/packages/ui/src/views/chatmessage/ChatMessage.jsx
@@ -890,7 +890,7 @@ const ChatMessage = ({ open, chatflowid, isAgentCanvas, isDialog, previews, setP
         try {
             uploads = await handleFileUploads(uploads)
         } catch (error) {
-            handleError('Unable to upload documents')
+            handleError(error?.response?.data?.message || error?.message || 'An unexpected error occurred')
             return
         }
 
@@ -955,7 +955,7 @@ const ChatMessage = ({ open, chatflowid, isAgentCanvas, isDialog, previews, setP
                 }
             }
         } catch (error) {
-            handleError(error.response.data.message)
+            handleError(error?.response?.data?.message || error?.message || 'An unexpected error occurred')
             return
         }
     }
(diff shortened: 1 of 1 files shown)


packages/ui/src/views/chatmessage/ChatMessage.jsx
```diff
@@ -964,83 +964,107 @@ const ChatMessage = ({ open, chatflowid, isAgentCanvas, isDialog, previews, setP
...
+                        console.error('Error processing EventSource message:', error, 'Raw data:', ev.data)
+                        // Don't close the connection for a single malformed message
+                        // but log it for debugging
+                    }
+                },
+                async onclose() {
+                    closeResponse()
+                },
+                async onerror(err) {
+                    console.error('EventSource Error: ', err)
```
HenryHengZJ, line 1058: we have to throw error here otherwise `fetchEventSource` will keep retrying non stop causing unlimited calls to the LLM.

previously this had caught us surprised with LLM bills up to thousands due to not throwing the error

HenryHengZJ: thanks for the help! however I dont think that's the cause to the issue

PR #4846, opened 2025-07-11
Added support for cloudflare Workers AI via the OpenAI endpoint

Stono:
Hi, 
This PR adds support for Cloudflare Workers AI
It's quite a nice PaaS model environment.
https://developers.cloudflare.com/workers-ai/

I did originally add `@langchain/cloudflare`, however that seemed incomplete, didn't support tools etc.  Cloudflare provide openai compatible endpoints for their chat models so I'm instead using that.

The model data is generated by hitting `GET https://api.cloudflare.com/client/v4/accounts/{ACCOUNT_ID}/ai/models/search`; some of them are 0 cost because "beta" models don't have a cost.


Stono:
Hmm. I suppose one thing I don't like actually is that function calling does not work on all the models, it's only supported by 4 of them:

https://developers.cloudflare.com/workers-ai/models/?capabilities=Function+calling

Attempting it on another model and you get a 400.

Event with the "supported" models, it doesn't appear to work:

```
curl https://api.cloudflare.com/client/v4/accounts/account-id/ai/v1/chat/completions \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer token" \
  -d '{
    "model": "@hf/nousresearch/hermes-2-pro-mistral-7b",
    "messages": [
      {
        "role": "user",
        "content": "What is the weather like in Boston today?"
      }
    ],
    "tools": [
      {
        "type": "function",
        "function": {
          "name": "get_current_weather",
          "description": "Get the current weather for a given location",
          "parameters": {
            "type": "object",
            "properties": {
              "location": {
                "type": "string",
                "description": "The city and state, e.g., San Francisco, CA"
              },
              "unit": {
                "type": "string",
                "enum": ["celsius", "fahrenheit"]
              }
            },
            "required": ["location"]
          }
        }
      }
    ],
    "tool_choice": "auto"
  }'
```

Gives: 
```
{"id":"id-1752242603477","object":"chat.completion","created":1752242603,"model":"@hf/nousresearch/hermes-2-pro-mistral-7b","choices":[{"index":0,"message":{"role":"assistant","content":null},"logprobs":null,"finish_reason":"stop"}],"usage":{"prompt_tokens":0,"completion_tokens":0,"total_tokens":0}}
```

eg; null content, no tool calls, just no content.  Same with `@cf/meta/llama-4-scout-17b-16e-instruct`.

I'll poke this over to some people I know at Cloudflare to see what they say.

thatsKevinJain:
Hello @Stono, thank you raising this, we've released an update in the OpenAI endpoint that should fix tool calling for Workers AI models. Can you try out on your end and let us know if everything looks good?

Stono:
Hey, thanks so much for sorting!  I'm away until the middle of next week so will test it when I get back

Stono:
@thatsKevinJain works a treat with streaming disabled:

Screenshot 2025-07-23 at 18 04 55

However with streaming enabled we get back:

Screenshot 2025-07-23 at 18 06 35

eg a content response rather than a tool call.

This can be replicated with the curl setting `stream: true`

thatsKevinJain:
Yes, hermes-2-pro is an older model in our catalog that has a different output style than some of the newer models that adhere more to the openai format, thanks for the raising this, we can look into it, meanwhile can you try with @cf/meta/llama-4-scout-17b-16e-instruct with streaming?

PR #4872, opened 2025-07-15
Added yaml as a file upload option

Stono:
This simple PR adds the ability to attach yaml to chats


HenryHengZJ:
not sure if we have the appropriate loader to process YAML file, does it work?

Stono:
It did!

*Karl Stoney*
https://karlstoney.com

PR #4876, opened 2025-07-15
Fix mcp json to zod

Stono:
Hi, 
Currently MCP services are loaded and their parameters JSON is parsed really crudely.
This manifests at runtime as failing invocations of the tool as parameters are not being generated correctly, or even hallucination of additional parameters (due to `additionalProperties` being silently ignored).

This PR adds a library (`zod-from-json-schema`) which takes care of this far more accurately.


Stono:
I'm going to give up raising PRs!  I'm currently trying to maintain 10 different branches so that i'm contributing back to your project, but it's just getting overwhelming.

I'm happy to join as a contributor if you like, please reach out, but managing package locks across N feature branches is just a nightmare.

HenryHengZJ:
Don't worry about package lock, we will handle that

PR #4880, opened 2025-07-16
fix: normalize Milvus similarity scores for threshold filtering

ejrtks1020:
### Summary
Added score normalization logic to Milvus vector store to ensure SimilarityThresholdRetriever works correctly across different metric types.

### Changes
- Added normalization logic for L2, IP, and COSINE metric types
- L2 distances converted to similarity scores using `1/(1+distance)`
- IP and COSINE scores normalized from `[-1,1]` to `[0,1]` range

### Problem Solved
- SimilarityThresholdRetriever was not working properly with Milvus due to unnormalized scores
- Different metric types returned scores in different ranges, causing inconsistent threshold filtering
- Raw distance/similarity values were passed directly without proper normalization

### Testing
- Tested with L2, IP, and COSINE metric types
- Verified SimilarityThresholdRetriever now works consistently across all metric types
- Confirmed scores are properly normalized to 0-1 range

### Files Modified
- `packages/components/nodes/vectorstores/Milvus/Milvus.ts`

### DIFF
--- a/packages/components/nodes/vectorstores/Milvus/Milvus.ts
+++ b/packages/components/nodes/vectorstores/Milvus/Milvus.ts
@@ -395,7 +395,19 @@ const similaritySearchVectorWithScore = async (query: number[], k: number, vecto
                 }
             }
         })
-        results.push([new Document(fields), result.score])
+        
+        let normalizedScore = result.score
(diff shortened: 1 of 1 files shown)


HenryHengZJ (approved):
thank you!

PR #4909, opened 2025-07-21
fix(server): Add ordering expression to select list in chat messages …

Cranbaerry:
The previous query was failing with the error Error: chatMessagesService.getAllChatMessages - for SELECT DISTINCT, ORDER BY expressions must appear in select list because the ORDER BY clause was using a column that was not in the SELECT list.

This commit fixes the issue by adding the ordering column (MAX(chat_message.createdDate)) to the SELECT list and using an alias to refer to it in the ORDER BY clause.


HenryHengZJ:
this was fixed here: https://github.com/FlowiseAI/Flowise/pull/4865/files

will push to new release soon

almirb:
> this was fixed here: https://github.com/FlowiseAI/Flowise/pull/4865/files
> 
> will push to new release soon

Hi @HenryHengZJ !

Would it be possible to publish a new version, since this issue prevents us from viewing the chat log?

Thank you!

PR #4943, opened 2025-07-24
Add GitHub document loader streaming

russelj1:
This pull request adds a streaming mode for processing GitHub document loaders, allowing documents to be processed incrementally and stored in chunks. It also add a utility methods for managing loader statuses.


russelj1:
@HenryHengZJ just checking is there any changes necessary on this ?

PR #4944, opened 2025-07-24
[Feature?] - Speech to text to not override question parameter

fpmirabile:
I am in a use case where I need to send some information through question but the user can also send audio files.
I could use the state of course but it will make harder to make my LLMs work since I based everything on the user input. So I built this. Probably someone could use it. Maybe would be a good idea to make this under a flag instead of always?

Let me know your thoughts.


HenryHengZJ:
curious how are you using both at the same time?

fpmirabile:
> curious how are you using both at the same time?

Is a weird use case but I need to download urls from authenticated sources. So I created a tool to download them and reuse credentials from Flowise system and doing it all together here. Then those files are uploaded.

The only way I found easily for now is doing this.

Another is use is buffering messages on the api. So probably you can buffer messages and audio streams altogether

Issue #4653, opened 2025-06-13
[FEATURE] Improving the Loop Node - Agentflow V2


toi500:
The current state of the loop node is not at its best potential. Even though this node gives you the option to loop back to any previous node connected in its branch, the reality is that it only works when looping back to the LLM or Agent Node.

![Image](https://github.com/user-attachments/assets/ebb73261-9542-4a8b-b035-44fe00e8bd41)

### So we have several alternatives here:
1.  Make it as basic as the v1 one (current state) and limit the options that don't work, as I explained above.

2.  Implement custom logic to loop back node by node (very time consuming and not very realistic, I guess).

3.  Finding a middle ground between 1 and 2, allowing users to **insert variables into the loop node as the value to override in the target node.** Here is a very basic example to make this understandable:

    *   User wants to search about FLOWISE.
    *   The Brave search is executed and returns some data.
    *   A quality agent checks if the search is good enough.
    *   If not, it routes to the refine agent. The refine agent will propose new search terms.

    And here is the new thing:

    *   We would be able to, in the loop node, select the output of the Refine Search Terms agent {{agentAgentflow_0}}, forward that data back to the tool node again and OVERRIDE the tool node input (for this example, `{{question}}`).

If we do this right, this would provide Agentflow V2 a whole new dimension, IMHO.

**FINAL NOTE:** This sort of mapping data manually btw nodes is something that we do in n8n with no problem.

**FINAL NOTE 2:** Being able to update flow state in the loop node would do the job too.

![Image](https://github.com/user-attachments/assets/949fc649-367b-4884-9371-60e535a2a267)

HenryHengZJ:
I also have been thinking about this, additionally have a fallback option if the loop exhausted, maybe can use the last execution message or custom error message.

For starter, can you give me a couple of flows that can see the flaws and some notes around whats the ideal scenario looks like

toi500:
@HenryHengZJ 100% agree on the fallback for exhaustion. Here is my high level thinking on how we could implement this:

1.  For targeting an LLM or Agent node, by default, it should continue to work as it does right now.

2.  **Input overrides:**
    *   When a user selects a `Loop Back To` target node, ideally, the UI should dynamically fetch that current node's configurable inputs.
    *   The user can then map a variable to override a specific input for the next iteration of the loop.

3.  **Fallback on max loops:**
    *   **A) Integrated reply node functionality.** Add a text field directly inside the Loop Node for a fallback message if the loop exhausts.
[...]

PR #4957, opened 2025-07-27
Feature/Update Loop Agentflow

HenryHengZJ:
Include fallback message and version increment to 1.1

- Added a new input parameter 'fallbackMessage' to the Loop Agentflow for displaying a message when the loop count is exceeded.
- Incremented the version of Loop Agentflow from 1.0 to 1.1.
- Updated the processing logic to handle the fallback message appropriately when the maximum loop count is reached.

### DIFF
--- a/packages/components/nodes/agentflow/Loop/Loop.ts
+++ b/packages/components/nodes/agentflow/Loop/Loop.ts
@@ -19,7 +19,7 @@ class Loop_Agentflow implements INode {
     constructor() {
         this.label = 'Loop'
         this.name = 'loopAgentflow'
-        this.version = 1.0
+        this.version = 1.1
         this.type = 'Loop'
         this.category = 'Agent Flows'
         this.description = 'Loop back to a previous node'
@@ -40,6 +40,15 @@ class Loop_Agentflow implements INode {
                 name: 'maxLoopCount',
                 type: 'number',
                 default: 5
+            },
+            {
+                label: 'Fallback Message',
+                name: 'fallbackMessage',
+                type: 'string',
+                description: 'Message to display if the loop count is exceeded',
+                placeholder: 'Enter your fallback message here',
+                rows: 4,
+                acceptVariable: true
             }
         ]
     }
@@ -64,6 +73,7 @@ class Loop_Agentflow implements INode {
     async run(nodeData: INodeData, _: string, options: ICommonObject): Promise<any> {
         const loopBackToNode = nodeData.inputs?.loopBackToNode as string
         const _maxLoopCount = nodeData.inputs?.maxLoopCount as string
+        const fallbackMessage = nodeData.inputs?.fallbackMessage as string
 
         const state = options.agentflowRuntime?.state as ICommonObject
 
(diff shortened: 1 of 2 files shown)


HenryHengZJ:
This new addition allows user to specify fallback message after loop is exhaused, and use the new built in variable `{{ loop_count }}` to check current loop count, so a fallback action can be performed

image

The ideal way is to allow Loop node to have 2 outputs, 1 to loopback, 1 to fallback, but since we are not able to have the loopback edges, thus having just 1 fallback outgoing path is going to confuse user. Hence, we'd prefer to stick with no outgoing paths

image

toi500:
@HenryHengZJ following the conversation from here https://github.com/FlowiseAI/Flowise/issues/4653#issuecomment-3124383627, if what we want is simplicity for new users, the visual approach with 1 ouput works better, IMHO. This allows us to add a simple reply node, as we do in other nodes, or of if needed, create much more advanced applications by adding whatever new workflow logic we can implement after exhaustion. I don't know, it feels quite intuitive to me and provides a new dimension to the flow that it currently does not have.

chrome_TwSBe5nGmO

I know you want to get this simple, but this draft implementation I did took me 45 mins (only 2 files), just the time I needed to figure out the loop counting, but of course, as always, this is up to you.

Maybe a new toggle option to disable the visual output by default?

toi500:
By the way, I have never been a fan of the loopback edges, as other people suggest on Discord. In my opinion, they would just add visual complexity to the canvas, going against the data flow direction. And in any event, what prevents us from adding a non-functional, visual discontinued line ---- from the loop node to the target node if that is something users really want to have? It feels pointless to me, but could solve the two-output scenario for the loop node you were imagining.

toi500:
~By the way, I have never been a fan of the loopback edges, as other people suggest on Discord. In my opinion, they would just add visual complexity to the canvas, going against the data flow direction. And in any event, what prevents us from adding a non-functional, visual loopback edge (discontinued line ----) from the loop node to the target node if that is something users really want to have?~

~## Visual loopback edge
By "visual," I mean preserving the current logic of loop back to the target node but instead of using a drop-down menu, use it like this. It feels pointless to me (since drop-down menu works well), but could solve the 2-output scenario for the loop node you were visualizing.~

chrome_0m6u2xzDng

HenryHengZJ:
Good suggestion, I think the best approach is to add another toggle option - `Enable Fallback Action` that when turned on, will add the outgoing connection point, allowing user to connect the edge to other nodes. This will need lil bit more work from UI as well to have that dynamic capability - but I think its probably the best we can do

toi500:
@HenryHengZJ For what it's worth, I have aligned the fallback implementation (now working in all the scenarios and nodes I tested out) with the original code https://github.com/FlowiseAI/Flowise/compare/main...toi500:Flowise:exhausted-output-loop-node, making only minimal and targeted changes. If you want, I can make a draft PR so you can wrap this out with that new UI dynamic capability to make the fallback output disabled by default.

HenryHengZJ:
thanks, that's fine we dont have to open a draft PR, I can just refer to the link you pasted

HenryHengZJ:
merging this one and will open separate PR for adding `Enable Fallback Action`

PR #4977, opened 2025-07-30
Add logs

0xi4o:
- Adds logging for queuemanager (to debug bullmq connection issues)
- Adds better logging for buildChatflow (with org id, chatflow id, and chat id)
- Updates how queuemanager connects to redis (could be a potential fix for streaming not working)
  - currently uses both url and host+port+username+password
  - switched to conditional checking same as how redis pub-sub code connects to redis


HenryHengZJ:
how bout redis pub sub?

PR #4982, opened 2025-07-30
Fix: ignored agentflow LLM structuredOutput keys

russelj1:
This pull request fixes the issue where structuredOutput[key] values of 0 or false are ignored


HenryHengZJ (approved):
thanks!

PR #4983, opened 2025-07-30
Fix: dot notation for nested output variable resolution

russelj1:
This pull request improves how dot notation variable replacements are handled, ensuring that number and boolean types are handled correctly.

### DIFF
--- a/packages/server/src/utils/buildAgentflow.ts
+++ b/packages/server/src/utils/buildAgentflow.ts
@@ -351,8 +351,14 @@ export const resolveVariables = async (
                         const formattedValue =
                             Array.isArray(variableValue) || (typeof variableValue === 'object' && variableValue !== null)
                                 ? JSON.stringify(variableValue)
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
thanks! can you give me a scenario/flow for this?

russelj1:
The scenario I ran into was using the AgentFlow LLM component to generate structured json ie

```json
{
  "id": "llmAgentflow_0",
  "name": "llmAgentflow",
  "input": {},
  "output": {
    "content": "{\n  \"jql\": \"status = Closed\",\n  \"maxResults\": 10}",
    "timeMetadata": {
      "start": 1753882808973,
      "end": 1753882809792,
      "delta": 819
    },
    "maxResults": 10
  },
  "state": {},
  "chatHistory": []
}
```

and you want to use the maxResults via {{ llmAgentflow_0.output.maxResults }} in a tool call, currently that will return it as a string '10' rather than a number

PR #5039, opened 2025-08-07
Fixes state interpolation when using {{ output }}

corentin-hoareau:
When a state update includes the {{ output }} variable, the whole value is replaced by the node output.

I assume this is unintentional, so I've used replaceAll instead.


HenryHengZJ (approved):
thank you!

PR #5041, opened 2025-08-07
docs: clarify Node.js heap memory instructions for cross-platform usage

x85432:
As a Windows user, I noticed that the current instructions for resolving the "JavaScript heap out of memory" error only mention the export command, which is not valid in PowerShell or CMD.

Added instructions for:
- macOS / Linux / Git Bash
- Windows PowerShell
- Windows CMD

This makes it more accessible for users on different platforms, especially Windows users who may not be familiar with Bash syntax.


HenryHengZJ (approved):
thanks!

PR #5062, opened 2025-08-13
Text to speech

0xi4o:
- Adds a "Text to Speech" tab in flow configuration
- Users can configure two TTS providers — OpenAI and Eleven Labs
  - Configure credentials
  - Select available voices for each provider
  - Play a test audio for each voice
  - Option to automatically play audio after receiving a message from chatflow/agentflow
- Show speaker button below bot messages
  - Clicking speaker button will use the provided TTS configuration to convert text to audio
  - If "Automatically play audio" is enabled, play audio automatically


packages/ui/src/ui-component/extended/TextToSpeech.jsx
```diff
@@ -0,0 +1,624 @@
...
+            })
+            if (providerName !== 'none' && newVal['none']) {
+                newVal['none'].status = false
+            }
+        }
+        setTextToSpeech(newVal)
+        return newVal
+    }
+
+    const handleProviderChange = (provider, configOverride = null) => {
```
HenryHengZJ, line 195: after selecting, the popup doesnt close:

image

packages/ui/src/views/chatmessage/ChatMessage.jsx
```diff
@@ -1497,9 +1558,439 @@ const ChatMessage = ({ open, chatflowid, isAgentCanvas, isDialog, previews, setP
...
+                    isBuffering: true
+                }
+            } catch (error) {
+                console.error('Error appending chunk to buffer:', error)
+                return prevState
+            }
+        })
+    }
+
+    const handleTTSStart = (data) => {
```
HenryHengZJ, line 1882: When audio is auto played, the STOP button for chat doesnt go away unless the audio has finished playing:

image

As a user, I would've thought clicking that will stop the audio playing as well, but in fact nothing really happens, it just keep spinning.

Expected:
1.) Either clicking the chat STOP button will stop the audio play OR
2.) Just have the audio STOP button:
image

packages/ui/src/views/chatmessage/ChatMessage.jsx
```diff
@@ -1497,9 +1558,439 @@ const ChatMessage = ({ open, chatflowid, isAgentCanvas, isDialog, previews, setP
...
+        setIsTTSLoading((prevState) => ({
```
HenryHengZJ, line 1888: When the audio is playing, sending another chat trigger another audio as well, causing mix of audio streaming:

image

Expected: Only 1 audio stream at a time, as soon as another audio stream is played, stop the previous one

HenryHengZJ:
@0xi4o For the test audio, I changed it to this:
image

HenryHengZJ:
I've fixed:

## 1:
- Bug: Audio doesnt play the active selected one
- Fix: Updated the text-to-speech controller to select the active provider based on status instead of the first available provider

## 2:
- Added audio waveform controls and test audio functionality in the TextToSpeech component, allowing users to play and pause test audio

## 3:
- Bug: Selected voice does not displayed correctly
- Fix: Integrated Autocomplete for voice selection in the TextToSpeech component

## 4:
- Bug: When playing audio, it auto scroll to bottom
- Fix: Implemented TTS action management in ChatMessage to prevent auto-scrolling during TTS actions

## 5:
- Bug: When the audio is playing, sending another chat trigger another audio as well, causing mix of audio streaming
- Fix: Implemented stopAllTTS function calls to halt existing TTS audio before playing new audio or starting a new TTS stream

## 6:
- Bug: When select None for TTS, the speech volume button still exists
- Fix: Updated the condition for enabling TTS providers to exclude the 'none' provider, ensuring only valid providers are considered for text-to-speech functionality.

packages/ui/src/views/chatmessage/ChatMessage.jsx
HenryHengZJ: resolved

package.json
```diff
@@ -127,5 +127,8 @@
                 }
             ]
         ]
+    },
+    "dependencies": {
+        "@elevenlabs/elevenlabs-js": "^2.8.0"
```
HenryHengZJ: why are we adding this at root package json?

packages/server/src/utils/buildChatflow.ts
```diff
@@ -833,6 +900,38 @@ export const executeFlow = async ({
...
+        if (shouldAutoPlayTTS(chatflow.textToSpeech) && result.text) {
+            const options = {
+                orgId,
+                chatflowid,
+                chatId,
+                appDataSource,
+                databaseEntities
+            }
+
+            if (streaming && sseStreamer) {
```
HenryHengZJ: whats the difference between `if` and `else if ` when both are using `await generateTTSForResponseStream`

package.json
0xi4o: Added by mistake. Removed it.

packages/server/src/utils/buildChatflow.ts
0xi4o: This one was added early on when building TTS. Not needed anymore so removed it.

PR #5065, opened 2025-08-13
Fix chatflow's type null or blank

chungyau97:
## Description
Sets the `type` column in the `chat_flow` table to non-nullable and assigns a default value of `CHATFLOW`.

### Enterprise
| Database | Before | After |
| --- | --- | --- |
| SQLite | | |
| PostgreSQL | agentflowchatflowdatabase-datadatabase-migrationsdatabase-properties | agentflowchatflowdatabase-datadatabase-migrationsdatabase-properties |
| MySQL | | |
| MariaDB | | |

### Cloud
| Database | Before | After |
| --- | --- | --- |
| PostgreSQL |  |

### DIFF
--- a/packages/server/src/Interface.ts
+++ b/packages/server/src/Interface.ts
@@ -68,7 +68,7 @@ export interface IChatFlow {
     followUpPrompts?: string
     apiConfig?: string
     category?: string
-    type?: ChatflowType
+    type?: string
     workspaceId?: string
 }
 
--- a/packages/server/src/database/entities/ChatFlow.ts
+++ b/packages/server/src/database/entities/ChatFlow.ts
@@ -1,6 +1,13 @@
 /* eslint-disable */
 import { Entity, Column, CreateDateColumn, UpdateDateColumn, PrimaryGeneratedColumn } from 'typeorm'
-import { ChatflowType, IChatFlow } from '../../Interface'
+import { IChatFlow } from '../../Interface'
+
+export enum ChatflowType {
+    CHATFLOW = 'CHATFLOW',
+    AGENTFLOW = 'AGENTFLOW',
+    MULTIAGENT = 'MULTIAGENT',
+    ASSISTANT = 'ASSISTANT'
+}
 
 @Entity()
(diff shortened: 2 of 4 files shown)


packages/server/src/Interface.ts
```diff
@@ -68,7 +68,7 @@ export interface IChatFlow {
     followUpPrompts?: string
     apiConfig?: string
     category?: string
-    type?: ChatflowType
+    type?: string
```
HenryHengZJ: why to string not ChatflowType?

packages/server/src/database/entities/ChatFlow.ts
```diff
@@ -40,8 +47,8 @@ export class ChatFlow implements IChatFlow {
     @Column({ nullable: true, type: 'text' })
     category?: string
 
-    @Column({ nullable: true, type: 'text' })
-    type?: ChatflowType
+    @Column({ nullable: false, type: 'text', default: ChatflowType.CHATFLOW })
+    type?: string
```
HenryHengZJ: same here

packages/server/src/services/chatflows/index.ts
```diff
@@ -254,57 +263,51 @@ const saveChatflow = async (
...
-            dbResponse?.type === 'MULTIAGENT' ? FLOWISE_METRIC_COUNTERS.AGENTFLOW_CREATED : FLOWISE_METRIC_COUNTERS.CHATFLOW_CREATED,
-            { status: FLOWISE_COUNTER_STATUS.SUCCESS }
-        )
-
-        return dbResponse
-    } catch (error) {
-        throw new InternalFlowiseError(
-            StatusCodes.INTERNAL_SERVER_ERROR,
-            `Error: chatflowsService.saveChatflow - ${getErrorMessage(error)}`
+    validateChatflowType(newChatFlow.type)
```
HenryHengZJ, line 266: try catch block is not needed?

chungyau97: Yes, controller have try catch block.

If put try catch block here will cause 500 instead of 400.

PR #5067, opened 2025-08-13
Chore/remove redundant loggers

HenryHengZJ:


packages/server/src/queue/BaseQueue.ts
```diff
@@ -57,8 +57,6 @@ export abstract class BaseQueue {
     }
 
     public createWorker(concurrency: number = WORKER_CONCURRENCY): Worker {
-        logger.info(`[BaseQueue] Creating worker for queue "${this.queue.name}" with concurrency: ${concurrency}`)
```
anatolyburtsev, line 60: Hey! Don't have the whole context, but consider replacing with `logger.debug`, so it will be hidden in prod (if minimum logging level is set to info there). But would be much easier to enable it back if need to debug some weird connectivity issue.

HenryHengZJ: We have a weird Redis issue where after 1-2days workers are not able to stream the message back, we thought of adding these logs to help troubleshoot the bug, but unfortunately they are not helpful at all. We're reworking the logging functionality to have better traceability. Currently these logs are scattered, we dont know which is coming from which worker etc

PR #5096, opened 2025-08-17
Bugfix/Upsert API for file loader

HenryHengZJ:


grzegorz-wawrzak-ams:
Thank you very much, @HenryHengZJ, I really appreciate your work! 
One more question, will that be delivered within flowise@3.0.6?

PR #5111, opened 2025-08-20
feat: Add AWS DynamoDB KV Storage tool

anatolyburtsev:
## Summary
This PR adds a Key-Value Storage tool to Flowise, enabling persistent storage and retrieval of data with built-in versioning capabilities. It uses AWS DynamoDB as a backbone.

Tool has 2 modes: store or retrieval. 
- Store mode - stored object by key. Example response:  
```Successfully stored value with key "total_comp" at 2025-08-20T05:41:09.608Z```
- Retrieval mode - return object by key with timestamp when it was created. Consumes optional argument nth-latest, e.g. value `2` returns second last version of the object. Example response: 
```
{
  "value": "200",
  "timestamp": "2025-08-20T05:41:09.608Z"
}
```

## Note
Retrieval mode works great with JSONPathExtractor introduced [here](https://github.com/FlowiseAI/Flowise/pull/5052)

## Example use-case
Monitor a website for changes by storing snapshots and comparing different versions over time. This is useful for tracking competitor updates, detect when news articles are updated or corrected, or track feature announcements and product updates. Diff together with old and new versions can be feed directly to LLM to make sense of it.

  ## Implementation Details
  - Added shared AWS utilities in `/packages/components/src/awsToolsUtils.ts` for credential management
  - DynamoDB's has to have Partition key `pk` and Search key `sk`. Corresponding validation is done during configuration phase. The schema is
    - pk: The actual key name
    - sk: Timestamp as string (for sorting)
    - value: The actual data
    - timestamp: Human-readable ISO timestamp

  ## Configuration

  The tool requires:
  - AWS Credentials (via credential manager)
  - AWS Region selection
  - DynamoDB table name. 

  ## Testing
  - Added unit tests
  - Manual testing with actual AWS DynamoDB instances

Example of the flow:
Screenshot 2025-08-19 at 10 42 20 PM

Screenshot 2025-08-19 at 10 42 32 PM

Store mode node configuration:
Screenshot 2025-08-20 at 10 45 30 AM

Retrieval mode node configuration:
Screenshot 2025-08-20 at 10 45 40 AM

### DIFF
--- a/packages/components/nodes/tools/AWSSNS/AWSSNS.ts
+++ b/packages/components/nodes/tools/AWSSNS/AWSSNS.ts
@@ -1,6 +1,7 @@
 import { Tool } from '@langchain/core/tools'
 import { ICommonObject, INode, INodeData, INodeOptionsValue, INodeParams } from '../../../src/Interface'
-import { getBaseClasses, getCredentialData, getCredentialParam } from '../../../src/utils'
+import { getBaseClasses } from '../../../src/utils'
+import { AWS_REGIONS, DEFAULT_AWS_REGION, getAWSCredentials } from '../../../src/awsToolsUtils'
 import { SNSClient, ListTopicsCommand, PublishCommand } from '@aws-sdk/client-sns'
 
 class AWSSNSTool extends Tool {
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: packages/components/src/awsToolsUtils.ts [hunk 1]
```diff
@@ -0,0 +1,65 @@
+import { ICommonObject, INodeData } from './Interface'
+import { getCredentialData, getCredentialParam } from './utils'
+
+// AWS Regions constant
+export const AWS_REGIONS = [
```
HenryHengZJ, line 5: can we have `AWSSNS.ts` reuse this as well?

OrhanA-Innodare:
Dear Anatolyburtsev,
Thank you for developing the AWS DynamoDB tool for Flowise. I understand that it can only store the attributes one by one and when asked for Retrieve gives only the latest one.  But if you need to get the latest attributes for a specific user, if I have more than one user, I could not find anyway of doing it. I might have 100 users, and I want to get the latest attributes for a user. It may be that, in the data base there are 50 other users which loaded data on the table before the specific user that I want to get the data for. Is there any way of doing this in flowise or is that tool for a limited use only?  Thank you in advance

PR #5130, opened 2025-08-24
docker: expose DATABASE_URL in Dockerfile for container deployments

sawenaa:
## Summary
Adds `ENV DATABASE_URL=${DATABASE_URL}` to the Dockerfile so platform deployments
(Render, k8s, Docker Compose, etc.) can configure Flowise with an external DB
(Postgres or a SQLite file path) via environment variable.

## Motivation
Many hosts inject DB connection strings at runtime as env vars. Today, users
either rebuild images or rely on defaults. This change makes the variable
explicit in the official Docker image so it can be set cleanly at runtime.

## Change
```diff
-  # previous lines…
+  ENV DATABASE_URL=${DATABASE_URL}
   WORKDIR /usr/src


HenryHengZJ:
Can you not set the env variable when running docker compose? https://github.com/FlowiseAI/Flowise/blob/main/docker/docker-compose.yml

Issue #5133, opened 2025-08-25
Blank page after copying empty values in Format Prompt Values


lcanavesio:
### Describe the bug

When opening the Format Prompt Values dialog and leaving the input field empty, clicking on Copy to clipboard throws a console error, leaves the page completely blank, and any unsaved flow is lost.

### To Reproduce

Go to the Format Prompt Values section.
Leave the first field (input) empty ("").
Click the Copy to clipboard icon.
The page turns blank and the following error appears in the console:

`index-C6GKaUTA.js:62 TypeError: Cannot read properties of undefined (reading 'writeText')
    at V (CredentialListDialog-BNh2u_Ji.js:6:1433)
    at enableClipboard (CredentialListDialog-BNh2u_Ji.js:6:2124)
    at t.<anonymous> (main-zDAGC98s.js:13:18279)`

<img width="840" height="3979" alt="Image" src="https://github.com/user-attachments/assets/dd7977d5-11e1-4a0f-8a9d-00d9095b2494" />

### Expected behavior

It should either:
Copy an empty string ("") to the clipboard, or
Gracefully handle the undefined value without throwing an error or refreshing the page.

### Use Method

Docker

### Flowise Version

3.0.5	

### Operating System

Linux

### Browser

Chrome

PR #5152, opened 2025-08-29
fix: handle empty input and missing clipboard API in Format Prompt Values dialog

sauravbaid3:
Problem

- When leaving the input field empty in the Format Prompt Values dialog and clicking on Copy to clipboard, the app:
Throws a console error:
TypeError: Cannot read properties of undefined (reading 'writeText')

- Turns the page blank

- Causes unsaved flow data to be lost

Root Cause

- navigator.clipboard.writeText was being called without checking if the Clipboard API was available
- Empty string values ("") were not being handled safely, leading to JSON parsing and clipboard issues

Solution

This PR improves clipboard handling across components by:

- Adding Clipboard API availability checks (if (navigator?.clipboard?.writeText))

- Wrapping copy operations in try/catch blocks for safer error handling

- Gracefully handling empty values ("") without breaking the UI

- Applying the same safety pattern consistently across components:
       JsonEditor.jsx
       ExpandedChunkDialog.jsx
       NodeExecutionDetails.jsx
       
Result

No more crashes or blank pages when copying empty/invalid input
Consistent clipboard handling across the app
Better browser compatibility and error resilience

### DIFF
--- a/packages/ui/src/ui-component/json/JsonEditor.jsx
+++ b/packages/ui/src/ui-component/json/JsonEditor.jsx
@@ -17,7 +17,15 @@ export const JsonEditorInput = ({
     isDarkMode = false,
     isSequentialAgent = false
 }) => {
-    const [myValue, setMyValue] = useState(value ? JSON.parse(value) : {})
+    const [myValue, setMyValue] = useState(() => {
+        if (!value || value === '') return {}
+        try {
+            return JSON.parse(value)
+        } catch (error) {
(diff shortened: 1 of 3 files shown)


HenryHengZJ:
image

I left it empty, and clicked Copy, it seems to work just fine?

HenryHengZJ:
https://github.com/FlowiseAI/Flowise/issues/5133

Issue #5000, opened 2025-08-01
BUG: Full file upload not working anymore (PDF Use Legacy Build)


okurayhan:
### Describe the bug

BUG: Full file upload not working anymore

### To Reproduce

Enable full file upload.
Upload pdf to chat screen. And could not upload document.

### Expected behavior

Can upload pdf document.

### Use Method

None

### Operating System

None

### Browser

None

HenryHengZJ:
Is this cloud or self hosted?

Self-hosted:

<img width="1358" height="795" alt="Image" src="https://github.com/user-attachments/assets/9a7c61c3-ff3d-4298-9e14-df83191c62e0" />

Cloud:

<img width="1416" height="773" alt="Image" src="https://github.com/user-attachments/assets/7eb79da4-ecda-49f5-994c-ddd398937db0" />

Both works fine

okurayhan:
Self-hosted version:

Checked: Use Legacy Build (for PDF compatibility issues)

`Error: Cannot find module 'pdfjs-dist/legacy/build/pdf.js'

<img width="2504" height="1346" alt="Image" src="https://github.com/user-attachments/assets/781ba4f2-93b2-4656-93c5-0522e2373b31" />

This commit may caused error.
https://github.com/FlowiseAI/Flowise/commit/ee5ab1bd6dc7520a2632cdd2ec7b929643e07fd4

okurayhan:
> Is this cloud or self hosted?
> 
> Self-hosted:
> 
[...]

PR #5175, opened 2025-09-07
Update pdfjs-dist version to ^3.11.174

okurayhan:
The file `pdfjs-dist/legacy/build/pdf.js` is only available in version 3.11.174 of `pdfjs-dist`. In later versions, this file is missing. Therefore, the version has been set to 3.11.174

### DIFF
--- a/packages/components/package.json
+++ b/packages/components/package.json
@@ -131,7 +131,7 @@
         "openai": "^4.96.0",
         "papaparse": "^5.4.1",
         "pdf-parse": "^1.1.1",
-        "pdfjs-dist": "^5.3.93",
+        "pdfjs-dist": "^3.11.174",
         "pg": "^8.11.2",
         "playwright": "^1.35.0",
         "puppeteer": "^20.7.1",


HenryHengZJ:
There's a security vulnerability in that version, any reason why do you still want to use the legacy build? can the latest version works as expected?

HenryHengZJ:
https://github.com/FlowiseAI/Flowise/issues/5000

okurayhan:
> #5000

I need the legacy build because in some PDFs, when there are numeric values in table fields, the latest version fails to extract them properly and the numbers overlap. With the legacy build, the extraction works correctly.

HenryHengZJ:
can you give me the example PDF file?

okurayhan:
> can you give me the example PDF file?

Here is the pdf file and extracted text.
[test.pdf](https://github.com/user-attachments/files/22621561/test.pdf)

CleanShot 2025-09-30 at 17 26 12@2x

PR #5184, opened 2025-09-10
feat(ui): adapt GitHub MCP server for the new version

nogataka:
The GitHub MCP server currently used in Flowise (@modelcontextprotocol/server-github) was deprecated in April 2025. From that point onward, it will no longer receive updates, including security patches and bug fixes. Therefore, it is necessary to use an alternative MCP server or another method of connecting to GitHub. To support the new MCP server as the migration destination, I created code and submitted a PR.

Migration source:
https://github.com/modelcontextprotocol/servers-archived/tree/main/src/github

Migration destination:
http://github.com/github/github-mcp-server

Thanks in advance!


HenryHengZJ:
Thanks for the update! I think we can just directly replace the old `GithubMCP.ts` with your new implementation, dont need to create a separate one

PR #5211, opened 2025-09-15
Bugfix/add validation for file path

HenryHengZJ:


bengoetzinger:
Hi Henry, 
what has changed here? I had documents for the "Folder with documents" node in /root/.flowise/document_store, but after update there was the message "Error: Invalid folder path: Path traversal detected. Please provide a safe folder path." So I moved the folder to /root/.flowise/storage/document_store and adapted the folder path in the node, but the problem persists. What am I doing wrong here? Reverted back to 3.0.6 for production.
Thanks for your great work, btw!

mf-sk:
Hi, there seems to be more confused users - https://github.com/FlowiseAI/Flowise/discussions/5326.
I understand that this is a security check to prevent path traversal. However, this makes it difficult to use legitimate absolute paths within the Flowise environment, for example when loading files from a pre-defined data folder.
It would be helpful if either absolute paths inside Flowise’s allowed workspace could be used without triggering a false positive, or
there is a clear documentation on how to structure folder paths so that they pass the check.
This would make the upsert endpoint easier to use in controlled environments.

A quick workaround for the issue is to either downgrade or modify the relevant part of the Folder.js file to:

>         if ((0, validator_1.isPathTraversal)(folderPath)) {
>             console.log('Invalid folder path: Path traversal detected. Please provide a safe folder path.');
>         }

Thank you for looking into this!

Best regards,
Martin

PR #5219, opened 2025-09-15
fix: add Redis connection resilience settings to prevent infinite hangs

### LINKED ISSUE #5126: Redis connection failures cause indefinite blocking in queue mode with no timeout or recovery mechanism
### Describe the bug

When Redis connection is lost or becomes unstable (ECONNRESET errors), Flowise enters a state where:
  - UI appears functional and responsive
  - Message requests are accepted but never respond
  - System eventually returns 504 Gateway Timeout
  - No automatic recovery mechanism exists
  - Entire system becomes unusable until manual restart

  Root Cause: buildChatflow.ts:986 contains await job.waitUntilFinished(queueEvents) with no timeout, causing indefinite blocking when Redis queue jobs cannot be processed.

  Architecture Impact: Redis becomes a single point of failure that can render the entire system non-functional, even though core AI processing could work fine without queuing.

### To Reproduce

  1. Deploy Flowise in queue mode (MODE=queue) with Redis backend (e.g., Google Cloud Memorystore)
  2. Configure with unstable TLS setup (e.g., NODE_TLS_REJECT_UNAUTHORIZED=0) or network conditions causing Redis ECONNRESET
  3. Send a message through the UI
  4. Observe Redis connection drops/resets in logs
  5. Message request hangs indefinitely
  6. All subsequent message requests also hang
  7. System requires restart to recover

### Expected behavior

  Resilient behavior should include:
  - Job processing timeout configuration (e.g., JOB_TIMEOUT environment variable)
  - Graceful fallback to direct processing when Redis is unavailable
  - Circuit breaker pattern for Redis connection failures
  - Automatic retry logic with exponential backoff
  - Health checks that detect Redis connectivity issues
  - Error responses instead of indefinite hanging

### Screenshots

_No response_

### Flow

_No response_

### Use Method

Docker

### Flowise Version

3.0.4

### Operating System

Linux

### Browser

None

### Additional context

Code Location: /packages/server/src/utils/buildChatflow.ts:986
  // Currently: No timeout specified
  const result = await job.waitUntilFinished(queueEvents)

  // Should be: Configurable timeout
  const result = await job.waitUntilFinished(queueEvents, {
    timeout: parseInt(process.env.JOB_TIMEOUT || '300000') // 5 min default
  })

  Missing Resilience Patterns:
  - No timeout on job.waitUntilFinished()
  - No fallback when Redis unavailable
  - No circuit breaker for Redis failures
  - Health checks don't verify Redis connectivity
  - No job processing timeout configuration via environment variables

  BullMQ Support: BullMQ v5.45.2 supports timeouts, but Flowise doesn't expose this configuration.

utsav-bx:
## Problem
  Fixes #5126 - Flowise hangs indefinitely when Redis connections drop, causing `ECONNRESET` errors during queue processing.

 ## Solution
  This PR implements a two-part fix to prevent indefinite hangs:

  ### 1. Redis Connection Resilience (Optional Configuration)
  Adds configurable Redis connection settings via environment variables **without hardcoded defaults**:
  - `REDIS_CONNECT_TIMEOUT` - Connection timeout in milliseconds
  - `REDIS_COMMAND_TIMEOUT` - Command timeout in milliseconds
  - `REDIS_MAX_RETRIES_PER_REQUEST` - Maximum number of retries per request
  - `REDIS_LAZY_CONNECT` - Enable lazy connection (true/false)

  These settings are **only applied if explicitly set**, allowing Redis to use its own defaults otherwise.

  ### 2. Job Execution Timeouts (Core Fix)
  Adds `JOB_TIMEOUT` environment variable (default: 300000ms / 5 minutes) and implements explicit timeouts on all `job.waitUntilFinished()` calls across:
  - `buildChatflow.ts`
  - `agentflowv2-generator/index.ts`
  - `documentstore/index.ts` (5 occurrences)
  - `nodes/index.ts`
  - `upsertVector.ts`

  This prevents jobs from hanging indefinitely when Redis connections fail.

  ## Key Changes
  - Enhanced `packages/server/src/queue/QueueManager.ts` with conditional Redis resilience settings
  - Updated `packages/server/.env.example` with new configuration options
  - Added explicit timeouts to all 9 `waitUntilFinished()` calls
  - Maintains full backward compatibility with existing Redis environment variables

  ## Addresses PR Feedback
  - ✅ Removed hardcoded defaults per @HenryHengZJ's feedback
  - ✅ Uses conditional property spreading - only includes settings when env vars are defined
  - ✅ Allows Redis to use its own internal defaults when not configured

  ## Testing
  - ✅ Build passes with `pnpm build`
  - ✅ TypeScript compilation successful
  - ✅ All existing Redis env vars preserved
  - ✅ Backward compatible - works with or without new env vars

### DIFF
--- a/packages/server/.env.example
+++ b/packages/server/.env.example
@@ -167,14 +167,11 @@ JWT_REFRESH_TOKEN_EXPIRY_IN_MINUTES=43200
 # REDIS_KEY=
 # REDIS_CA=
 # REDIS_KEEP_ALIVE=
-
-# Redis Connection Resilience Settings (fixes infinite hangs on connection drops)
-# REDIS_CONNECT_TIMEOUT=10000 # Connection timeout in milliseconds
-# REDIS_COMMAND_TIMEOUT=30000 # Command timeout in milliseconds (prevents waitUntilFinished hanging)
-# REDIS_MAX_RETRIES_PER_REQUEST=3 # Maximum retries per Redis command
-# REDIS_LAZY_CONNECT=true # Connect only when needed
-# REDIS_KEEPALIVE_INTERVAL=30000 # Keep-alive interval in milliseconds
-
+# REDIS_CONNECT_TIMEOUT= # Connection timeout in milliseconds
+# REDIS_COMMAND_TIMEOUT= # Command timeout in milliseconds
+# REDIS_MAX_RETRIES_PER_REQUEST= # Maximum number of retries per request
+# REDIS_LAZY_CONNECT= # Enable lazy connection (true/false)
+# JOB_TIMEOUT=300000 # Job execution timeout in milliseconds (default: 5 minutes)
 # ENABLE_BULLMQ_DASHBOARD=
 
 
--- a/packages/server/src/services/agentflowv2-generator/index.ts
+++ b/packages/server/src/services/agentflowv2-generator/index.ts
@@ -210,7 +210,8 @@ const generateAgentflowv2 = async (question: string, selectedChatModel: Record<s
             })
             logger.debug(`[server]: Generated Agentflowv2 Job added to queue: ${job.id}`)
             const queueEvents = predictionQueue.getQueueEvents()
-            response = await job.waitUntilFinished(queueEvents)
+            const jobTimeout = process.env.JOB_TIMEOUT ? parseInt(process.env.JOB_TIMEOUT, 10) : 300000 // 5 minutes default
+            response = await job.waitUntilFinished(queueEvents, jobTimeout)
         } else {
             response = await generateAgentflowv2_json(
                 { prompt, componentNodes: getRunningExpressApp().nodesPool.componentNodes, toolNodes, selectedChatModel },
--- a/packages/server/src/services/documentstore/index.ts
+++ b/packages/server/src/services/documentstore/index.ts
@@ -683,7 +683,8 @@ const previewChunksMiddleware = async (
             logger.debug(`[server]: [${orgId}]: Job added to queue: ${job.id}`)
 
             const queueEvents = upsertQueue.getQueueEvents()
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -127,5 +127,6 @@
                 }
             ]
         ]
-    }
+    },
+    "packageManager": "pnpm@10.15.0+sha512.486ebc259d3e999a4e8691ce03b5cac4a71cbeca39372a9b762cb500cfdf0873e2cb16abe3d951b1ee2cf012503f027b98b6584e4df22524e0c7450d9ec7aa7b"
```
HenryHengZJ: what's the purpose of this?

utsav-bx: This is just to pin the package manager version to avoid package conflicts arising due to different package manager. Completely optional change.

HenryHengZJ: can we remove this change?

packages/server/src/queue/QueueManager.ts
```diff
@@ -15,6 +15,13 @@ import { ExpressAdapter } from '@bull-board/express'
 
 const QUEUE_NAME = process.env.QUEUE_NAME || 'flowise-queue'
 
+// Redis resilience configuration with environment variable support
+const REDIS_CONNECT_TIMEOUT = process.env.REDIS_CONNECT_TIMEOUT ? parseInt(process.env.REDIS_CONNECT_TIMEOUT, 10) : 10000
```
HenryHengZJ: I prefer not to hardcode a default value for these.

If these are set, we only pass it in, if not, we leave it

utsav-bx: Sorry, I had missed this comment. Have made the requested feedback now. Ready for review.

PR #5226, opened 2025-09-17
Add environment variable control for trust proxy setting

chungyau97:
## Description
This PR adds the ability to configure Express's trust proxy setting through an environment variable. By default, the setting remains `true` to maintain backward compatibility, but can now be explicitly disabled by setting TRUST_PROXY=false in the environment.

## Documentation
image

| TRUST_PROXY = undefined | TRUST_PROXY = true | TRUST_PROXY = false | TRUST_PROXY = 3 | TRUST_PROXY = loopback | TRUST_PROXY = loopback, linklocal, 123.123.123.123 |
| --- | --- | --- | --- | --- | --- |
| undefined | true | false | number | string | csv | 

**Enterprise**
image

**Cloud**
image


packages/server/src/index.ts
```diff
@@ -163,7 +163,8 @@ export class App {
         this.app.use(express.urlencoded({ limit: flowise_file_size_limit, extended: true }))
 
         // Enhanced trust proxy settings for load balancer
-        this.app.set('trust proxy', true) // Trust all proxies
+        const trust_proxy: boolean = process.env.TRUST_PROXY === 'false' ? false : true // Default to trust all proxies
```
hob: The change mostly looks great.  The trust proxy value isn't binary, though.  It also supports more complex networking configurations:
https://expressjs.com/en/guide/behind-proxies.html

You could support the full range of values with this:
```
const trust_proxy: string = process.env.TRUST_PROXY || "true" // Default to trust all proxies
```

It would probably also be a good idea to put a link to that documentation along side the value in the .env.example files so folks know what to expect from it.

HenryHengZJ:
@hob is good with the changes

PR #5227, opened 2025-09-17
Fix issu in gpt-5*, o1*, and o3* models as not supporting `stop` to avoid 400 Unsupported parameter errors.

DyanGalih:
fix(agents): avoid sending unsupported 'stop' param to models (e.g. gpt-5-nano)

Detect whether a model supports the `stop` parameter and only bind stop sequences when supported. Prevents 400 errors from models that reject `stop`. Files touched: ConversationalAgent, agents utils, XMLAgent (conditional stop binding).

### DIFF
--- a/packages/components/src/agents.ts
+++ b/packages/components/src/agents.ts
@@ -24,6 +24,7 @@ import {
 } from 'langchain/agents'
 import { formatLogToString } from 'langchain/agents/format_scratchpad/log'
 import { IUsedTool } from './Interface'
+import { modelSupportsStop } from './utils'
 import { getErrorMessage } from './error'
 
 export const SOURCE_DOCUMENTS_PREFIX = '\n\n----FLOWISE_SOURCE_DOCUMENTS----\n\n'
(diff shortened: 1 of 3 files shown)


HenryHengZJ:
The latest version should already fixed that, see [here](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/nodes/chatmodels/AzureChatOpenAI/AzureChatOpenAI.ts#L278)

DyanGalih:
This bugfix is related to the unsupported 'stop' message for all gpt-5*, o1*, and o3* models.

This is the issue in the main branch:
image

this is from my curent branch
image

packages/components/src/agents.ts
```diff
@@ -24,6 +24,7 @@ import {
 } from 'langchain/agents'
 import { formatLogToString } from 'langchain/agents/format_scratchpad/log'
 import { IUsedTool } from './Interface'
+import { modelSupportsStop } from './utils'
```
HenryHengZJ, line 27: where does this come from? I dont see any changes in utils file?

HenryHengZJ:
Or, can you try using AgentflowV2?

packages/components/src/agents.ts
DyanGalih: I added the changes from the utils.ts file. Please check. Thanks

PR #5231, opened 2025-09-18
Chore/Disable Available Dep By Default

HenryHengZJ:
Disable available dependencies by default, only allow when `ALLOW_BUILTIN_DEP` is set to true


toi500:
@HenryHengZJ Isn't this a breaking change? Several apps I had with custom function nodes suddenly broke. I just got lucky and saw this PR by chance.
 
image

HenryHengZJ:
There are multiple securities vulnerabilities that causes remote code execution have been reported and published widely. Unfortunately we have to perform the strick check by default. You can turn that off via the env variable

HenryHengZJ:
Added fix to allow `node-fetch` and `axios` by default: https://github.com/FlowiseAI/Flowise/pull/5264

PR #5284, opened 2025-10-01
Ability to support variable in the URL for HTTP Node.

ourdex:
Currently HTTP Node doesn't support the URL to be dynamically set from the variables. 

So, enabled this flag


ourdex:
@HenryHengZJ , can you check this?

okurayhan:
I need this.

HenryHengZJ:
thanks!

agroho:
Variable dropdown not triggered if 'URL' fields contain any value. It only trigger if URL field is empty. Please fix this.

PR #5294, opened 2025-10-06
Updates to change/reset password functionality

0xi4o:
- Change password now requires old password to be filled and validated before applying new password
- Migrate all platforms - oss, cloud, and enterprise - to use account settings page
- When a user changes/resets password, delete all the sessions for that user id

### DIFF
--- a/packages/ui/src/routes/MainRoutes.jsx
+++ b/packages/ui/src/routes/MainRoutes.jsx
@@ -50,7 +50,6 @@ const Evaluators = Loadable(lazy(() => import('@/views/evaluators')))
 
 // account routing
 const Account = Loadable(lazy(() => import('@/views/account')))
-const UserProfile = Loadable(lazy(() => import('@/views/account/UserProfile')))
 
 // files routing
 const Files = Loadable(lazy(() => import('@/views/files')))


packages/server/src/enterprise/services/user.service.ts
```diff
@@ -158,10 +158,18 @@ export class UserService {
...
-                const hash = bcrypt.hashSync(newUserData.password, salt)
+            if (newUserData.oldPassword && newUserData.newPassword && newUserData.confirmPassword) {
+                if (!oldUserData.credential) {
+                    throw new InternalFlowiseError(StatusCodes.BAD_REQUEST, UserErrorMessage.INVALID_USER_CREDENTIAL)
+                }
+                // verify old password
+                if (!compareHash(newUserData.oldPassword, oldUserData.credential)) {
+                    throw new InternalFlowiseError(StatusCodes.UNAUTHORIZED, UserErrorMessage.INVALID_USER_CREDENTIAL)
+                }
+                if (newUserData.newPassword !== newUserData.confirmPassword) {
```
hob, line 170: For consistency, shouldn't this also use the compareHash() function to do the comparison?

packages/server/src/enterprise/services/user.service.ts
```diff
@@ -158,10 +158,18 @@ export class UserService {
                 this.validateUserStatus(newUserData.status)
             }
 
-            if (newUserData.password) {
-                const salt = bcrypt.genSaltSync(parseInt(process.env.PASSWORD_SALT_HASH_ROUNDS || '5'))
-                // @ts-ignore
-                const hash = bcrypt.hashSync(newUserData.password, salt)
+            if (newUserData.oldPassword && newUserData.newPassword && newUserData.confirmPassword) {
+                if (!oldUserData.credential) {
```
hob, line 163: Is this because you can't change your password if you authenticated with social login?  If that's the case, is there nothing more explicit on the user model that could be used to drive this condition rather than the presence or absence of a password?

0xi4o: That's correct. There's no login type column on the user model.

packages/server/src/enterprise/services/user.service.ts
0xi4o: The `compareHash` function compares a plain-text password to an already-hashed password. It's faster to compare the new password and confirm password in plain-text than hashing one of them and then calling `compareHash`. The end result is the same.

HenryHengZJ:
@0xi4o as part of the PR, can we also disable user from changing their email for security purpose?

mbiesiad:
Hi @HenryHengZJ if I may, this is not a good idea - it will be security issue in this case. The user should be able to change their password (e.g. in case of compromise), as well as change their email address (both are credentials).

HenryHengZJ:
okay then ignore my comment, lets keep this PR as password change feature

hob:
lgtm :shipit:

HenryHengZJ:
1 small bug: if the old password is incorrect, I saw the error message, when refresh the page, it still logs me out - is this expected?

0xi4o:
> 1 small bug: if the old password is incorrect, I saw the error message, when refresh the page, it still logs me out - is this expected?

Fixed it.

packages/ui/src/routes/MainRoutes.jsx
```diff
@@ -308,10 +304,6 @@ const MainRoutes = {
                 </RequireAuth>
             )
         },
-        {
```
HenryHengZJ, line 311: user profile no longer exists so imports fail

0xi4o: Fixed

PR #5309, opened 2025-10-10
remove allowed origins from public chatbot config response

0xi4o:


0xi4o:
Closing in favor of #5310

Issue #4479, opened 2025-05-22
[BUG] Memory leak on API upsert request


cuca24:
**Describe the bug**
When i use API request to upsert document to vector store i get memory leak, For 3k items it usses additional 5GB of ram and it doesn't free up. It fills up memory until PC crashes. 

I setup TextFile and enabled API and enabled Override Configuration for text file. I'm sending this file via API curl requests from another system

**To Reproduce**
here is simple flow that leaks

![Image](https://github.com/user-attachments/assets/77bce520-9767-4f16-953f-5d4d962f4b74)

**Expected behavior**
Memory would free up after some time

**Setup**

-   Installation docker,
-   Flowise Version latest
-   OS: [Linux]

HenryHengZJ:
Can you try using the Document Store Upsert API? https://docs.flowiseai.com/using-flowise/document-stores#upsert-api

cuca24:
I already tried with document store, but i really don't need to have additional store for the data. I just need to send the data to vectore store in one go, which does works nicely by sending text, but unfortunately over time it leaks memory.  

This is might connected: https://github.com/FlowiseAI/Flowise/discussions/3733
I tried to use  @bernejorge image but it just won't start the container.

toi500:
> I already tried with document store, but i really don't need to have additional store for the data. I just need to send the data to vectore store in one go, which does works nicely by sending text, but unfortunately over time it leaks memory.
> 
> This is might connected: [#3733](https://github.com/FlowiseAI/Flowise/discussions/3733) I tried to use [@bernejorge](https://github.com/bernejorge) image but it just won't start the container.

I see a similar memory leak issue with Custom MCPs, losing around 150MB every time you add a Custom MCP node (even deleting all flows will not relase it). Over time, during iterative testing, memory consumption can escalate to around 3 or 4 GB easily.

The only way I found to reclaim the leaked memory was by restarting the Flowise container via the implementation of a cron job on the host server to automate the restart process.

toi500:
> that was already resolved with the fix [#4375](https://github.com/FlowiseAI/Flowise/pull/4375) no?

No, you leave it open for further investigation https://github.com/FlowiseAI/Flowise/issues/4369#issuecomment-2853106883

AbhishekHupele:
Hi, facingg same issue in agent v2, for sequential agentflow, any update ?

AbhishekHupele:
[...]

PR #5315, opened 2025-10-11
Ensure proper cleanup of the database connection in `createSchema`

### LINKED ISSUE #4255: Record manager keeps open connections to mysql
**Describe the bug**
When embedding the content and using Record Manager it seems like it keeps open connections to mysql and opens up a new connection on each api request. I soon get Too many connections error from mysql 

**Expected behavior**
Connection should be terminated once the API request finishes 

**Setup**

-   Installation docker
-   Flowise Version 2.2.7-patch.1
-   OS: Ubuntu
-   Browser Chrome

cuca24:
fixes #4255 
Potentially fixes the #4479

As per TypeORM documentation Disconnection (closing all connections in the pool) occurs when the destroy method is called.

All methods in MysqlRecordManager contains destroy call except createSchema call. 

I tested locally and this keeps connection to 0 after the calls.

### DIFF
--- a/packages/components/nodes/recordmanager/SQLiteRecordManager/SQLiteRecordManager.ts
+++ b/packages/components/nodes/recordmanager/SQLiteRecordManager/SQLiteRecordManager.ts
@@ -179,8 +179,8 @@ class SQLiteRecordManager implements RecordManagerInterface {
     }
 
     async createSchema(): Promise<void> {
+        const dataSource = await this.getDataSource()
         try {
-            const dataSource = await this.getDataSource()
             const queryRunner = dataSource.createQueryRunner()
             const tableName = this.sanitizeTableName(this.tableName)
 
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
thanks! with this fix, have you tried testing again and see if that free up the ram?

cuca24:
@HenryHengZJ 
I tested both with this fix and without this fix. And i can confirm that this fix also fixes issue i reported in https://github.com/FlowiseAI/Flowise/issues/4479 

Without a fix memory consumption rises constantly, while with the fix, it always returns to the same memory usage after i run thousands api requests. I even run with 20 parallel processes at once and it seems to be working ok. Memory does rises a bit but it's not leaking (it's does not rise constantly) and after it finishes it returns to the memory consumption that was before making upsert api calls.

HenryHengZJ:
are you able to build and run it? because `await this.getDataSource('createSchema')` is throwing an error, `getDataSource` doesnt take in any input parameter

cuca24:
@HenryHengZJ yes it should be  const dataSource = await this.getDataSource(). without additional param. I used it for testing purposes but forgot to clean it out.

PR #5310, opened 2025-10-10
Fix: CORS-related issues

0xi4o:
## Summary by CodeRabbit

* **New Features**
  - Domain-aware validation for prediction requests with clearer unauthorized-origin messages.
  - Dynamic CORS handling that evaluates origins per-request and applies special rules for prediction endpoints.

* **Changes**
  - Public chatbot configuration responses no longer expose origin-related fields to external clients.

PR #5335, opened 2025-10-16
Fixed environment level CORS validation with right substring check

vasunous:


hob:
:shipit:

packages/server/src/utils/XSS.ts
```diff
@@ -28,10 +28,15 @@ export function getCorsOptions(): any {
     const corsOptions = {
         origin: function (origin: string | undefined, callback: (err: Error | null, allow?: boolean) => void) {
             const allowedOrigins = getAllowedCorsOrigins()
-            if (!origin || allowedOrigins == '*' || allowedOrigins.indexOf(origin) !== -1) {
+            if (!origin || allowedOrigins == '*') {
                 callback(null, true)
             } else {
-                callback(null, false)
+                const allowedOriginsList = allowedOrigins.split(',').map((o) => o.trim().toLowerCase())
```
wowmsi, line 34: This can be pushed up for better readability and then the `if` condition can be rewritten asL

```
if (!origin || allowedOrigins == '*' || allowedOriginsList.includes(origin.toLowerCase())) {
     callback(null, true)
} else {
    callback(null, false)
}
```

vasunous: > allowedOriginsList.includes

I don't want to split "allowedOrigins" if it is just '*'. If it is not * then only it splits and compares with origin.

0xi4o:
Thanks @vasunous. I've integrated your changes in #5310. Closing this PR.

PR #5348, opened 2025-10-21
Fix/variable resolution to support arrays of config objects

HenryHengZJ:
## Problem
The `resolveVariables` function in `buildAgentflow.ts` was not properly resolving variables in array-based configuration objects. When processing parameters like `agentTools` (which is an array), only the first item's config object was being processed, leaving subsequent array items with unresolved variables.

## Root Cause
1. The `findParamValue` function only returned the first match found in arrays
2. The `processConfigParams` function didn't handle arrays of config objects
3. The processing logic only handled single config objects, not arrays of configs

## Solution
- **Enhanced `processConfigParams`**: Added array handling to recursively process each item in config arrays
- **Created `findAllConfigValues`**: New function to find ALL config objects in arrays instead of just the first one
- **Updated processing logic**: Now processes all found config objects instead of just the first match

## Changes Made
- Modified `processConfigParams` to handle arrays of config objects
- Added `findAllConfigValues` helper function for comprehensive array traversal
- Updated the config processing loop to iterate through all found config values

## Testing
- Tested with `agentTools` array containing multiple `requestsGet` tools
- Verified that variables like `{{ $flow.state.test }}` are now resolved in all array items
- Confirmed backward compatibility with single config objects

## Example
Before: Only first tool's `requestsGetUrl` had variables resolved
```json
"agentTools": [
  {
    "agentSelectedToolConfig": {
      "requestsGetUrl": "http://localhost:5566/events?resolved_value" // ✅ Resolved
    }
  },
  {
    "agentSelectedToolConfig": {
      "requestsGetUrl": "http://localhost:5566/events?{{ $flow.state.test }}" // ❌ Not resolved
    }
  }
]
```

After: All tools' configs have variables resolved
```json
"agentTools": [
  {
    "agentSelectedToolConfig": {
      "requestsGetUrl": "http://localhost:5566/events?resolved_value" // ✅ Resolved
    }
  },
  {
    "agentSelectedToolConfig": {
      "requestsGetUrl": "http://localhost:5566/events?resolved_value" // ✅ Resolved
    }
  }
]
```

## Impact
- Fixes variable resolution in array-based node configurations
- Improves reliability of agent tools and other array-based parameters
- Maintains backward compatibility with existing single config objects

### DIFF
--- a/packages/server/src/utils/buildAgentflow.ts
+++ b/packages/server/src/utils/buildAgentflow.ts
@@ -489,6 +489,14 @@ export const resolveVariables = async (
                 return
             }
 
+            // Handle arrays of config objects
+            if (Array.isArray(configObj)) {
+                for (const item of configObj) {
+                    await processConfigParams(item, configParamWithAcceptVariables)
+                }
+                return
+            }
+
             for (const [key, value] of Object.entries(configObj)) {
                 // Only resolve variables for parameters that accept them
                 // Example: requestsGetHeaders is in configParamWithAcceptVariables, so resolve "Bearer {{ $vars.TOKEN }}"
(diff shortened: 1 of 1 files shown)


maggi2508:
@HenryHengZJ 
Is this fixed in flowise cloud as well (or) when would it get fixed ?

maggi2508:
Hi @HenryHengZJ  - Can you please suggest when this would be fixed in Flowise Cloud ?

maggi2508:
Can someone confirm when this fix will be applied and available on Flowise Cloud ? We have tested this multiple times and hasn’t been working yet.
It would be very helpful if there are some pointers around when we can see this made available.

PR #5368, opened 2025-10-25
feature/Add DATABASE_SCHEMA support for PostgreSQL

azoz-w:
This PR adds support for the DATABASE_SCHEMA environment variable. This allows users to specify a database schema rather than always defaulting to "public" when using a PostgreSQL database, which is necessary for multi-schema environments.
This also adds config inputs for the vector store and record manager to select the desired schema in PostgreSQL, 

## Summary by CodeRabbit

* **New Features**
  * Added support for custom PostgreSQL database schemas through the DATABASE_SCHEMA environment variable, enabling schema configuration across Docker services and all PostgreSQL-backed components.

* **Chores**
  * Updated Docker and environment configuration files to include the DATABASE_SCHEMA option for PostgreSQL schema customization.


packages/server/src/DataSource.ts
```diff
@@ -63,20 +63,26 @@ export const init = async (): Promise<void> => {
...
                 username: process.env.DATABASE_USER,
                 password: process.env.DATABASE_PASSWORD,
                 database: process.env.DATABASE_NAME,
+                ...(isCustomSchema && { schema: dbSchema }), // set custom schema if provided
                 ssl: getDatabaseSSLFromEnv(),
                 synchronize: false,
                 migrationsRun: false,
                 entities: Object.values(entities),
                 migrations: postgresMigrations,
+                migrationsTableName: 'migrations',
```
HenryHengZJ: is this necessary? can we use the default table name instead of hardcoding `migrations` as the table name

azoz-w: True, im just completing the change as it seems a bit more larger. And ill refactor it better.

Thanks for the comment

packages/server/src/DataSource.ts
```diff
@@ -63,20 +63,26 @@ export const init = async (): Promise<void> => {
             })
             break
         case 'postgres':
+            const dbSchema = process.env.DATABASE_SCHEMA
```
HenryHengZJ: you need to wrap these in curly brackets if using `const`

HenryHengZJ:
also if you can give a couple of screenshots to make sure this works, that'd be great

packages/server/src/DataSource.ts
```diff
@@ -70,13 +70,16 @@ export const init = async (): Promise<void> => {
...
                 ssl: getDatabaseSSLFromEnv(),
                 synchronize: false,
                 migrationsRun: false,
                 entities: Object.values(entities),
                 migrations: postgresMigrations,
                 extra: {
-                    idleTimeoutMillis: 120000
+                    idleTimeoutMillis: 120000,
+                    // set the search_path for the migrations and queries to the desired schema
+                    options: `-c search_path=${process.env.DATABASE_SCHEMA},public`
```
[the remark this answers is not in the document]
HenryHengZJ: this is valid find, please address

HenryHengZJ:
thanks for the changes, do you have screenshots to share? record manager, vector store etc

Issue #5402, opened 2025-11-03
Feature Request: Add Mermaid Renderer Support in Chat Window


snowfluke:
### Feature Description

Enable Mermaid.js rendering in the chat window so that Flowise can visualize charts and diagrams written using Mermaid syntax directly within chat responses.

### Feature Category

UI/UX Improvement

### Problem Statement

Currently, when an LLM or custom node outputs Mermaid syntax (e.g. mermaid graphs) in the chat window, it is displayed as plain text.
This limits the ability to present structured information visually such as flowcharts, sequence diagrams, mind maps, and other relationships that are naturally expressed with Mermaid.

```mermaid
graph TD
    A[Start] --> B{Is it working?}
    B -->|Yes| C[Great!]
    B -->|No| D[Fix it]
```

Instead of rendering as a diagram, the above appears as text.

### Proposed Solution

Integrate Mermaid.js into the chat message renderer so that any Markdown code block with a mermaid language tag is rendered as a live diagram.

Example desired behavior:
When a node or LLM outputs:

```mermaid
graph TD
    A[User Input] --> B[Flowise Node]
    B --> C[LLM Response]
```

…it should render the corresponding diagram visually in the chat.

### Mockups or References

https://mermaid.js.org/

aibysid:
Hi @snowfluke was exploring this project this looks like something that might help me as well. will create a PR for this.

snowfluke:
@aibysid yepp it will definitely help, here im experimenting it as well, but I think I'll leave the PR for the appropriate people as Im not confident in following the coding standard

<img width="734" height="1286" alt="Image" src="https://github.com/user-attachments/assets/e512f917-44d8-4c40-a2e3-9c4020e69889" />

<img width="824" height="1351" alt="Image" src="https://github.com/user-attachments/assets/a4bfcd78-f3bb-4931-9678-1af81dbf4b5a" />

<img width="788" height="1340" alt="Image" src="https://github.com/user-attachments/assets/3ef355b7-7497-4659-972a-c78f60c87f43" />

Maybe we should need to enforce the agent at following the syntax correctly

<img width="778" height="1298" alt="Image" src="https://github.com/user-attachments/assets/48360c1b-3e09-46bd-a9e1-99928f35aeed" />

PR #5410, opened 2025-11-04
feat: Add Mermaid renderer support in chat window (#5402)

aibysid:
- Add mermaid.js dependency to render Mermaid diagrams
- Modify CodeBlock component to detect and render Mermaid code blocks
- Support theme-aware rendering (light/dark mode)
- Add error handling with fallback to syntax highlighting
- Preserve existing copy/download functionality

Sample output:
Screenshot 2025-11-04 at 23 37 09

### DIFF
--- a/packages/ui/src/ui-component/markdown/CodeBlock.jsx
+++ b/packages/ui/src/ui-component/markdown/CodeBlock.jsx
@@ -46,24 +46,28 @@ export const CodeBlock = memo(({ language, chatflowid, isFullWidth, value }) => 
     useEffect(() => {
         if (!isMermaid || !value) return
 
-        mermaid.initialize({
-            startOnLoad: false,
-            theme: theme.palette.mode === 'dark' ? 'dark' : 'default',
-            securityLevel: 'strict'
-        })
+        const timer = setTimeout(() => {
+            mermaid.initialize({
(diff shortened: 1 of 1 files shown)


snowfluke:
+1

packages/ui/src/ui-component/markdown/CodeBlock.jsx
```diff
@@ -29,13 +30,41 @@ const programmingLanguages = {
...
+    const [mermaidSvg, setMermaidSvg] = useState(null)
+    const [mermaidError, setMermaidError] = useState(null)
+    const diagramIdRef = useRef(`mermaid-${Date.now()}-${Math.random().toString(36).substr(2, 9)}`)
     const openPopOver = Boolean(anchorEl)
+    const isMermaid = language?.toLowerCase() === 'mermaid'
+
+    useEffect(() => {
+        if (!isMermaid || !value) return
+
+        mermaid.initialize({
```
HenryHengZJ: this would be an issue, because when streaming, value changes, this will cause multiple initialization

aibysid: Wrapped the rendering logic in a setTimeout with a 300ms delay

PR #5456, opened 2025-11-08
fix: return proper status codes for invalid/non-existent chatflow IDs

### LINKED ISSUE #5379: Get chatflow by ID endpoint returns 500 when it should return 4xx
### Describe the bug

The chatflow API does not work as expected and as described in the [documentation](https://docs.flowiseai.com/api-reference/chatflows).

When getting chatflow by ID, with an invalid id, or a not existing id, instead of receiving a 400 or a 404 status code, status code is 500 with the following response 

`{
	"statusCode": 500,
	"success": false,
	"message": "Error: chatflowsService.getChatflowById - Chatflow e not found in the database!",
	"stack": {}
}`

### To Reproduce

Getting chatflow by ID, with an invalid id
1. Use chatflow - Get chatflow by ID endpoint : .../chatflows/invalid-id
2. Response is an error with statusCode 500 instead of 400

Getting chatflow by ID, with a not existing id
1. Use chatflow - Get chatflow by ID endpoint : .../chatflows/dafb70a2-c258-4a36-ae81-2a401c7cd147 (where no chatflow has this id)
2. Response is an error with statusCode 500 instead of 404
 

### Expected behavior

Getting chatflow by ID, with an invalid id
- Expected statusCode is 400

Getting chatflow by ID, with a not existing id
- Expected statusCode is 404

### Screenshots

_No response_

### Flow

_No response_

### Use Method

None

### Flowise Version

_No response_

### Operating System

None

### Browser

None

### Additional context

_No response_

aibysid:
- Add UUID validation to return 400 for invalid ID format
- Preserve InternalFlowiseError status codes (404 for not found)
- Fix error handling to prevent wrapping all errors as 500

Fixes #5379


salimzdn:
I would add that this error also happens for many APIs and not only for the chatflow API, a more global fix could be applied

aibysid:
> I would add that this error also happens for many APIs and not only for the chatflow API, a more global fix could be applied

Agreed will plan on doing it with next commit.

HenryHengZJ (approved):

Issue #5256, opened 2025-09-25
Agent Node - Add structured JSON response (similar to LLM Node)


GG3T:
### Feature Description

Currently, the Agent Node does not support structured JSON output, unlike the LLM Node.
This makes it harder to work with deterministic outputs for APIs, webhooks, or database integrations, forcing the use of workarounds such as regex, extra nodes, or manual post-processing.

Add an option in the Agent Node for structured JSON responses, with support for schema definition (similar to the LLM Node), allowing:

Define a JSON Schema.

Automatically validate the output.

Expose the parsed JSON in the node’s output.

Benefits:

Easier integrations.

Less need for post-processing.

More reliable and consistent agent responses.

### Feature Category

New Node/Component

aibysid:
Created a PR for this issue : https://github.com/FlowiseAI/Flowise/pull/5470

GG3T:
thanks

anteyk:
Hello, 
This is awesome must have feature, many thanks to @aibysid!!
Does anybody know - are there any chances that it will be merged shortly?

PR #5470, opened 2025-11-13
feat: Add structured JSON output support to Agent Node

aibysid:
- Add agentStructuredOutput input parameter matching LLM Node structure
- Implement configureStructuredOutput method to convert schema to Zod
- Add createZodSchemaFromJSON helper for complex JSON schemas
- Configure structured output before binding tools (required order)
- Disable streaming when structured output is enabled
- Extract structured fields in prepareOutputObject method
- Resolves issue #5256
Screenshot 2025-11-13 at 13 47 08

### DIFF
--- a/packages/ui/src/ui-component/input/suggestionOption.js
+++ b/packages/ui/src/ui-component/input/suggestionOption.js
@@ -112,7 +112,7 @@ export const suggestionOptions = (
                 category: 'Node Outputs'
             })
 
-            const structuredOutputs = nodeData?.inputs?.llmStructuredOutput ?? []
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
I dont think it is working:
image

I also dont think we can reuse the same logic as LLM node, because we cant have structured output and tools at the same time. If we want agent to have structured output, we probably have to do it in 2 stages:
- get the output from agent
- call LLM again to output structured data from the answer following the specified schema

aibysid:
@HenryHengZJ agreed on the two stage part , will try to raise a new PR with this approach.

HenryHengZJ:
@aibysid I've modified it now:
image

PR #5473, opened 2025-11-13
fix: agentflow list refresh after deletion (Issue #5360)

### LINKED ISSUE #5360: Deleting an Agentflow from the List view won't show the updates List view
### Describe the bug

When deleting an AgentFlow from the list view using the options dropdown, the list does not refresh correctly.
Instead of showing the updated list, it appears empty until the page is manually reloaded.

### To Reproduce

1. Go to the Agentflows page
2. Select the list view
3. Try to delete an Agentflow
4. Observe, that the Agentflow list is not shown anymore, and a javascript error can be seen:

```
index.jsx:167 TypeError: Cannot read properties of undefined (reading 'length')
```

### Expected behavior

After deleting an AgentFlow, the list should automatically refresh and display the remaining entries.

### Screenshots
Image

### Flow

_No response_

### Use Method

pnpm start

### Flowise Version

_No response_

### Operating System

None

### Browser

None

### Additional context

It seems that when refreshing `getAllAgentflows.data` in the `useEffect` within `agentflows/index.jsx`, no paging parameters are provided.
As a result, the function returns a plain array instead of the expected object structure { data: [], total: 0 }.

aibysid:
Root Cause:
When deleting an Agentflow from the list view, updateFlowsApi.request() was called without pagination parameters. This caused the backend to return a plain array instead of the expected { data: [], total: 0 } format, leading to a TypeError when accessing .data.length.

Solution:
- Created refreshAgentflows callback that includes current pagination state
- Passed callback through FlowListTable to FlowListMenu
- Updated handleDelete to use callback instead of direct API call
- Maintains backward compatibility for components without callback

Note : This change is backward compatible will not break existing delete functionality cheers !

Fixes #5360

### DIFF
--- a/packages/ui/src/ui-component/button/FlowListMenu.jsx
+++ b/packages/ui/src/ui-component/button/FlowListMenu.jsx
@@ -74,7 +74,7 @@ const StyledMenu = styled((props) => (
     }
 }))
 
-export default function FlowListMenu({ chatflow, isAgentCanvas, isAgentflowV2, setError, updateFlowsApi }) {
+export default function FlowListMenu({ chatflow, isAgentCanvas, isAgentflowV2, setError, updateFlowsApi, currentPage, pageLimit }) {
     const { confirm } = useConfirm()
     const dispatch = useDispatch()
     const updateChatflowApi = useApi(chatflowsApi.updateChatflow)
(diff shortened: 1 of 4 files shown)


HenryHengZJ:
@aibysid thank you so much for all the PR! We'll get these reviewed ASAP!

Issue #5477, opened 2025-11-14
Interpolating variable in the Custom MCP config


mskitroot:
### Describe the bug

When adding a Custom MCP in both AgentFlow and Chatflow, and trying to use variables for the Token, I get 401 on my MCP. Debugging on my MCP shows that the Authorization header is sent without the variable being interpolated.

Being no expert I debuged the server code and found that workspaceId is not sent to CustomMCP and when getVars is called it returns an empty set. See screenshots

### To Reproduce

1. Create a variable i.e. mcptoken
2. Add in Chatflow a CustomMCP. Configure it to get the Bearer Token as per example
``` json 
{
    "url": "http://localhost:3001/sse",
    "headers": {
        "Authorization": "Bearer {{$vars.mcptoken}}"
    }
}
```
3. Check on the MCP server side the value of Authorization Header

### Expected behavior

The variable to be interpolated

## MCP Server Logs
<img width="673" height="82" alt="Image" src="https://github.com/user-attachments/assets/b6c1af85-d1f6-40d7-a758-0ebbe663d532" />

## Snippet from packages/components/src/utils.ts:getVars
<img width="669" height="249" alt="Image" src="https://github.com/user-attachments/assets/be54eb96-3171-4d53-964e-cd8433338f5f" />

## Snippet from packages/components/nodes/tools/MCP/CustomMCP/CustomMCP.ts:loadMethods
<img width="827" height="197" alt="Image" src="https://github.com/user-attachments/assets/8a564c75-6cc4-4e82-9228-b929de60350b" />

## Debuging output
<img width="572" height="135" alt="Image" src="https://github.com/user-attachments/assets/3414c115-8d35-4d00-bf18-000880996d91" />

### Use Method

pnpm start

### Flowise Version

3.0.10

### Operating System

macOS

### Browser

Safari

mskitroot:
Further investigation showed that getSingleNodeAsyncOptions in packages/server/src/services/nodes/index.ts does not add workspaceId in the response.

Adding that line seems to have fixed the issue but I am not sure if that is the correct way to do it.

<img width="1026" height="635" alt="Image" src="https://github.com/user-attachments/assets/f3ba54fa-a7d6-43c0-b2b0-042eb7b6536f" />

Wanyi9988:
> Further investigation showed that getSingleNodeAsyncOptions in packages/server/src/services/nodes/index.ts does not add workspaceId in the response.
> 
[...]

PR #5503, opened 2025-11-20
Add workspaceId in options if not already there.

mskitroot:
Bugfix for [issue](https://github.com/FlowiseAI/Flowise/issues/5477).

WorkspaceId is needed for variable interpolation but it's not part of the options object. The fix adds it if not there.


HenryHengZJ (approved):
thank you!

PR #5505, opened 2025-11-22
fix: correct Lunary tool-error logging

moona3k:
Lunary tool-error logs were using the LLM run tracker, so tool call failures showed under the wrong span. This PR points the tool-error handler to the tool tracker so failures tag the correct span. Every other analytics provider (LangSmith, LangFuse, etc) already follows this pattern.


HenryHengZJ:
thank you!

PR #5506, opened 2025-11-24
Standardize to single non-root Dockerfile

chungyau97:
## Description
This PR will replace https://github.com/FlowiseAI/Flowise/pull/5495.
Standardizes the project to use a single non-root Dockerfile instead of maintaining multiple Dockerfiles.

## Manual Data Migration from `~/.flowise` to `flowise_data`
```shell
# Create temporary container with the volume
docker create --name temp-flowise -v docker_flowise_data:/home/node/.flowise node:20-alpine

# Copy data
Linux/Mac: docker cp ~/.flowise/. temp-flowise:/home/node/.flowise/
Win: docker cp C:\Users\[YOUR_COMPUTER_NAME]\.flowise\. temp-flowise:/home/node/.flowise/

# Fix permissions
docker run --rm -v docker_flowise_data:/data node:20-alpine chown -R node:node /data

# Remove temporary container
docker rm temp-flowise
```
## Result

| Initialization with old docker-compose.yml | flowiseaiyau/test:0.0.2 with new docker-compose.yml |
| --- | --- |
| image | image |

| flowiseaiyau/test:0.0.2 with new docker-compose-queue-prebuilt.yml Conversation Chain |  flowiseaiyau/test:0.0.2 with new docker-compose-queue-prebuilt.yml Conversational Retrieval QA Chain Upsert |  flowiseaiyau/test:0.0.2 with new docker-compose-queue-prebuilt.yml Conversational Retrieval QA Chain Chat|
| --- | --- | --- |
| image | image | image |

| Conversation Chain | Conversational Retrieval QA Chain Upsert | Conversational Retrieval QA Chain Chat |
| --- | --- | --- |
| image | image | image |

### DIFF
--- a/docker/docker-compose-queue-prebuilt.yml
+++ b/docker/docker-compose-queue-prebuilt.yml
@@ -13,17 +13,15 @@ services:
         restart: always
 
     flowise:
-        image: flowiseai/flowise:latest
+        image: flowiseaiyau/test:0.0.2
         container_name: flowise-main
         restart: always
         ports:
             - '${PORT:-3000}:${PORT:-3000}'
-        volumes:
-            - ~/.flowise:/root/.flowise
         environment:
             # --- Essential Flowise Vars ---
             - PORT=${PORT:-3000}
-            - DATABASE_PATH=${DATABASE_PATH:-/root/.flowise}
+            - DATABASE_PATH=${DATABASE_PATH}
             - DATABASE_TYPE=${DATABASE_TYPE}
             - DATABASE_PORT=${DATABASE_PORT}
             - DATABASE_HOST=${DATABASE_HOST}
@@ -154,22 +152,22 @@ services:
             timeout: 5s
             retries: 5
             start_period: 30s
-        entrypoint: /bin/sh -c "sleep 3; flowise start"
+        volumes:
+            - flowise_data:/home/node/.flowise
+        entrypoint: /bin/sh -c "sleep 3; pnpm start"
         depends_on:
             - redis
         networks:
             - flowise-net
 
     flowise-worker:
-        image: flowiseai/flowise-worker:latest
+        image: flowiseaiyau/test:0.0.2
         container_name: flowise-worker
         restart: always
-        volumes:
-            - ~/.flowise:/root/.flowise
         environment:
             # --- Essential Flowise Vars ---
             - WORKER_PORT=${WORKER_PORT:-5566}
-            - DATABASE_PATH=${DATABASE_PATH:-/root/.flowise}
+            - DATABASE_PATH=${DATABASE_PATH}
             - DATABASE_TYPE=${DATABASE_TYPE}
             - DATABASE_PORT=${DATABASE_PORT}
             - DATABASE_HOST=${DATABASE_HOST}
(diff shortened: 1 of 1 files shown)


docker/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: '3.1'
 
 services:
     flowise:
-        image: flowiseai/flowise:latest
+        image: flowiseaiyau/test:0.0.2
```
chungyau97, line 5: Remember to change this back to `flowiseai/flowise:latest`.

docker/docker-compose-queue-prebuilt.yml
```diff
@@ -13,17 +13,15 @@ services:
         restart: always
 
     flowise:
-        image: flowiseai/flowise:latest
+        image: flowiseaiyau/test:0.0.2
```
chungyau97, line 16: Remember to change this back to `flowiseai/flowise:latest`.

docker/docker-compose-queue-prebuilt.yml
```diff
@@ -154,22 +152,22 @@ services:
...
+            - flowise_data:/home/node/.flowise
+        entrypoint: /bin/sh -c "sleep 3; pnpm start"
         depends_on:
             - redis
         networks:
             - flowise-net
 
     flowise-worker:
-        image: flowiseai/flowise-worker:latest
+        image: flowiseaiyau/test:0.0.2
```
chungyau97, line 164: Remember to change this back to `flowiseai/flowise:latest`.

HenryHengZJ:
Here's another idea:

1. Container starts as root (needed for migration)
2. Entrypoint script checks if `/root/.flowise` has data:
```
#!/bin/sh
set -e

OLD_DIR="/root/.flowise"
NEW_DIR="/home/node/.flowise"

# If old directory exists and has content, and new directory doesn't exist or is empty
if [ -d "$OLD_DIR" ] && [ "$(ls -A $OLD_DIR 2>/dev/null)" ] && [ ! -d "$NEW_DIR" -o ! "$(ls -A $NEW_DIR 2>/dev/null)" ]; then
    echo "Migrating data from $OLD_DIR to $NEW_DIR..."
    
    # Create new directory if it doesn't exist
    mkdir -p "$NEW_DIR"
    
    # Copy all data from old to new location
    cp -rp "$OLD_DIR"/* "$NEW_DIR"/ 2>/dev/null || true
    
    # Fix ownership for node user
    chown -R node:node "$NEW_DIR"
    
    echo "Migration complete!"
else
    # Ensure the new directory exists with correct permissions
    mkdir -p "$NEW_DIR"
    chown -R node:node "$NEW_DIR"
fi

# Switch to node user and run flowise
exec su-exec node flowise start "$@"
```
4. If yes, it copies data to `/home/node/.flowise` and fixes permissions
5. Script then uses `su-exec` to switch to node user before running flowise
6. Application runs as non-root node user with access to migrated data

However, if we were to simplify to using 1 root `Dockerfile`, we have to consider the scenario where majority of the users are still using the old `docker-compose.yml` with the entrypoint: `/bin/sh -c "sleep 3; flowise start`. This will break since the root `Dockerfile` is now using `pnpm start` not `flowise start`.

We need to find a way that has backward compatibility
- Simplifying to 1 Dockerfile
- Running as **non-root user**
- Automatic migration from old to new location
- Make sure it doesnt break even if user is still using the same `docker-compose.yml`

Extra: We should also look at `Dockerfile` at `docker/worker` directory.

TLDR: a much bigger effort is needed to revamp the strategy of Docker implementation, making sure it is backward compatible, and battle tested on major cloud providers. This is a huge change with big impact.

Note: `/root/.flowise` can be configured to other path, so we cant hardcode that into the entrypoint

PR #5558, opened 2025-12-09
Fix/variable not read in custom loader

Nek-11:
## Problem

After PR #4891 ("Bugfix/update workspaceId to vars"), the `getVars()` function in `packages/components/src/utils.ts` was updated to require `workspaceId` - returning an empty array if not provided:

```
if (!options.workspaceId) {
    return []
}
```

However, `_splitIntoChunks()` in `packages/server/src/services/documentstore/index.ts` was not updated to pass `workspaceId` to the document loader's `init()` options. This caused `$vars` to be empty when using Custom Document Loader during the "Process" step, resulting in undefined API keys and authentication failures.

The "Execute" button worked because it uses a different code path (`executeCustomNodeFunction`) that correctly passes `workspaceId`.

## Solution

- Added `workspaceId` parameter to `_splitIntoChunks()` function
- Updated `previewChunks()` to destructure `workspaceId` from `IExecutePreviewLoader` params
- Passed `workspaceId` to `_splitIntoChunks()` call
- Included `workspaceId` in the options object passed to document loader's `init()` method

## Testing

Tested on production deployed Flowise app:
- Custom Document Loader with `$vars` for API keys (FireCrawl, Mistral, Azure OpenAI)
- Verified "Execute" button works ✅
- Verified "Process" button now works ✅ (previously failed with 401 Unauthorized)
- Variables are correctly resolved in both code paths

### DIFF
--- a/packages/server/src/controllers/documentstore/index.ts
+++ b/packages/server/src/controllers/documentstore/index.ts
@@ -514,7 +514,11 @@ const deleteVectorStoreFromStore = async (req: Request, res: Response, next: Nex
                 `Error: documentStoreController.deleteVectorStoreFromStore - workspaceId not provided!`
             )
         }
-        const apiResponse = await documentStoreService.deleteVectorStoreFromStore(req.params.storeId, workspaceId)
+        const apiResponse = await documentStoreService.deleteVectorStoreFromStore(
+            req.params.storeId,
+            workspaceId,
+            (req.query.docId as string) || undefined
+        )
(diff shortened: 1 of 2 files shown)


Nek-11:
@gemini-code-assist check again now I merged from main

HenryHengZJ:
thank you!

danieldabate:
Thank you so much for the fix. I was not updating to >=3.0.5 because of this issue.

I can't wait to get the new version released.

PR #5590, opened 2025-12-16
Feature/Add Permissions To API Keys

yau-wd:
# Description
## API Key
1. Migration scripts to add `permissions` column into `apikey` table.
2. API key permissions exclude `workspace` and `admin` categories.
3. Cloud feature-based filtering for API keys excludes permissions associated with disabled feature flags.
4. User-level filtering for API keys restricts non-admin users to only their explicitly granted permissions.
5. Deprecate import API keys functionality.

## Permissions 
1. Permissions remove `apikeys import`.
2. Cloud role permissions exclude `sso`.
3. Migration scripts to clean up role permissions values.

---
# API Key

### Before
| SQLite | PostgreSQL | MySQL | MariaDB |
| --- | --- | --- | --- |
| image | image | image | image |

### After
| SQLite | PostgreSQL | MySQL | MariaDB |
| --- | --- | --- | --- |
| image | image | image | image |

### Admin
| Open Source | Enterprise | Cloud Free | Cloud Starter | Cloud Pro |
| --- | --- | --- | --- | --- |
| image | image | image | image | image |

| Open Source | Enterprise | Cloud Free | Cloud Starter | Cloud Pro |
| --- | --- | --- | --- | --- |
| NA | image | NA | NA | image |

### Admin
| Open Source | Enterprise | Cloud Free | Cloud Starter | Cloud Pro |
| --- | --- | --- | --- | --- |
| image | image | NA | NA | image |

| Open Source | Enterprise | Cloud Free | Cloud Starter | Cloud Pro |
| --- | --- | --- | --- | --- |
| NA | image | NA | NA | image |

### View
| High Permissions| Low Permissions |
| --- | --- |
| image | image |

| Cannot create API key with workspace or admin permissions | Cannot create API key with permissions that exceed your own permissions | Pass |
| --- | --- | --- |
| image | image | image |

| Cannot update API key with workspace or admin permissions | Cannot update API key with permissions that exceed your own permissions | Pass |
| --- | --- | --- |
| image | image | image |

### Delete
| API key without delete permissions | API key with delete permissions |
| --- | --- |
| image | image |

---
# Role

## Database
```sql
-- SQLITE | POSTGRESQL
SELECT  COUNT(*) FROM "role" r WHERE r.permissions LIKE '%sso:manage%' OR r.permissions LIKE '%apikeys:import%'

-- MYSQL | MARIADB
SELECT COUNT(*) FROM `role` r WHERE r.permissions LIKE '%sso:manage%' OR r.permissions LIKE '%apikeys:import%'
```
### Before
| SQLite | PostgreSQL | MySQL | MariaDB |
| --- | --- | --- | --- |
| image | image | image | image |

### After
| SQLite | PostgreSQL | MySQL | MariaDB |
| --- | --- | --- | --- |
| image | image | image | image |

### Admin
| Open Source | Enterprise | Cloud Free | Cloud Starter | Cloud Pro |
| --- | --- | --- | --- | --- |
| NA | image | NA | NA | image |

| Open Source | Enterprise | Cloud Free | Cloud Starter | Cloud Pro |
| --- | --- | --- | --- | --- |
| NA | image | NA | NA | image |

### Admin
| Open Source | Enterprise | Cloud Free | Cloud Starter | Cloud Pro |
| --- | --- | --- | --- | --- |
| NA | image | NA | NA | image |

| Open Source | Enterprise | Cloud Free | Cloud Starter | Cloud Pro |
| --- | --- | --- | --- | --- |
| NA | image | NA | NA | image |

### DIFF
--- a/packages/server/src/services/dataset/index.ts
+++ b/packages/server/src/services/dataset/index.ts
@@ -1,11 +1,11 @@
 import { StatusCodes } from 'http-status-codes'
-import { getRunningExpressApp } from '../../utils/getRunningExpressApp'
-import { InternalFlowiseError } from '../../errors/internalFlowiseError'
-import { getErrorMessage } from '../../errors/utils'
-import { Dataset } from '../../database/entities/Dataset'
-import { DatasetRow } from '../../database/entities/DatasetRow'
 import { Readable } from 'stream'
 import { In } from 'typeorm'
+import { Dataset } from '../../database/entities/Dataset'
+import { DatasetRow } from '../../database/entities/DatasetRow'
+import { InternalFlowiseError } from '../../errors/internalFlowiseError'
+import { getErrorMessage } from '../../errors/utils'
+import { getRunningExpressApp } from '../../utils/getRunningExpressApp'
 
 import csv from 'csv-parser'
 
@@ -50,6 +50,7 @@ const getDataset = async (id: string, workspaceId: string, page: number = -1, li
             id: id,
             workspaceId: workspaceId
         })
+        if (!dataset) throw new InternalFlowiseError(StatusCodes.NOT_FOUND, `Dataset ${id} not found`)
         const queryBuilder = appServer.AppDataSource.getRepository(DatasetRow).createQueryBuilder('dsr').orderBy('dsr.sequenceNo', 'ASC')
         queryBuilder.andWhere('dsr.datasetId = :datasetId', { datasetId: id })
         if (page > 0 && limit > 0) {


packages/server/src/database/migrations/postgres/1765360298674-AddApiKeyPermission.ts
```diff
@@ -0,0 +1,21 @@
...
+
+            await queryRunner.query(`UPDATE "${tableName}" SET "${columnName}" = '${permission}';`)
```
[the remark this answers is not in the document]
yau-wd: False Positive

packages/server/src/database/migrations/postgres/1765360298674-AddApiKeyPermission.ts
```diff
@@ -0,0 +1,21 @@
...
...tools:export","assistants:view","assistants:create","assistants:update","assistants:delete","credentials:view","credentials:create","credentials:update","credentials:delete","credentials:share","variables:view","variables:create","variables:update","variables:delete","apikeys:view","apikeys:create","apikeys:update","apikeys:delete","apikeys:import","documentStores:view","documentStores:create","documentStores:update","documentStores:delete","documentStores:add-loader","documentStores:delete-loader","documentStores:preview-process","documentStores:upsert-config","datasets:view","datasets:create","datasets:update","datasets:delete","executions:view","executions:delete","evaluators:view","evaluators:create","evaluators:update","evaluators:delete","evaluations:view","evaluations:create","evaluations:update","evaluations:delete","evaluations:run","templates:marketplace","templates:custom","templates:custom-delete","templates:toolexport","templates:flowexport","templates:custom-share","workspace:view","workspace:create","workspace:update","workspace:add-user","workspace:unlink-user","workspace:delete","workspace:export","workspace:import","users:manage","roles:manage",null,"admin:view"]'
```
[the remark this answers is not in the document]
yau-wd: False Positive

packages/server/src/database/migrations/mariadb/1765360298674-AddApiKeyPermission.ts
```diff
@@ -0,0 +1,21 @@
...
+
+            await queryRunner.query(`UPDATE \`${tableName}\` SET \`${columnName}\` = '${permission}';`)
+        }
+    }
+
+    public async down(): Promise<void> {}
```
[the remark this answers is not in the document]
yau-wd: The `down()` method is intentionally empty. Our rollback procedure requires restoring from database backup and reverting the platform version. Implementing `down()` is redundant and would be bypassed by our backup-restore process anyway.

packages/server/src/services/apikey/index.ts
```diff
@@ -71,42 +120,55 @@ const getApiKeyById = async (apiKeyId: string) => {
...
+    newKey.id = uuidv4()
+    newKey.apiKey = apiKey
+    newKey.apiSecret = apiSecret
+    newKey.keyName = keyName
+    newKey.permissions = permissions
+    newKey.workspaceId = workspaceId
+    const key = appServer.AppDataSource.getRepository(ApiKey).create(newKey)
+    await appServer.AppDataSource.getRepository(ApiKey).save(key)
+    return await getAllApiKeys(userPermissions, isOrganizationAdmin, workspaceId)
 }
```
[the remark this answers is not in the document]
yau-wd: False positive.

The controller already wraps all service calls in try-catch and passes errors to Express middleware via `next(error) `.

packages/server/src/controllers/apikey/index.ts
```diff
@@ -62,8 +88,8 @@ const importKeys = async (req: Request, res: Response, next: NextFunction) => {
         if (!req.user?.activeWorkspaceId) {
             throw new InternalFlowiseError(StatusCodes.PRECONDITION_FAILED, `Workspace ID is required`)
         }
-        req.body.workspaceId = req.user?.activeWorkspaceId
-        const apiResponse = await apikeyService.importKeys(req.body)
+        const user = req.user as LoggedInUser
+        const apiResponse = await apikeyService.importKeys(user.permissions, user.isOrganizationAdmin, req.body)
```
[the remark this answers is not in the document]
yau-wd: Resolved as import apikey functionality has been removed.

packages/server/src/database/migrations/mariadb/1765360298674-AddApiKeyPermission.ts
```diff
@@ -0,0 +1,38 @@
...
+
+        const sso = 'sso:manage'
+        const apikey = 'apikeys:import'
+        const itemsToRemove = [sso, apikey]
+        const roles: Role[] = await queryRunner.query(
+            `SELECT * FROM \`role\` WHERE \`${columnName}\` LIKE '%${sso}%' OR \`${columnName}\` LIKE '%${apikey}%';`
+        )
+        if (roles.length > 0) {
+            for (const role of roles) {
+                let permissions = JSON.parse(role.permissions)
```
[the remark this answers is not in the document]
yau-wd: Done: [fix(migrations): handle corrupted permissions gracefully](https://github.com/FlowiseAI/Flowise/pull/5590/commits/16d996a0e93a9131016c45b52e4c719ec0984cd2)

packages/server/src/index.ts
```diff
@@ -272,14 +259,14 @@ export class App {
...
                             activeOrganizationId: activeOrganizationId,
                             activeOrganizationSubscriptionId: subscriptionId,
                             activeOrganizationCustomerId: customerId,
                             activeOrganizationProductId: productId,
-                            isOrganizationAdmin: true,
-                            activeWorkspaceId: apiKeyWorkSpaceId!,
+                            isOrganizationAdmin: false,
+                            activeWorkspaceId: workspace.id,
                             activeWorkspace: workspace.name
                         }
```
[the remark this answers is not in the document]
yau-wd: Done:
1. [fix(server/src/index.ts): add error handling for API key permissions parsing](https://github.com/FlowiseAI/Flowise/pull/5590/commits/92beda6a6e9707447d77aa2d31c9ddd46a75e818)
2. [Potential fix for code scanning alert no. 83: Clear-text logging of sensitive information](https://github.com/FlowiseAI/Flowise/pull/5590/commits/cc1876f5ec3e415f920b85522727ca65b81990f7)

packages/server/src/index.ts
```diff
@@ -269,17 +257,23 @@
                         const customerId = org.customerId as string
                         const features = await this.identityManager.getFeaturesByPlan(subscriptionId)
                         const productId = await this.identityManager.getProductIdFromSubscription(subscriptionId)
-
+                        let permissions: string[] = []
+                        try {
+                            permissions = JSON.parse(apiKey.permissions)
+                        } catch (error) {
+                            logger.error(`Error parsing permissions for API key ${apiKey.id}:`, error)
```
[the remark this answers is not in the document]
yau-wd: Done: [Potential fix for code scanning alert no. 83: Clear-text logging of sensitive information](https://github.com/FlowiseAI/Flowise/pull/5590/commits/cc1876f5ec3e415f920b85522727ca65b81990f7)

packages/ui/src/views/apikey/APIKeyDialog.jsx
```diff
@@ -45,22 +57,172 @@ const APIKeyDialog = ({ show, dialogProps, onCancel, onConfirm, setError }) => {
...
+            Object.keys(permissionsData).forEach((category) => {
+                if (permissionsData[category].length === 0) {
+                    delete permissionsData[category]
+                }
+            })
+
+            setPermissions(permissionsData)
+
+            if (dialogProps.type === 'EDIT' && dialogProps.key) {
+                const keyPermissions = JSON.parse(dialogProps.key.permissions || '[]')
```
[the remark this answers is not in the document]
yau-wd: Done: [fix(APIKeyDialog.jsx): handle corrupted permissions when editing API key](https://github.com/FlowiseAI/Flowise/pull/5590/commits/61e765d2806b7383bab3fcd31cb159379d16b0cc)

packages/ui/src/views/apikey/index.jsx
```diff
@@ -124,6 +112,29 @@ function APIKeyRow(props) {
...
+                                WebkitLineClamp: '2',
+                                WebkitBoxOrient: 'vertical'
+                            }}
+                        >
+                            {JSON.parse(props.apiKey.permissions || '[]').map((d, key) => (
+                                <React.Fragment key={key}>
+                                    {d}
+                                    {', '}
+                                </React.Fragment>
+                            ))}
```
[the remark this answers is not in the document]
yau-wd: Done: [fix(views/apikey/index.jsx): handle corrupted permissions in API key list](https://github.com/FlowiseAI/Flowise/pull/5590/commits/772eedeaa73ac447a3c4d904939e7f0d89230e22)

packages/server/src/services/apikey/index.ts
```diff
@@ -1,41 +1,143 @@
...
+
+/**
+ * Validates that requested permissions are allowed for API keys
+ * @param user - The logged-in user
+ * @param permissions - JSON string of requested permissions
+ * @param operation - The operation being performed (for error message)
+ * @throws InternalFlowiseError if validation fails
+ */
+function validatePermissions(user: LoggedInUser, permissions: string, operation: string) {
+    const requestedPermissions = JSON.parse(permissions)
```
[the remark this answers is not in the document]
yau-wd: Done: [fix(services/apikey): validate permissions JSON in API key operations](https://github.com/FlowiseAI/Flowise/pull/5590/commits/84835d72c361005e7607b2c491f42ae20cad19bd)

packages/server/src/database/migrations/mariadb/1765360298674-AddApiKeyPermission.ts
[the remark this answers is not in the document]
jocelynlin-wd: can you elaborate more on how rollback works? Are you referring to the rollback on cloud instance or is there a rollback script that would work for non-enterprise version as well?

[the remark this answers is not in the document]
yau-wd: For Enterprise and Open Source users:
https://docs.flowiseai.com/configuration/databases#backup

For Cloud users:
We will restore the database snapshot from before the migration script was run and revert to the previous version of Flowise.

Regarding the `down` migration:
A down migration isn't feasible here because the TypeScript entity definitions would still reference columns that the down migration removes from the database. This mismatch would cause the service layer to fail when attempting to insert data into non-existent columns.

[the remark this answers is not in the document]
yau-wd: If something goes seriously wrong:
1. Display a maintenance page (we don't currently have one)
2. Implement a hotfix directly into the main branch

packages/server/src/database/migrations/mariadb/1765360298674-AddApiKeyPermission.ts
```diff
@@ -0,0 +1,45 @@
...
+
+            const permission =
+                '["chatflows:view","chatflows:create","chatflows:update","chatflows:duplicate","chatflows:delete","chatflows:export","chatflows:import","chatflows:config","chatflows:domains","agentflows:view","agentflows:create","agentflows:update","agentflows:duplicate","agentflows:delete","agentflows:export","agentflows:import","agentflows:config","agentflows:domains","tools:view","tools:create","tools:update","tools:delete","tools:export","assistants:view","assistants:create","assistants:update","assistants:delete","credentials:view","credentials:create","credentials:update","credentials:delete","variables:view","variables:create","variables:update","variables:delete","apikeys:view","apikeys:create","apikeys:update","apikeys:delete","documentStores:view","documentStores:create","documentStores:update","documentStores:delete","documentStores:add-loader","documentStores:delete-loader","documentStores:preview-process","documentStores:upsert-config","executions:view","executions:delete","templates:marketplace","templates:custom","templates:custom-delete","templates:toolexport","templates:flowexport"]'
```
jocelynlin-wd, line 16: is this a preferred default permissions used across different db platforms? can we define them in the `rbac` module and reference the value here?  I suspect we will want this for first time app setup too?

packages/server/src/database/entities/ApiKey.ts
```diff
@@ -15,6 +14,9 @@ export class ApiKey implements IApiKey {
     @Column({ type: 'text' })
     keyName: string
 
+    @Column({ nullable: false, type: 'text', default: '[]' })
```
jocelynlin-wd: question: do we use seed or something similar for new app installation to include the default permissions added in the migration scripts?

jocelynlin-wd: also is it possible to use `simple-json` instead?
```
@Column({
    type: 'simple-json', 
    nullable: false, 
    default: '[]' 
})
permissions: string[]; // TypeORM handles the stringify/parse automatically
```
I am under the impression that TypeORM can automatically pick the best column type for the underlying DB (it will use JSONB in Postgres, JSON in MySQL, and TEXT in SQLite), would you be able to help verify?

This would allow us to mitigate the performance and maintenance bottleneck especially if in the future we need to scale to combine JWT scopes with other services

packages/server/src/database/migrations/mariadb/1765360298674-AddApiKeyPermission.ts
yau-wd: Is this a preferred default permissions used across different db platforms? can we define them in the rbac module and reference the value here?
Nope, why:
1. This is only for backward compatibility, this migration is a one-time backfill for API keys created before the permissions column existed.
2. There are no "default permissions" for new API keys, users must explicitly select permissions at creation time.
3. Migration scripts cannot be changed once run in production. They're timestamped snapshots that must remain stable even if:
   - The Permissions class evolves
   - New permissions are added
   - The permission system changes entirely

    That's why the permission list is hardcoded rather than referencing the RBAC module external dependencies could break old migrations.
    
I suspect we will want this for first time app setup too?
First-time app installations won't have any API Key records, so this migration won't affect them. API keys will be created through the normal flow where users explicitly select permissions.

packages/server/src/database/entities/ApiKey.ts
yau-wd: do we use seed or something similar for new app installation to include the default permissions added in the migration scripts?

No. The `default: '[]'` is required because the column is `NOT NULL`, without it, adding the column would fail on tables with existing data. The migration only backfills existing API keys with full permissions for backward compatibility.

New installations start with no API keys, and when users create them, they explicitly select permissions.

yau-wd: I'm not sure, but this PR is to address Jira2 Ticket Flowise-12 quickly. Migrating to `simple-json` could be a separate initiative if this becomes a bottleneck.

Priority is testing and ensuring no unintended side effects in other parts of the codebase.

packages/server/src/database/migrations/mariadb/1765360298674-AddApiKeyPermission.ts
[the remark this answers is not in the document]
jocelynlin-wd: thanks for explaining. Would ORM and the app continue working if the db has extra column while the code changes is rolled back (the entity definition does not have the permissions)?

packages/server/src/database/migrations/mariadb/1765360298674-AddApiKeyPermission.ts
jocelynlin-wd: sounds good, thanks for the additional context

packages/server/src/database/entities/ApiKey.ts
jocelynlin-wd: depends on how much effort it would take to try out `simple-json` since changing the data type later on will be much more expensive (and the type was a recommendation by gemini for scalability and to avoid tech debt)

packages/server/src/database/migrations/mariadb/1765360298674-AddApiKeyPermission.ts
yau-wd: Would ORM and the app continue working if the db has extra column while the code changes is rolled back (the entity definition does not have the permissions)?

Yup, it will work:
| Rollback Flowise Version only | Upgrade Flowise Version again |
| --- | --- |
| image | image |

packages/server/src/database/entities/ApiKey.ts
yau-wd: ## Test PostgreSQL 
### 1. Initial attempt with LIKE (fails on PostgreSQL)
```ts
                    const apiKeysWithChatflowsView = await this.AppDataSource.getRepository(ApiKey)
                        .createQueryBuilder('apikey')
                        .where('apikey.permissions LIKE :permission', { permission: '%chatflows:view%' })
                        .getMany()
                    // Log or use the filtered API keys as needed
                    logger.info(`Found ${apiKeysWithChatflowsView.length} API keys with chatflows:view permission`)
```
**Error:**          
```sh
query failed: SELECT "apikey"."id" AS "apikey_id", "apikey"."apiKey" AS "apikey_apiKey", "apikey"."apiSecret" AS "apikey_apiSecret", "apikey"."keyName" AS "apikey_keyName", "apikey"."permissions" AS "apikey_permissions", "apikey"."updatedDate" AS "apikey_updatedDate", "apikey"."workspaceId" AS "apikey_workspaceId" FROM "apikey" "apikey" WHERE "apikey"."permissions" LIKE $1 -- PARAMETERS: ["%chatflows:view%"]
error: error: operator does not exist: jsonb ~~ unknown
```
### 2. Fixed with PostgreSQL JSONB operator (works)
```ts
                    const apiKeysWithChatflowsView = await this.AppDataSource.getRepository(ApiKey)
                        .createQueryBuilder('apikey')
                        .where('apikey.permissions::jsonb @> :permission', { permission: '"chatflows:view"' })
                        .getMany()
                    // Log or use the filtered API keys as needed
                    logger.info(`Found ${apiKeysWithChatflowsView.length} API keys with chatflows:view permission`)
```
**Success:**
```sh
2026-01-06 20:55:44 [INFO]: Found 1 API keys with chatflows:view permission
``` 

## My opinion
Without maintaining multiple queries, we'd have to fall back to loading all records and filtering in-memory, which defeats the purpose of database-level optimization, or using `TEXT` across different database platforms.

SQLite does not support the `jsonb` type and PostgreSQL's JSON operators (`?`, `@>`, etc.).

**Database-specific solutions:**
1. **PostgreSQL**: Use `jsonb` operators (`?`, `@>`)
2. **SQLite**: Use `LIKE` with text pattern matching
3. **MySQL & MariaDB**: Use `JSON_CONTAINS()`

yau-wd: Done: [feat(migrations): change ApiKey.permissions column from TEXT to JSON](https://github.com/FlowiseAI/Flowise/pull/5590/commits/7ee29cf107c45b9e04c4b6e9700832f4509c4e33)

packages/server/src/services/apikey/index.ts
```diff
@@ -1,41 +1,130 @@
...
-        return keysWithChatflows
+    // User permission validation - only applies to non-admins (authorization check)
+    if (!user.isOrganizationAdmin) {
+        // Check if all requested permissions are included in user permissions
+        const hasInvalidPermissions = requestedPermissions.some(
+            (permission: string) => permission !== null && !user.permissions.includes(permission)
+        )
+        if (hasInvalidPermissions) {
+            throw new InternalFlowiseError(
+                StatusCodes.BAD_REQUEST,
```
jocelynlin-wd: the message indicates that there is a permission issue but we are using `BAD_REQUEST` (vs `FORBIDDEN`) status? is this intentional?

packages/server/src/services/apikey/index.ts
```diff
@@ -1,41 +1,130 @@
...
+    // API Keys should not have workspace or admin permissions
+    // This applies to ALL users, including admins (platform constraint)
+    const hasRestrictedPermissions = requestedPermissions.some((permission: string) => {
+        if (permission === null) return false
+        return permission.startsWith('workspace:') || permission.startsWith('admin:')
+    })
 
-const getAllApiKeysFromDB = async (workspaceId: string, page: number = -1, limit: number = -1) => {
+    if (hasRestrictedPermissions) {
+        throw new InternalFlowiseError(StatusCodes.BAD_REQUEST, `Cannot ${operation} API key with workspace or admin permissions`)
```
jocelynlin-wd: do we want to use the status FORBIDDEN instead? or the reason we do BAD_REQUEST is because this validation function is actually meant for update operations?

packages/server/src/services/apikey/index.ts
```diff
@@ -1,41 +1,130 @@
...
-import { ApiKey } from '../../database/entities/ApiKey'
-import { Not, IsNull } from 'typeorm'
-import { getWorkspaceSearchOptions } from '../../enterprise/utils/ControllerServiceUtils'
-import { v4 as uuidv4 } from 'uuid'
+import logger from '../../utils/logger'
+
+/**
+ * Validates that requested permissions are allowed for API keys
+ * @param user - The logged-in user
+ * @param permissions - JSON string of requested permissions
```
jocelynlin-wd:
```suggestion
 * @param permissions - string array of requested permissions
```

packages/server/src/services/apikey/index.ts
```diff
@@ -1,41 +1,130 @@
...
+ * @param operation - The operation being performed (for error message)
+ * @throws InternalFlowiseError if validation fails
+ */
+function validatePermissions(user: LoggedInUser, requestedPermissions: string[], operation: string) {
```
jocelynlin-wd, line 20: [minor] do we want to limit operation with a type or enum?

yau-wd: Probably in the next phase, as we also need to take Stripe product metadata into consideration so everything can be standardized and we can determine whether a type, enum, or other declaration or method is most appropriate.

tianwei-liu (approved):

yau-wd:
Edge Case: Pro plan users downgrading while retaining API keys with sharing permissions

Done:  [feat(PricingDialog.jsx): handle downgrade with excessive apikey permissions](https://github.com/FlowiseAI/Flowise/pull/5590/commits/02ecb3d0511701ba667321e06f561d7e19bdebf5)

## Test
### UI
> organizationId: c34ee07a-7d67-41bc-a2ad-dd4e68ec2b09

| Cannot Downgrade | Can Downgrade |
| --- | --- |
| image image | image image image |

### Database
> organizationId: c34ee07a-7d67-41bc-a2ad-dd4e68ec2b09

| Before | After |
| --- | --- |
| image | image |

HenryHengZJ (approved):

tianwei-liu (approved):

packages/server/src/services/apikey/index.ts
```diff
@@ -1,41 +1,145 @@
...
+ * @param user - The logged-in user
+ * @param permissions - string array of requested permissions
+ * @param operation - The operation being performed (for error message)
+ * @throws InternalFlowiseError if validation fails
+ */
+function validatePermissions(user: LoggedInUser, requestedPermissions: string[], operation: string) {
+    // API Keys should not have workspace or admin permissions
+    // This applies to ALL users, including admins (platform constraint)
+    const hasRestrictedPermissions = requestedPermissions.some((permission: string) => {
+        if (permission === null) return false
```
taraka-vishnumolakala: @chungyau97 Why are we allowing `null` values?

packages/server/src/controllers/apikey/index.ts
```diff
@@ -24,10 +26,14 @@ const createApiKey = async (req: Request, res: Response, next: NextFunction) =>
         if (typeof req.body === 'undefined' || !req.body.keyName) {
             throw new InternalFlowiseError(StatusCodes.PRECONDITION_FAILED, `Error: apikeyController.createApiKey - keyName not provided!`)
         }
-        if (!req.user?.activeWorkspaceId) {
-            throw new InternalFlowiseError(StatusCodes.PRECONDITION_FAILED, `Workspace ID is required`)
+        if (!req.body.permissions || !Array.isArray(req.body.permissions)) {
```
taraka-vishnumolakala: This doesn't verify that all array elements are strings. For example, {"permissions": [12345, true, false, null, "logs:read"]} might still be allowed. Strong type validation is needed. I would recommend adding test cases to validate the same.

packages/server/src/services/apikey/index.ts
yau-wd: Because previously: `["chatflows:read",null,null,null,"apikeys:read"]`

Done: [chore(services/apikey): remove redundant null checks](https://github.com/FlowiseAI/Flowise/pull/5590/commits/8968f3fa50fad66c171f65d91318216d1bf76f3b)

This [feat(controllers\apikey): enforce permissions validation](https://github.com/FlowiseAI/Flowise/pull/5590/commits/72bdb46c7a6392c35076a811fc633e9b7248ee38) will ensure no more `null`.

packages/server/src/controllers/apikey/index.ts
yau-wd: Done: [feat(controllers\apikey): enforce permissions validation](https://github.com/FlowiseAI/Flowise/pull/5590/commits/72bdb46c7a6392c35076a811fc633e9b7248ee38)

# Test
## Create
| emty array | num array | bool array | null array | mix array |
| --- | --- | --- | --- | --- |
| image | image | image | image | image |

## Update
| emty array | num array | bool array | null array | mix array |
| --- | --- | --- | --- | --- |
| image | image | image | image | image |

yau-wd:
I will not be able to test due to other high priority items. - @taraka-vishnumolakala 
I have finished testing this, not seeing any issues with this particular PR - @igor-magun-wd

PR #5599, opened 2025-12-18
fix: allow Unicode characters in document store tool names

### LINKED ISSUE #5598: Agent execution fails when Document Store title is in Korean – Invalid 'tools[0].function.name': empty string
### Describe the bug

When a Document Store with a Korean (non-ASCII) title is upserted and then connected to an Agentflow as a Tool, the Agent fails immediately on execution with the following error:
`Error in Agent node:
400 Invalid 'tools[0].function.name': empty string.
Expected a string with minimum length 1, but got an empty string instead.`

### To Reproduce

Create a Document Store

Upload documents and set the Document Store title in Korean

Generate embeddings and confirm Vector Store creation

Create an Agentflow

Connect the Document Store to the Agent as a Tool

Run the Agent and enter any query

Observe the error:

### Expected behavior

The Agent should execute normally regardless of the language used in the Document Store title

The Document Store Tool should be properly registered with a valid function name

The Agent should retrieve documents and generate a response

### Screenshots

Image

Image

Image

### Flow

_No response_

### Use Method

None

### Flowise Version

_No response_

### Operating System

None

### Browser

None

### Additional context

_No response_

doubleclip118:
This PR fixes the 'Invalid tools[0].function.name: empty string' error that occurs with non-ASCII document store titles. Instead of allowing Unicode characters (which would cause LLM API errors), it sanitizes tool names to ASCII-only and generates unique fallback names for non-ASCII titles.

Fixes #5598

### DIFF
--- a/packages/components/nodes/agentflow/Agent/Agent.ts
+++ b/packages/components/nodes/agentflow/Agent/Agent.ts
@@ -21,6 +21,7 @@ import { flatten } from 'lodash'
 import zodToJsonSchema from 'zod-to-json-schema'
 import { getErrorMessage } from '../../../src/error'
 import { DataSource } from 'typeorm'
+import { randomBytes } from 'crypto'
 import {
     addImageArtifactsToMessages,
     extractArtifactsFromResponse,
@@ -79,7 +80,7 @@ const sanitizeToolName = (name: string): string => {
 
     // If the result is empty (e.g., non-ASCII only input), generate a unique fallback name
     if (!sanitized) {
-        return `tool_${Date.now()}_${Math.random().toString(36).slice(2, 7)}`
+        return `tool_${Date.now()}_${randomBytes(4).toString('hex').slice(0, 5)}`
     }
 
     // Enforce 64 character limit common for tool names


doubleclip118:
Hi @HenryHengZJ and @vinodkiran,

Thank you for your work on Flowise.

This PR addresses an issue related to the HTTP node in Agentflow v2 where certain responses fail to be handled properly under specific conditions. It includes improved handling for streams and empty response bodies, and adds safety checks to avoid runtime errors in common workflows.

Could you please take a look when you have time?  
I can update this further based on feedback.

Thank you!

packages/components/nodes/agentflow/Agent/Agent.ts
```diff
@@ -1,3 +1,4 @@
+import crypto from 'crypto'
```
tianwei-liu: nit: `import { randomBytes } from 'crypto'`

packages/components/nodes/agentflow/Agent/Agent.ts
```diff
@@ -65,6 +66,27 @@ interface ISimpliefiedTool {
...
+ */
+const sanitizeToolName = (name: string): string => {
+    const sanitized = name
+        .toLowerCase()
+        .replace(/ /g, '_')
+        .replace(/[^a-z0-9_-]/g, '') // ASCII only for LLM API compatibility
+    
+    // If the result is empty (e.g., non-ASCII only input), generate a unique fallback name
+    if (!sanitized) {
+        return `tool_${crypto.randomBytes(8).toString('hex')}`
```
tianwei-liu: I noticed you already used `Math.random()` below. I think it should be enough for the random name.
```return `tool_${Date.now()}_${Math.random().toString(36).slice(2, 7)}`;```
This can avoid importing the crypto library just for generating a random name, since it doesn't require cryptographic security.

It seems that Gemini proposed to use `crypto.randomBytes`.  I'm open to either option.

PR #5628, opened 2026-01-11
Fix/langfuse trace usage cost

vellanki-santhosh:
I TIRED TO FIX THE BUG , PLS REVIEW THE CHANGES

### DIFF
--- a/packages/components/nodes/tools/Composio/Composio.ts
+++ b/packages/components/nodes/tools/Composio/Composio.ts
@@ -63,11 +63,11 @@ class Composio_Tools implements INode {
                 refresh: true
             },
             {
-                label: 'Auth Status',
-                name: 'authStatus',
+                label: 'Connected Account',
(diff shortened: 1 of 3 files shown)


vellanki-santhosh:
Thanks for the review! 🙏  
Good catch on the `usage` check — I’ll update it to use `!== undefined` so 0 tokens are handled correctly.  
Also noted the redundant call in `Agent.ts`; I’ll clean that up. Refactor for `onLLMEnd` / `onLLMError` duplication can be a follow‑up.

vellanki-santhosh:
Thanks for the feedback!  
- Fixed token usage check to handle 0 correctly.  
- Removed redundant call in Agent.ts.  
Refactor of onLLMEnd/onLLMError duplication will be done in a separate PR.  
Ready for re-review.

HenryHengZJ:
do you have screenshots?

PR #5632, opened 2026-01-12
Feat: Add flow versioning system with master-version pattern

AhmedRaafat14:
## Summary

Implements a comprehensive versioning system for ChatFlows and AgentFlows, allowing users to create, manage, and switch between multiple versions of their flows.

### Problem

Currently, I face some problems when I need to update an existing flow without losing my existing one, I faced these problems:
- Manually duplicate the entire flow
- Make changes to the duplicated copy
- Manage multiple separate flows that are conceptually versions of the same flow
- With every new flow, users have to update the client by calling the `predications` endpoint with the new flow ID.

This leads to cluttered flow lists, no clear version history, difficulty rolling back, and risk of deploying the wrong version.

### Solution

Implements a **Master-Version Pattern** that:
- Maintains a stable master flow ID that never changes
- Stores each version as a separate record linked to the master
- Allows users to view, create, and manage versions through the UI
- Supports draft/publish workflow for safe iterative development

I have implemented these changes in my self-hosted version of Flowise and have been using it for some time now, so it would be useful to share with the community and give back.

## Key Features

- **Version History Modal**: View all versions with metadata (timestamp, creator, description)
- **Active Version Management**: Designate which version is used by APIs/predictions
- **Version Indicator**: Shows current version status in canvas header
- **Create From Any Version**: Fork new versions from any existing version
- **Backward Compatible**: Existing flows auto-migrate to v1, all existing API endpoints continue to work
- **Query Specific Version**: Allow to call API like this `api/v1/prediction/{ID}/versions/{VERSION_NUMBER}`, this allows you to test a specific version before publishing it.

### New Tables
- `chat_flow_master` - Stores high-level flow metadata with stable ID
- `chat_flow_version` - Stores individual versions with complete configuration

### Migrations
- Schema migrations for all 4 supported databases (PostgreSQL, MySQL, MariaDB, SQLite)
- Data migrations to convert existing `chat_flow` records to a versioned structure
- Existing flows become version 1 with `isActive = true`.

## API Endpoints

| Method | Endpoint | Description |
|--------|----------|-------------|
| GET | `/api/v1/chatflows/:id/versions` | List all versions |
| GET | `/api/v1/chatflows/:id/versions/:version` | Get specific version |
| POST | `/api/v1/chatflows/:id/versions` | Create new version |
| PUT | `/api/v1/chatflows/:id/versions/:version` | Update version |
| PUT | `/api/v1/chatflows/:id/active-version` | Set active version |
| DELETE | `/api/v1/chatflows/:id/versions/:version` | Delete version |

### Backend (New)
- `packages/server/src/database/entities/ChatFlowMaster.ts`
- `packages/server/src/database/entities/ChatFlowVersion.ts`
- `packages/server/src/services/chatflow-versions/index.ts`
- `packages/server/src/controllers/chatflow-versions/index.ts`
- `packages/server/src/routes/chatflow-versions/index.ts`
- Database migrations for PostgreSQL, MySQL, MariaDB, SQLite

### Frontend (New)
- `packages/ui/src/ui-component/dialog/VersionHistoryDialog.jsx`.

### Modified
- `packages/server/src/Interface.ts` - Added interfaces
- `packages/server/src/routes/index.ts` - Registered routes
- `packages/ui/src/api/chatflows.js` - Added API methods
- `packages/ui/src/views/canvas/index.jsx` - Version state & handlers
- `packages/ui/src/views/canvas/CanvasHeader.jsx` - Version UI components
- `packages/ui/src/views/agentflowsv2/Canvas.jsx` - AgentFlow version support

## Test Plan

- [x] Create new flow → v1 is created automatically
- [x] Publish new version → new version created and set as active
- [x] View version history → all versions displayed with metadata
- [x] Load specific version → canvas updates with that version's data
- [x] Set active version → API uses correct version for predictions
- [x] Delete non-active version → deletion succeeds
- [x] Try delete active version → error message shown
- [x] Try delete last version → error message shown
- [x] Create version from older version → new version created with source tracking
- [x] Existing flows after migration → work as v1 with no breaking changes

## Breaking Changes

None - fully backward compatible. All existing API endpoints and flows continue to work without modification.

## Recording

https://github.com/user-attachments/assets/ccfb0bcb-1175-46da-b5ca-1ca80eaa2747


HenryHengZJ:
thank you so much @AhmedRaafat14 for the PR! we were also having similar work under development, although there is some difference. Main thing being we also wanted to change how API access should be together with publishing and versioning:
- Everry new flow will now opened as draft, and all existing flows by default will be published to avoid disruption.
- Users can save the draft, but will not create a new version
- Once user is happy with the flow, it can be published. This will then create a new version
- Users can then do the same as what you did on this PR, going back and forth between different version, and make a new draft from a specific version
- Only published flows can be called by Prediction API with an API key  or a valid session token

AhmedRaafat14:
Thanks @HenryHengZJ for the feedback, do you want me to try and adapt to these points or just rely on you folks to bring this soon 😃 !

HenryHengZJ:
No, you dont have to. You can still continue on your forked branch with the changes while we'll continue working on the versioning & publishing

PR #5661, opened 2026-01-21
fix: route ordering in auth routes

0xi4o:
- Wrong route order was causing `/api/v1/auth/sso-success`endpoint to be handled by the `getAllPermissions`endpoint.


ilmaya0211:
> ## Code Review
> 
> This pull request correctly addresses a routing precedence issue by reordering the Express.js routes. By placing the more specific `/sso-success` route before the more general `/:type` route, it ensures that requests to `/api/v1/auth/sso-success` are handled by the `ssoSuccess` controller, preventing them from being incorrectly caught by `getAllPermissions`. This is a necessary and effective fix.

PR #5667, opened 2026-01-22
Fix Paramater Override Bypass

christopherholland-workday:
1. Today, if the overrideConfig value included `FILE-STORAGE::` it would bypass validations. This PR removes that check as it was only used for overriding - now it will rely on `isParameterEnabled` like in the other validations
2. Add `NODE_OPTIONS` as a dangerous env var

### Testing
Recreated an issue and ensured it was mitigated locally. Please chat offline with @igor-magun-wd and myself for details.

### DIFF
--- a/packages/ui/src/views/vectorstore/VectorStoreDialog.jsx
+++ b/packages/ui/src/views/vectorstore/VectorStoreDialog.jsx
@@ -578,7 +578,7 @@ formData.append("openAIApiKey[openAIEmbeddings_0]", "sk-my-openai-2nd-key")`
                                                                             <a
                                                                                 rel='noreferrer'
                                                                                 target='_blank'
-                                                                                href='https://docs.flowiseai.com/using-flowise/api#override-config'
+                                                                                href='https://docs.flowiseai.com/using-flowise/prediction#configuration-override'
                                                                             >
                                                                                 here
                                                                             </a>{' '}


packages/server/src/utils/index.ts
```diff
@@ -1189,10 +1189,7 @@ export const replaceInputsWithConfig = (
                     continue
                 }
             } else {
-                // Skip if it is an override "files" input, such as pdfFile, txtFile, etc
-                if (typeof overrideConfig[config] === 'string' && overrideConfig[config].includes('FILE-STORAGE::')) {
-                    // pass
-                } else if (!isParameterEnabled(flowNodeData.label, config)) {
+                if (!isParameterEnabled(flowNodeData.label, config)) {
```
HenryHengZJ, line 1192: Just a note for future reference, this is breaking change for users that are using `/vector/upsert` endpoint.

```
curl http://localhost:3000/api/v1/vector/upsert/d7002e48-8448-4cb6-b09e-c916f92ac889 \
     -X POST \
     -F "files=@/home/user1/Desktop/example.txt" \
     -H "Content-Type: multipart/form-data"
```

Previously you don't have to explicitly enable `files` in the [Override Configuration](https://docs.flowiseai.com/using-flowise/prediction#configuration-override). With this changes, now you do.

PR #5678, opened 2026-01-26
Use secureFetch in WebScraperTool

christopherholland-workday:
This code was using `node-fetch` instead of the custom-built `secureFetch` method. `secureFetch` has SSRF protection built in and should be the standard used in Flowise when making fetch calls.


yau-wd:
The relevant changes are included inPR https://github.com/FlowiseAI/Flowise/pull/5653, specifically in the commit [feat(WebScraperTool.ts): swap fetch to secureFetch](https://github.com/FlowiseAI/Flowise/pull/5653/changes/203462040829eeaba4bf654af9cc4038191d0177)

PR #5680, opened 2026-01-27
fix: remove basic auth

0xi4o:
- Removes all basic auth related code since it's deprecated


igor-magun-wd (approved):

christopherholland-workday (approved):

yau-wd (changes requested):
left comment in slack

yau-wd (approved):

PR #5696, opened 2026-02-02
Deep web crawler document loader

carliblaz:


HenryHengZJ:
thanks! quick question: how is it different from other crawlers we have in the doc loader?

carliblaz:
1. it respect robots.txt
2. removes common parts
3. have setting for deepth
4. have option for concurency

The bigest change is that it removes common parts. Why is this important? Because other crawlers get links from first page, then crawl next page, where they get the same links as on first page, and so one ... every page links are almost the same, links in header.

PR #5657, opened 2026-01-20
Fix: Add regex comparison support to Condition Agentflow

shazee257:
Fixes [#5650 -](https://github.com/FlowiseAI/Flowise/issues/5650) - Bug: "Regex" operation in Condition node is not implemented in backend

The Condition Agentflow node exposes a Regex operation in the UI but the backend did not implement it, causing runtime errors during evaluation. This update adds the missing backend behavior and ensures regex conditions evaluate safely.

- **Backend condition evaluation**
  - Added a `regex` comparator with defensive limits and invalid-pattern handling to prevent runtime errors and mitigate expensive pattern evaluation.

- **Tests**
  - Added unit coverage for regex match, invalid patterns, and safety limit enforcement.

Example:

```ts
const isMatch = compareOperationFunctions.regex('Hello123', '^Hello\\d+$')
// → true
```

> [!WARNING]
>
> <details>
> <summary>Firewall rules blocked me from connecting to one or more addresses (expand for details)</summary>
>
> #### I tried to connect to the following addresses, but was blocked by firewall rules:
>
> - `cdn.sheetjs.com`
>   - Triggering command: `/usr/local/bin/node node /usr/local/bin/pnpm install` (dns block)
>
> If you need me to access, download, or install something from one of these locations, you can either:
>
> - Configure [Actions setup steps](https://gh.io/copilot/actions-setup-steps) to set up my environment, which run before the firewall is enabled
> - Add the appropriate URLs or hosts to the custom allowlist in this repository's [Copilot coding agent settings](https://github.com/shazee257/Flowise/settings/copilot/coding_agent) (admins only)
>
> </details>

<details>

<summary>Original prompt</summary>

> read the issue at https://github.com/FlowiseAI/Flowise/issues/5650 and create new pull request with description , fix it

</details>

---

💡 You can make Copilot smarter by setting up custom instructions, customizing its development environment and configuring Model Context Protocol (MCP) servers. Learn more [Copilot coding agent tips](https://gh.io/copilot-coding-agent-tips) in the docs.

PR #5711, opened 2026-02-05
fix: implement regex operation in Condition node

### LINKED ISSUE #5650: Bug: "Regex" operation in Condition node is not implemented in backend
### Describe the bug

### Error

### Root Cause

### To Reproduce

3. Select Operation: "Regex"
6. Run the flow

### Expected behavior

### Screenshots

_No response_

### Flow

_No response_

### Use Method

None

### Flowise Version

_No response_

### Operating System

None

### Browser

None

### Additional context

_No response_

thecaptain789:
## Description

The 'Regex' operation was available in the frontend dropdown for Condition nodes but was not implemented in the backend's `compareOperationFunctions` map. This caused a runtime error: `compareOperationFunctions[operation] is not a function`.

## Root Cause

- **Lines 265-280**: The `compareOperationFunctions` object did NOT include a `regex` handler

## Solution

Added a `regex` handler to the `compareOperationFunctions` map that:
- Creates a `RegExp` from the value2 pattern
- Tests value1 against the pattern using `pattern.test()`
- Returns `false` gracefully if the regex pattern is invalid (try/catch)

## Changes

```typescript
regex: (value1: CommonType, value2: CommonType) => {
    try {
        const pattern = new RegExp(value2 as string)
        return pattern.test((value1 || '').toString())
    } catch {
        return false
    }
}
```

## Testing

3. Select Operation: "Regex"
4. Set Value 1 to a test string (e.g., "hello123")
5. Set Value 2 to a regex pattern (e.g., "[0-9]+")
6. Run the flow
7. The condition should now evaluate correctly

## Related Issue

Fixes #5650

### DIFF
--- a/packages/components/nodes/agentflow/Condition/Condition.ts
+++ b/packages/components/nodes/agentflow/Condition/Condition.ts
@@ -277,7 +277,7 @@ class Condition_Agentflow implements INode {
             startsWith: (value1: CommonType, value2: CommonType) => (value1 as string).startsWith(value2 as string),
             regex: (value1: CommonType, value2: CommonType) => {
                 try {
-                    const pattern = new RegExp(value2 as string)
+                    const pattern = new RegExp(String(value2 ?? ''))
                     return pattern.test((value1 || '').toString())
                 } catch {
                     return false


HenryHengZJ:
we have a similar [PR](https://github.com/FlowiseAI/Flowise/pull/5657)

HenryHengZJ:

PR #5713, opened 2026-02-06
fix(embeddings): replace deprecated gemini-embedding-exp-03-07 with gemini-embedding-001

adolfousier:
## Summary
- Replace deprecated `gemini-embedding-exp-03-07` model with the new `gemini-embedding-001` in Google Generative AI Embeddings node

## Test plan
- [ ] Verify the new model appears in the GoogleGenerativeAI Embeddings node dropdown
- [ ] Test embedding generation with the new model

### DIFF
--- a/packages/components/models.json
+++ b/packages/components/models.json
@@ -2082,16 +2082,16 @@
             "name": "googleGenerativeAiEmbeddings",
             "models": [
                 {
+                    "label": "gemini-embedding-001",
+                    "name": "gemini-embedding-001"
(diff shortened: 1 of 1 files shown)


adolfousier:
I just noticed that the other two models in the gemini embeddings node are deprecated, not only exp 03-07. 

**The only available now is the gemini-embedding-001** 
@HenryHengZJ

HenryHengZJ (approved):

PR #5717, opened 2026-02-06
fix(embeddings): remove deprecated Google embedding models, keep only gemini-embedding-001

adolfousier:
## Summary
- Remove deprecated `embedding-001` (retired Aug 2025), `text-embedding-004` (shut down Jan 2026), and `gemini-embedding-exp-03-07` (retired Aug 2025) from Google Generative AI Embeddings model list
- Keep only `gemini-embedding-001` which is the sole supported model as of Feb 2026
- Update default model in GoogleGenerativeAIEmbedding node from `embedding-001` to `gemini-embedding-001`

## References
- https://ai.google.dev/gemini-api/docs/embeddings
- https://ai.google.dev/gemini-api/docs/changelog

## Test plan
- [ ] Verify `gemini-embedding-001` appears in the GoogleGenerativeAI Embeddings node dropdown
- [ ] Verify deprecated models no longer appear
- [ ] Test embedding generation with the new default model

### DIFF
--- a/packages/components/models.json
+++ b/packages/components/models.json
@@ -2198,14 +2198,6 @@
                 {
                     "label": "gemini-embedding-001",
                     "name": "gemini-embedding-001"
-                },
-                {
-                    "label": "embedding-001",
-                    "name": "embedding-001"
-                },
-                {
-                    "label": "text-embedding-004",
-                    "name": "text-embedding-004"
                 }
             ]
         },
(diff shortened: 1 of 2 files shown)


adolfousier:
Yo sorry to not have pushed this together with the other commit, just found out those two remaining models were deprecated after.

HenryHengZJ:
thanks!

PR #5718, opened 2026-02-08
fix: handle humanInputAgentflow conditions in resolved value

prd-hoang-doan:
# Fix: Resolve humanInputAgentflow variables in subsequent nodes

## Problem
When referencing humanInputAgentflow node outputs in subsequent nodes using handlebars syntax (e.g., `{{ humanInputAgentflow_0 }}`), the variables were not being interpolated. Instead, the literal handlebars string was passed through, making it impossible to use human input values in downstream nodes.

### Current

https://github.com/user-attachments/assets/0a4ec578-6252-45c7-9fa6-9511dc3f6b91

### After fixing

https://github.com/user-attachments/assets/28d74d1f-79c0-49d6-b2f1-529057053b3e

### DIFF
--- a/packages/server/src/utils/buildAgentflow.ts
+++ b/packages/server/src/utils/buildAgentflow.ts
@@ -398,6 +398,15 @@ export const resolveVariables = async (
                 // Replace the reference with actual value
                 const nodeOutput = nodeData.data['output'] as ICommonObject
                 const actualValue = nodeOutput?.content ?? nodeOutput?.http?.data
+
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
@prd-hoang-doan this will always fill in the variable with `proceed` or `reject`, it does not capture the feedback:
image

HenryHengZJ:
Closing as https://github.com/FlowiseAI/Flowise/pull/5818 fixes it

PR #5719, opened 2026-02-08
fix: update import paths for pdf.js to use .mjs extension

prd-hoang-doan:
# Fix: Correct PDF.js legacy build import path
## Problem
The PDF loader components were importing the legacy PDF.js build using an incorrect path (pdf.js) instead of the correct ESM module path (pdf.mjs). This caused issues when users enabled the "Use Legacy Build" option for PDF compatibility.

image

## Testing:

https://github.com/user-attachments/assets/e6878793-ef81-474b-8729-8d070049dc1a

### DIFF
--- a/packages/components/nodes/documentloaders/File/File.ts
+++ b/packages/components/nodes/documentloaders/File/File.ts
@@ -236,13 +236,13 @@ class File_DocumentLoaders implements INode {
                           splitPages: false,
                           pdfjs: () =>
                               // @ts-ignore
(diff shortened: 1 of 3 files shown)


okurayhan:
I opened this issue before: https://github.com/FlowiseAI/Flowise/pull/5175
[test.pdf](https://github.com/user-attachments/files/25514986/test.pdf)

please can you check this pdf with legacy build?

prd-hoang-doan:
@okurayhan I re-checked the issue and realized that the pull request doesn't fix the issue with the legacy PDF from version 4 and 5. The current CommonJS build is not allowed to import a `.mjs` file directly. So I will try to fix it in the future.

PR #5722, opened 2026-02-08
fix(agentflow): implement regex operation in Condition node

### LINKED ISSUE #5650: Bug: "Regex" operation in Condition node is not implemented in backend

kaigritun:
## Summary
Fixes #5650

The Condition node's frontend UI offers a 'Regex' operation for string comparisons, but the backend was missing the corresponding handler in `compareOperationFunctions`. This caused the runtime error:
```
compareOperationFunctions[operation] is not a function
```

## Changes
Added the `regex` handler to the `compareOperationFunctions` map that:
- Takes `value1` as the string to test
- Takes `value2` as the regex pattern  
- Returns `true` if `value1` matches the regex pattern
- Returns `false` for invalid regex patterns (graceful fallback with try/catch)

## Testing
To verify the fix:
3. Select Operation: "Regex"
4. Set Value 1 to any string variable (e.g., `hello world`)
5. Set Value 2 to a regex pattern (e.g., `^hello`)
6. Run the flow
7. The condition should now evaluate correctly instead of throwing an error

### DIFF
--- a/packages/components/nodes/agentflow/Condition/Condition.ts
+++ b/packages/components/nodes/agentflow/Condition/Condition.ts
@@ -1,5 +1,6 @@
 import { CommonType, ICommonObject, ICondition, INode, INodeData, INodeOutputsValue, INodeParams } from '../../../src/Interface'
 import removeMarkdown from 'remove-markdown'
+import safeRegex from 'safe-regex2'
 
 class Condition_Agentflow implements INode {
     label: string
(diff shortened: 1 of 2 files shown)


kaigritun:
Good catch on the ReDoS concern! I'll add safe-regex validation to protect against catastrophic backtracking. Pushing a fix shortly.

kaigritun:
✅ **ReDoS vulnerability fixed!**

I've addressed the security concern by adding `safe-regex2` validation:

**Changes made:**
- Added `safe-regex2` library as a dependency
- Replaced manual ReDoS pattern check with `safeRegex(pattern)` validation
- Regex pattern is now validated before `new RegExp()` is called

The `safe-regex2` library performs comprehensive checks for catastrophic backtracking patterns, providing more robust protection than the previous manual approach.

**Commit:** 404b5a13893500efaf8e6cb88736d89f448cb3fb

@gemini-code-assist This should resolve the ReDoS vulnerability you identified. The pattern is now validated using a well-tested library before regex execution.

HenryHengZJ:

PR #5731, opened 2026-02-09
feat: add AWS STS AssumeRole support to AWS credential

rohan-patil2:
## Summary

- Adds **Role ARN** and **External ID** fields to the AWS credential form (`awsApi`), enabling STS AssumeRole authentication across all 9 AWS-integrated nodes
- Introduces a centralized `getAWSCredentialConfig()` utility that handles static credentials, AssumeRole with explicit keys, and AssumeRole via the SDK default credential chain
- Includes Role ARN format validation (with support for `aws`, `aws-cn`, `aws-us-gov` partitions)

## Background

Flowise currently only supports static IAM access keys for AWS authentication. Many organizations require **role-based access** for cross-account workflows, least-privilege enforcement, and temporary credential rotation. This change adds AssumeRole support so users can enter a Role ARN in the credential form and have Flowise obtain temporary credentials via STS automatically.

## Supported Authentication Scenarios

| Access Key | Secret Key | Role ARN | External ID | Behavior |
|:---:|:---:|:---:|:---:|---|
| Yes | Yes | — | — | Static credentials (existing behavior, unchanged) |
| Yes | Yes | Yes | — | AssumeRole using static keys as base credentials |
| Yes | Yes | Yes | Yes | AssumeRole with External ID for cross-account trust |
| — | — | Yes | — | AssumeRole using SDK default credential chain (EC2/EKS) |
| — | — | Yes | Yes | AssumeRole with External ID using SDK default chain |
| — | — | — | — | SDK default chain fallback (Bedrock nodes) or error |

## Backward Compatibility

- Existing saved credentials work without modification — new fields default to empty
- Credential name (`awsApi`) is unchanged; no database migration required
- `getAWSCredentials()` signature and return type are preserved
- Workflows that don't use Role ARN behave identically to before

## Security

- Role ARN format is validated with a regex before any STS call is made (prevents confused deputy / injection)
- Temporary credentials are scoped to STS default and are obtained per-execution (no caching)
- External ID field supports the standard AWS confused deputy mitigation
- Session name (`FlowiseSession-{timestamp}`) provides CloudTrail audit traceability

aws_credential_explicit_assume_role

### DIFF
--- a/packages/components/nodes/tools/AWSDynamoDBKVStorage/AWSDynamoDBKVStorage.ts
+++ b/packages/components/nodes/tools/AWSDynamoDBKVStorage/AWSDynamoDBKVStorage.ts
@@ -249,11 +249,11 @@ class AWSDynamoDBKVStorage_Tools implements INode {
         ]
     }
 
-    loadMethods: Record<string, (nodeData: INodeData, options?: ICommonObject) => Promise<INodeOptionsValue[]>> = {
-        listTables: async (nodeData: INodeData, options?: ICommonObject): Promise<INodeOptionsValue[]> => {
+    loadMethods: Record<string, (_nodeData: INodeData, _options?: ICommonObject) => Promise<INodeOptionsValue[]>> = {
+        listTables: async (_nodeData: INodeData, _options?: ICommonObject): Promise<INodeOptionsValue[]> => {
             try {
-                const credentials = await getAWSCredentials(nodeData, options ?? {})
-                const region = (nodeData.inputs?.region as string) || DEFAULT_AWS_REGION
+                const credentials = await getAWSCredentials(_nodeData, _options ?? {})
+                const region = (_nodeData.inputs?.region as string) || DEFAULT_AWS_REGION
                 const dynamoClient = createDynamoDBClient(credentials, region)
 
                 const listCommand = new ListTablesCommand({})


harshit-flowise:
@rohan-patil2 
1. Please review and fix the issues highlighted by gemini-code-assist.
2. Below are few more things to fix -
 --Role ARN in error messages — The Role ARN is included in thrown errors. Ensure these errors are not exposed directly to end-users in API responses, as ARNs reveal account IDs. This is likely fine given Flowise's architecture but worth verifying.
 
-- Credential version bump (1.0 → 2.0) — The PR description says no migration is needed, but bumping the credential 
version to 2.0 may have implications depending on how Flowise handles credential versioning. If the system enforces version checks, older stored credentials could behave unexpectedly. Verify that the version bump is purely cosmetic or that the upgrade path is handled.

--AWS_REGIONS duplication — This regions list likely already exists elsewhere in the codebase (several AWS nodes define their own region lists). Centralizing it here is good, but only if the nodes are updated to import from awsToolsUtils.ts. If they still maintain their own copies, this adds duplication rather than reducing it.

--SDK default chain fallback inconsistency — When getAWSCredentialConfig returns { credentials: undefined }, some nodes check if (credentialConfig.credentials) before assigning, which correctly falls through to the SDK default chain. But in S3Directory.ts the diff shows credentials = credentialConfig.credentials without a guard — if credentials was previously initialized to a non-undefined value, this would overwrite it with undefined. Verify all consumer nodes handle the undefined case consistently.

--getAWSCredentials() breaks SDK default chain — The backward-compat wrapper throws "AWS Access Key ID and Secret Access Key are required" when config.credentials is undefined. This means Pattern A nodes (SNS, DynamoDB) can never use the SDK default credential chain (EC2/EKS instance profiles). This may be intentional but should be documented as a known limitation.

3. Build the image and test it in AWS to make sure it is working fine and not breaking anything.

harshit-flowise:
@rohan-patil2 Build step is failing. Please fix this before we can merge it to main.

rohan-patil2:
@harshit-flowise Done, fixed.

HenryHengZJ:
there are several unused vars `is defined but never used` - can you do `pnpm lint-fix` ?

rohan-patil2:
@HenryHengZJ 
Done, all lint issues related to the changes in this PR have been resolved.

PR #5732, opened 2026-02-09
Implement regex operation for Agentflow Condition

### LINKED ISSUE #5650: Bug: "Regex" operation in Condition node is not implemented in backend

Deeven-Seru:
## Summary
- add regex handler to the Agentflow Condition backend compare map so the frontend "Regex" option executes instead of throwing
- support simple /pattern/flags parsing and safely fall back to plain patterns; invalid regex input returns false instead of crashing
- guard against unknown operations before execution to avoid runtime TypeErrors

Fixes #5650.

## Testing
- not run (not configured in this environment)


HenryHengZJ:

PR #5749, opened 2026-02-13
feat(agentflow): add test infrastructure and unit tests

jocelynlin-wd:
Add Jest test suite with coverage enforcement for @flowise/agentflow. Includes 119 tests across core logic, API layer, search, and state management. Fixes trailing-slash bug in getNodeIconUrl and extracts reducer to its own module for cleaner separation.

### DIFF
--- a/packages/agentflow/src/__test_utils__/factories.ts
+++ b/packages/agentflow/src/__test_utils__/factories.ts
@@ -47,4 +47,4 @@ export const makeFlowEdge = (source: string, target: string, overrides?: Partial
  * makeNodeData({ name: 'llmAgentflow', label: 'LLM', category: 'AI' })
  */
 export const makeNodeData = (overrides?: Partial<NodeData>): NodeData =>
-    ({ id: '', name: 'testNode', label: 'Test Node', ...overrides }) as NodeData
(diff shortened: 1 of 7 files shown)


packages/agentflow/src/infrastructure/api/hooks/useApi.ts
```diff
@@ -11,6 +11,7 @@ interface UseApiReturn<T> extends UseApiState<T> {
     reset: () => void
 }
 
+// TODO: Review if still necessary — package uses ApiContext + dedicated hooks (useNodesApi, useChatflowsApi) instead
```
muratkaratekeli, line 14: useApi hook isn't used anywhere in the package, all components use useApiContext() instead. Fine to exclude from coverage for now, but can we create a follow up ticket to either remove it or document why we're keeping it? 
Don't want it sitting in limbo.

jocelynlin-wd: let me look at the appropriate ticket and add notes there to ensure there is the cleanup (I think this will be done when we update context)

jocelynlin-wd: updated the work item.

PR #5751, opened 2026-02-13
Remove flowData from getSinglePublicChatbotConfig response

christopherholland-workday:
`flowData` contains sensitive information that it shouldn't be. In fact, it looks like the `flowData` is not used anywhere when this method is called, thus we should remove it.


packages/server/src/services/chatflows/index.ts
```diff
@@ -388,7 +388,7 @@ const getSinglePublicChatbotConfig = async (chatflowId: string): Promise<any> =>
                 }
                 delete parsedConfig.allowedOrigins
                 delete parsedConfig.allowedOriginsError
-                return { ...parsedConfig, uploads: uploadsConfig, flowData: dbResponse.flowData, isTTSEnabled }
+                return { ...parsedConfig, uploads: uploadsConfig, isTTSEnabled }
```
HenryHengZJ: We cannot remove `flowData`, this is being used by [FlowiseChatEmbed](https://github.com/FlowiseAI/FlowiseChatEmbed/blob/main/src/components/Bot.tsx#L1369) to retrieve the data. 

`flowData` does not contains any sensitive data like apikeys, only credential ID.

christopherholland-workday: Ack, we will sanitize sensitive data instead

Issue #5736, opened 2026-02-10
Renaming a flow opens an empty flow overview


jlistemann-ish:
### Describe the bug

Renaming a flow opens an empty flow overview.

### To Reproduce

1. Chatflows
2. Renaming a flow and confirm
3. See an empty overview

### Expected behavior

the flow overview

### Screenshots

<img width="592" height="643" alt="Image" src="https://github.com/user-attachments/assets/034d347f-fb36-4c1b-99b4-d16387223c8b" />

<img width="797" height="307" alt="Image" src="https://github.com/user-attachments/assets/cd4060a6-838b-45fd-b663-2f92adb098c0" />

### Use Method

None

### Flowise Version

3.0.11 Enterprise
### Operating System

None

### Browser

None

PR #5754, opened 2026-02-14
fix: #5736 Navigate to canvas after FlowListTable rename

KUNDAN1334:
**Before:** Rename from List View → returns to chatflows list 
**After:** Rename from List View → navigates to canvas editor 

### Changes
-  **Removed** `updateFlowsApi.request()` table refresh after rename
-  **Added** `useNavigate()` to `/canvas/${chatflow.id}` (or agentcanvas routes)
-  **Cleaned** API payload: `{ name: chatflowName }` (removed extra `chatflow` prop)
-  **Added** success/error snackbar notifications
-  **Fixed** `useNavigate` hook placement (inside component)
-  **Multi-mode support**: Chatflow + Agentflow V1/V2 routing

### DIFF
--- a/packages/ui/src/ui-component/button/FlowListMenu.jsx
+++ b/packages/ui/src/ui-component/button/FlowListMenu.jsx
@@ -180,41 +180,27 @@ export default function FlowListMenu({ chatflow, isAgentCanvas, isAgentflowV2, s
     }
 
     const saveFlowRename = async (chatflowName) => {
-        const updateBody = {
-            name: chatflowName,
-            chatflow
-        }
         try {
-            await updateChatflowApi.request(chatflow.id, { name: chatflowName });  
-            
+            await updateChatflowApi.request(chatflow.id, { name: chatflowName })
+            setFlowDialogOpen(false)
             enqueueSnackbar({
                 message: 'Flow renamed successfully!',
                 options: { variant: 'success' }
-            });
-            
-            
+            })
+
             if (isAgentCanvas && isAgentflowV2) {
-                navigate(`/v2/agentcanvas/${chatflow.id}`, { replace: true });
+                navigate(`/v2/agentcanvas/${chatflow.id}`, { replace: true })
             } else if (isAgentCanvas) {
-                navigate(`/agentcanvas/${chatflow.id}`, { replace: true });
+                navigate(`/agentcanvas/${chatflow.id}`, { replace: true })
             } else {
-                navigate(`/canvas/${chatflow.id}`, { replace: true });
+                navigate(`/canvas/${chatflow.id}`, { replace: true })
             }
         } catch (error) {
-            if (setError) setError(error)
             enqueueSnackbar({
-                message: typeof error.response.data === 'object' ? error.response.data.message : error.response.data,
-                options: {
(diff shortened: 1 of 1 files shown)


packages/ui/src/ui-component/button/FlowListMenu.jsx
```diff
@@ -183,17 +185,20 @@ export default function FlowListMenu({ chatflow, isAgentCanvas, isAgentflowV2, s
...
             if (isAgentCanvas && isAgentflowV2) {
-                await updateFlowsApi.request('AGENTFLOW', params)
+                navigate(`/v2/agentcanvas/${chatflow.id}`, { replace: true });
             } else if (isAgentCanvas) {
-                await updateFlowsApi.request('MULTIAGENT', params)
+                navigate(`/agentcanvas/${chatflow.id}`, { replace: true });
             } else {
-                await updateFlowsApi.request(params)
+                navigate(`/canvas/${chatflow.id}`, { replace: true });
             }
```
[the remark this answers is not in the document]
KUNDAN1334: done the suggested changes

packages/ui/src/ui-component/button/FlowListMenu.jsx
```diff
@@ -178,38 +180,27 @@ export default function FlowListMenu({ chatflow, isAgentCanvas, isAgentflowV2, s
...
+            await updateChatflowApi.request(chatflow.id, { name: chatflowName })
+            setFlowDialogOpen(false)
+            enqueueSnackbar({
+                message: 'Flow renamed successfully!',
+                options: { variant: 'success' }
+            })
+
             if (isAgentCanvas && isAgentflowV2) {
-                await updateFlowsApi.request('AGENTFLOW', params)
+                navigate(`/v2/agentcanvas/${chatflow.id}`, { replace: true })
```
chloebyun-wd: Hi @KUNDAN1334! The `replace: true` navigates the user to the correct canvas page after rename but then the browser history isn't correctly stacked so if a user renames their board, and goes back a page, they would be incorrectly routed - effectively going back two pages. 

Is there context on why we want to make this change (ie. bringing the user to their canvas change once the board is renamed)? I see https://github.com/FlowiseAI/Flowise/issues/5736 but not sure what the exact expected behaviour they were wanting to see.

KUNDAN1334: Hi @chloebyun-wd  Great catches, thanks for the review.

**On `replace: true`:** You're right — that was a mistake on my part. I'll remove `replace: true` so the navigation uses a normal push, keeping the browser history stack intact. Users will be able to go back to the list as expected.

**On the UX rationale:** The issue (#5736) reports that after renaming a flow from the list view, the user sees an empty flow overview instead of their flow. The root cause is that the list view refreshes and loses state, leaving an empty/broken screen.

Navigating to the canvas after rename was my approach to avoid that broken state — the user lands somewhere meaningful instead of a broken overview. That said, if the preferred fix is to keep the user on the list view, I'm happy to instead fix the broken overview state directly (e.g. re-fetching the flows list correctly after rename) rather than navigating away. Let me know which direction you'd prefer and I'll update accordingly!

PR #5773, opened 2026-02-18
Add protections for loop-bound injections

christopherholland-workday:
Adding checks to prevent DoS due to large loop boundaries.


HenryHengZJ:
im not a fan of hardcoding limit to prevent DoS. Since `evaluateChatflow` function is calling the prediction endpoint, this is essentially the same as DoS-ing `/prediction` endpoint. IMO, this is more on the infra level protection. Our AWS WAF should already give us some default DoS protection

christopherholland-workday:
Not sure why this says it was merged, it is not in the history: https://github.com/FlowiseAI/Flowise/pulls?q=is%3Apr+is%3Aclosed

And I did not merge it

PR #5774, opened 2026-02-18
feat(agentflow): update canvas add node drag and drop behaviour

jocelynlin-wd:
- fix canvas add node drag and drop behaviour
- fix style issues and refactor components using design token with dark mode theme support
- add and update tests with additional jest config to mock canvas and libs

https://github.com/user-attachments/assets/bd35848c-9efa-419a-a0a2-ac37e40fc974

https://github.com/user-attachments/assets/999e1b02-5e48-4897-956e-7730fb60a882

### DIFF
--- a/packages/agentflow/src/core/theme/tokens.ts
+++ b/packages/agentflow/src/core/theme/tokens.ts
@@ -1,0 +1,170 @@
+/**
+ * Design Tokens for Agentflow
+ *
+ * Single source of truth for all design values (colors, spacing, shadows, etc.)
+ * These tokens are used by both MUI theme and CSS variables.
+ *
+ * Architecture:
+ * 1. Base colors - Primitive color values (single source of truth)
+ * 2. Semantic colors - Referenced from base colors for consistency
+ */
+
+// Base primitive colors - define once, reference everywhere
+const baseColors = {
+    // Neutral colors
+    white: '#fff',
+    black: '#000',
+
+    // Light mode grays
+    gray50: '#fafafa',
+    gray75: '#f5f5f5',
+    gray100: '#f8f9fa',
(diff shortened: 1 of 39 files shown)


packages/agentflow/src/core/theme/tokens.ts
```diff
@@ -0,0 +1,170 @@
...
+ * Single source of truth for all design values (colors, spacing, shadows, etc.)
+ * These tokens are used by both MUI theme and CSS variables.
+ *
+ * Architecture:
+ * 1. Base colors - Primitive color values (single source of truth)
+ * 2. Semantic colors - Referenced from base colors for consistency
+ */
+
+// Base primitive colors - define once, reference everywhere
+const baseColors = {
```
tianwei-liu, line 13: nit: colors can be taken out to colors.ts

tianwei-liu (approved):

packages/agentflow/src/core/theme/tokens.ts
jocelynlin-wd: Thanks for the feedback. Will break this out when this file become too large

> Here's why:
> 
> 1. They're part of the token definition — baseColors is a private implementation detail (not exported), and tokens.colors is the public API. Splitting them would break this clean encapsulation.
> 
> 2. Single source of truth — The file's stated architecture (lines 8-9) is that base colors feed into semantic colors. Moving colors out would split that relationship across files.
> 
> 3. The file is ~170 lines — It's well within a manageable size. Extraction would add indirection without reducing complexity.
> 
> The only reason to extract would be if baseColors grew significantly (e.g., 50+ brand colors) or needed to be shared with other token systems outside of tokens.ts. Right now neither applies.

PR #5778, opened 2026-02-18
Validate filepaths in file utilities

christopherholland-workday:
1. Validate that `chatId` and `chatflowId` are valid UUID's
2. Ensure that the `filename` does not contain path traversal patterns
3. Ensure that the sanitized `filename` does not contain path traversal patterns

Addresses CodeQL findings 45-50

Moved to https://github.com/FlowiseAI/Flowise/pull/5842 because this PR got polluted by mistake


packages/components/src/storageUtils.ts
```diff
@@ -752,13 +752,25 @@ export const streamStorageFile = async (
...
     }
 
-    // Check for path traversal attempts
+    if (!chatId || !isValidUUID(chatId)) {
+        throw new Error('Invalid chatId format - must be a valid UUID')
+    }
+
     if (isPathTraversal(chatflowId) || isPathTraversal(chatId)) {
         throw new Error('Invalid path characters detected in chatflowId or chatId')
     }
```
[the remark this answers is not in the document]
christopherholland-workday: This is more of a logical "backup" if things were moved around unknowingly later on

packages/components/src/storageUtils.ts
```diff
@@ -752,13 +752,25 @@ export const streamStorageFile = async (
         throw new Error('Invalid chatflowId format - must be a valid UUID')
     }
 
-    // Check for path traversal attempts
+    if (!chatId || !isValidUUID(chatId)) {
```
HenryHengZJ: We didnt include chatID uuid check is because in some cases, `chatID` can be overriden from the API call.

For example, users can override like this:
```
{
   question: how are you?
   chatId: sessionABC
}
```

packages/server/src/controllers/get-upload-file/index.ts
```diff
@@ -18,6 +18,20 @@ const streamUploadedFile = async (req: Request, res: Response, next: NextFunctio
         const fileName = req.query.fileName as string
         const download = req.query.download === 'true' // Check if download parameter is set
 
+        // Validate input formats to prevent path traversal attacks
+        if (!chatflowId || !isValidUUID(chatflowId)) {
+            return res.status(400).send(`Invalid chatflowId format`)
+        }
+
+        if (!chatId || !isValidUUID(chatId)) {
```
HenryHengZJ: same as above

PR #5797, opened 2026-02-20
feat(embeddings): add Matryoshka Embeddings node

### LINKED ISSUE #4361: [FEATURE] Allow using Matryoshka Embeddings
**Describe the feature you'd like**
Certain embedding models produce so called "Matryoshka Embeddings".
These embedding vectors have the most information in the first digits of the vector, and the further we go to the back of the vector the less information the digits convey.

This means, that I  can take an embedding with e.g. 1024 dimensions, cut off the last 768 dimensions and keep only the first 256 dimensions (or first 100, first 500 etc.). These first 256 have more meaning than the cut off 768 dimensions. 
This means that instead of storing 1024 dimensions, I can store only 256 dimensions which requires much less space, without loosing a lot of information.

Right now, I cannot use a different dimension when upserting than the embedding model yields via the DocStore. 
I guess it should be easy to implement. Instead of throwing an error ...
```
Status: 500
Error: documentStoreServices.insertIntoVectorStoreMiddleware - Error: documentStoreServices.insertIntoVectorStore - Error: documentStoreServices._insertIntoVectorStoreWorkerThread - Error: 400 Bad Request: Wrong input: Vector dimension error: expected dim: 256, got 1024
``` 
... let me specify the number of dimensions that I wanna store and cut off the rest:

Image

**Additional context**
Read more about the marvelous matryoshka embeddings on HF: https://huggingface.co/blog/matryoshka

hztBUAA:
## Summary

- Adds a new **Matryoshka Embeddings** node that wraps any existing embeddings model and truncates output vectors to a user-specified number of dimensions
- Enables efficient storage and retrieval with embedding models trained using the [Matryoshka loss function](https://huggingface.co/blog/matryoshka), where the most significant information is concentrated in the first dimensions of the vector
- Implemented as a composable wrapper node: users connect any Embeddings node as input and specify the target dimension count, with no changes needed to existing embedding or vector store nodes

## How It Works

The node takes two inputs:
1. **Embeddings** - Any existing embeddings node (OpenAI, Cohere, HuggingFace, Ollama, etc.)
2. **Dimensions** - The target number of dimensions to truncate to

After the underlying model generates full-dimensional vectors, the wrapper truncates them by keeping only the first N dimensions via `vector.slice(0, dimensions)`.

### New Files
- `packages/components/src/matryoshkaEmbeddings.ts` - Core `MatryoshkaEmbeddings` wrapper class extending LangChain's `Embeddings`
- `packages/components/nodes/embeddings/MatryoshkaEmbedding/MatryoshkaEmbedding.ts` - Flowise node implementation
- `packages/components/nodes/embeddings/MatryoshkaEmbedding/matryoshka.svg` - Node icon

## Test Plan

- [x] Unit tests for `MatryoshkaEmbeddings` class (vector truncation, edge cases, delegation)
- [x] Unit tests for the node's `init()` method (validation, integration)
- [x] All 16 tests passing

```
PASS nodes/embeddings/MatryoshkaEmbedding/MatryoshkaEmbedding.test.ts
PASS src/matryoshkaEmbeddings.test.ts
Test Suites: 2 passed, 2 total
Tests:       16 passed, 16 total
```

Closes #4361

### DIFF
--- a/packages/components/nodes/embeddings/MatryoshkaEmbedding/MatryoshkaEmbedding.test.ts
+++ b/packages/components/nodes/embeddings/MatryoshkaEmbedding/MatryoshkaEmbedding.test.ts
@@ -1,0 +1,130 @@
+jest.mock('../../../src/utils', () => ({
+    getBaseClasses: jest.fn(() => ['Embeddings'])
+}))
+
+const { nodeClass: MatryoshkaEmbeddingNode } = require('./MatryoshkaEmbedding')
+
+describe('MatryoshkaEmbedding Node', () => {
+    let node: InstanceType<typeof MatryoshkaEmbeddingNode>
+    let mockEmbeddings: { embedDocuments: jest.Mock; embedQuery: jest.Mock }
+
+    beforeEach(() => {
+        jest.clearAllMocks()
+        node = new MatryoshkaEmbeddingNode()
+        mockEmbeddings = {
+            embedDocuments: jest.fn().mockResolvedValue([[0.1, 0.2, 0.3, 0.4]]),
+            embedQuery: jest.fn().mockResolvedValue([0.1, 0.2, 0.3, 0.4])
+        }
+    })
+
+    describe('constructor', () => {
+        it('should have correct metadata', () => {
+            expect(node.label).toBe('Matryoshka Embeddings')
+            expect(node.name).toBe('matryoshkaEmbeddings')
+            expect(node.category).toBe('Embeddings')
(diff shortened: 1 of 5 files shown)


hztBUAA:
Thanks for the review and feedback. I am following up on this PR now and will either push the requested changes or reply point-by-point shortly.

hztBUAA:
Quick follow-up: I am reviewing the feedback and will update this PR shortly.

PR #5801, opened 2026-02-20
fix: preserve flow content when renaming a flow

### LINKED ISSUE #5736: Renaming a flow opens an empty flow overview

hztBUAA:
## Summary

Fixes #5736 - Renaming a flow from the list view caused the flow overview to appear empty.

**Root cause:** `saveFlowRename` and `saveFlowCategory` in `FlowListMenu` used the `useApi` hook to call `updateChatflow`. The `useApi` hook's internal error handling calls `ErrorContext.setError()`, which shares state with the parent Chatflows page. If the update triggered any error (even transient), it would cause the page to render an `ErrorBoundary` instead of the flow list, making it appear empty.

**Changes:**
- **Call `chatflowsApi.updateChatflow` directly** instead of through `useApi`, so errors are caught locally in try/catch without polluting the page-level error state
- **Close the rename dialog on success** (`setFlowDialogOpen(false)`) — previously it stayed open after confirming
- **Send only the changed field** (`{ name }` or `{ category }`) in the request body — previously the entire `chatflow` object was unnecessarily included
- **Add optional chaining** on `error.response?.data` to prevent crashes when the error lacks a response property

## Test plan

- [ ] Open Chatflows list view (table mode)
- [ ] Click Options > Rename on any flow
- [ ] Enter a new name and confirm
- [ ] Verify the dialog closes and the list refreshes with the updated name
- [ ] Verify the flow content is preserved (open the renamed flow in canvas)
- [ ] Repeat for Update Category from the same menu
- [ ] Repeat the above steps for Agentflows (both V1 and V2)

### DIFF
--- a/packages/ui/src/ui-component/button/FlowListMenu.jsx
+++ b/packages/ui/src/ui-component/button/FlowListMenu.jsx
@@ -22,7 +22,6 @@ import { IconX } from '@tabler/icons-react'
 
 import chatflowsApi from '@/api/chatflows'
 
-import useApi from '@/hooks/useApi'
 import useConfirm from '@/hooks/useConfirm'
 import { uiBaseURL } from '@/store/constant'
 import { closeSnackbar as closeSnackbarAction, enqueueSnackbar as enqueueSnackbarAction } from '@/store/actions'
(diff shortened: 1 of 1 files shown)


hztBUAA:
Thanks for the review and feedback. I am following up on this PR now and will either push the requested changes or reply point-by-point shortly.

hztBUAA:
Quick follow-up: I am reviewing the feedback and will update this PR shortly.

PR #5810, opened 2026-02-20
feat: add LMStudio

sohaieb:
## Changes
- Add [LMStudio](https://lmstudio.ai/) nodes support 
- Fix an issue when navigating between multilines of chat textarea
- Fix add user input to chat history

## Important note (Issue)
**[BACKGROUND]**
~~All LMStudio nodes work well, however the **LMStudio embedding are unstable:** the vector column values are always stored with "0" values.~~
~~Not sure but It seems like Followise bases on these values?~~
~~I tried to simulate the same embedding process with Ollama but it returns correct values.~~
~~BTW I opened an issue to LMStudio basing on this topic, [please check it out HERE](https://github.com/lmstudio-ai/lmstudio-bug-tracker/issues/1546)~~

**[UPDATE]**
Regarding the previously mentioned issue and basing on @0xi4o comment:
> this is a known issue with @langchain/openai sdk and LM Studio: https://github.com/langchain-ai/langchainjs/issues/8221

The issue should be fixed now with this Pull Request, basing on the following [Thread Comment](https://github.com/FlowiseAI/Flowise/pull/5810#discussion_r3032046245).


HenryHengZJ:
what is Mise? why is it needed here?

sohaieb:
Hi dear @HenryHengZJ , 
> what is Mise? why is it needed here?

Mise ([Official Docs](https://mise.jdx.dev/)) helps manage tools like Node and Python per project. It installs and configures them locally and prevents version conflicts.

Please check the docs and let me know if we should include it in Flowise, so I can keep or remove it from the PR.

HenryHengZJ:
it would be great if we can just keep this PR for LMStudio integration

sohaieb:
> it would be great if we can just keep this PR for LMStudio integration

@HenryHengZJ sure, I will remove the mise integration accordingly and update the branch as it is outdated

**[UPDATE]**
I updated the PR to remove Mise support ✅

packages/components/credentials/LmStudioApi.credential.ts
```diff
@@ -0,0 +1,24 @@
+import { INodeParams, INodeCredential } from '../src/Interface'
+
+class LMStudioApi implements INodeCredential {
+    label: string
+    name: string
+    version: number
+    inputs: INodeParams[]
+
+    constructor() {
+        this.label = 'LM Studio API'
```
0xi4o, line 11: Can you also add a `this.description` and mention how to get the API key? It'll be helpful for users and save them a few extra steps. Refer to `AWSCredential.credential.ts` on how to write a description.

packages/components/nodes/chatmodels/ChatLmStudio/ChatLmStudio.ts
```diff
@@ -0,0 +1,149 @@
...
+    type: string
+    icon: string
+    category: string
+    description: string
+    baseClasses: string[]
+    credential: INodeParams
+    inputs: INodeParams[]
+
+    constructor() {
+        this.label = 'Chat LMStudio'
```
0xi4o: `this.label` can be just "LMStudio" so it's consistent with the names of other chat model nodes.

0xi4o (changes requested):

0xi4o:
@HenryHengZJ Tested this PR:

Screenshot 2026-04-01 at 8 36 54 PM
Screenshot 2026-04-01 at 8 38 19 PM
Screenshot 2026-04-01 at 8 42 27 PM
Screenshot 2026-04-01 at 8 42 41 PM
Screenshot 2026-04-01 at 8 42 53 PM
Screenshot 2026-04-01 at 8 44 06 PM

As for the 0 usage data: afaik, we only measure usage for llm or chat models, not for embeddings. In my testing, using LMStudio Embeddings worked well as shown in the screenshots above.

packages/ui/src/views/chatmessage/ChatMessage.jsx
```diff
@@ -1021,16 +1023,16 @@ const ChatMessage = ({ open, chatflowid, isAgentCanvas, isDialog, previews, setP
...
-            if (input.trim()) {
-                inputHistory.addToHistory(input)
-            }
         } else if (typeof selectedInput === 'object') {
             input = Object.entries(selectedInput)
                 .map(([key, value]) => `${key}: ${value}`)
                 .join('\n')
         }
 
+        if (input.trim()) {
```
HenryHengZJ, line 1032: why the change here, whats not working previously?

sohaieb: Hi @HenryHengZJ ,
- **Short answer:** From what I remember, these changes make sure the input is always added to the history, no matter its type (string or object).

- **Background:** Previously (lines 1024–1029), we only added the `input` to `inputHistory` when it was a `string` , in the following condition ONLY:
  `if (typeof selectedInput === 'string') { ... }`
  This meant that object inputs were skipped and never saved.

  Since we now convert any input (including objects) into a string using:
```ts
... 
} else if (typeof selectedInput === 'object') {
            input = Object.entries(selectedInput)
                .map(([key, value]) => `${key}: ${value}`)
                .join('\n')
        }
...
```

  it makes sense to always store the input in `inputHistory` after both conditions, instead of skipping it when it's an object.

Please let me know if this answers your question 🙂

packages/components/credentials/LmStudioApi.credential.ts
sohaieb: Hi @0xi4o ,
thank you so much for the catch ! 
I updated the PR accordingly ✔️ 
can you please re-verify again ? 
thanks !

packages/components/nodes/chatmodels/ChatLmStudio/ChatLmStudio.ts
sohaieb: @0xi4o good point ! 
I updated basing on your comment ✔️

sohaieb:
> @HenryHengZJ Tested this PR:
> 
> Screenshot 2026-04-01 at 8 36 54 PM ...
> ...
> As for the 0 usage data: afaik, we only measure usage for llm or chat models, not for embeddings. In my testing, using LMStudio Embeddings worked well as shown in the screenshots above.

Hi @0xi4o 

I’m not 100% sure this is the issue, but could you double-check something for me: Can you look at the `embeddings` column in your database? You might find that the values are all `0,0,0,...`. Please confirm if that’s the case.

Another hint: try asking the AI a few different questions. If you notice that the document store keeps returning results from the same (first) embedding entry, even when the question changes, it could be related.

From what I understand, if all embedding values are `0,0,0,...`, the similarity search won’t work properly and will just fall back to the first entries. (But feel free to correct me if I’m wrong!)

If it's the case, I'm not sure if this is faced exactly in Flowise stage or in LmStudio stage, but as a reference I opened the following thread if you need more details: ([Embedding response usage (prompt_tokens, total_tokens) stats are always 0](https://github.com/lmstudio-ai/lmstudio-bug-tracker/issues/1546))

0xi4o:
@sohaieb You're right! I was able to verify that the values are all zeros. And the chat model kept returning the same chunks in the responses for different questions.

On further investigation, I found that this is a known issue with `@langchain/openai` sdk and LM Studio: https://github.com/langchain-ai/langchainjs/issues/8221 

The fix is to set `encodingFormat` explicitly in model params in `LmStudioEmbedding.ts`.

packages/components/nodes/embeddings/LMStudioAIEmbedding/LmStudioEmbedding.ts
```diff
@@ -0,0 +1,74 @@
...
+
+    async init(nodeData: INodeData, _: string, options: ICommonObject): Promise<any> {
+        const modelName = nodeData.inputs?.modelName as string
+        const baseURL = nodeData.inputs?.baseURL as string
+
+        const credentialData = await getCredentialData(nodeData.credential ?? '', options)
+        const lmStudioApiKey = getCredentialParam('lmStudioApiKey', credentialData, nodeData)
+
+        const obj: Partial<LmStudioEmbeddingsParams> & { configuration?: ClientOptions } = {
+            modelName,
```
0xi4o, line 61: Let's set the `encodingFormat` explicitly to fix the 0 values in embeddings.

```ts
encodingFormat: 'float',
```

0xi4o (changes requested):

packages/components/nodes/embeddings/LMStudioAIEmbedding/LmStudioEmbedding.ts
sohaieb: Hi @0xi4o 
oh thank you so much for your recommended solution ! 
I fixed it and tested it successfully ! 
the solution seems to be working very well now ! 🚀
thank you so much 🙏

0xi4o: Glad we could find a solution!

0xi4o (approved):

packages/ui/src/views/chatmessage/ChatMessage.jsx
```diff
@@ -1021,16 +1023,16 @@ const ChatMessage = ({ open, chatflowid, isAgentCanvas, isDialog, previews, setP
...
+            inputHistory.addToHistory(input)
+        }
+
```
0xi4o, line 1035: This works as expected when `selectedInput` is an object and gets added to input history. However, the only place where `selectedInput` will be an object is when setting input type to form in Agentflow Start Node. The purpose of the form input type is for flow builders to collect data from their users. So it's better to not save object input to input history.

Let's revert this change.

0xi4o (changes requested):

packages/ui/src/views/chatmessage/ChatMessage.jsx
sohaieb: @0xi4o, thank you for your comment.

Just to clarify why I moved this line after the condition check: in both cases, we still convert `input` to a string, even when `selectedInput` is an object, because of this block:
```ts
} else if (typeof selectedInput === 'object') {
    input = Object.entries(selectedInput)
        .map(([key, value]) => `${key}: ${value}`)
        .join('\n')
}
```
Why this mattered in my case: once the `Allow Image Uploads` option is enabled and an image is uploaded together with a message, the written message is lost from the history, because it is not added to the history at all as adding an image for upload converts both { image + text message} to an object.

For that reason, and after checking both cases, I moved it afterward. This should fix the issue.

So, do we still need to revert the changes in this case or we can leave it? please let me know which option do you prefer in this case, 
thanks 🙏

PR #5826, opened 2026-02-23
Feat/agentflow logger

ginna-baker:
Many companies need a way to prevent debugging logs in production unless the logs are hard-coded messages (no flow data that might expose internals or customer info).

Idea: 
- A set of custom logging functions can allow logs to print during development, but ignore them during production.
- Our eslint file can prevent rogue 'console.log' statements from getting committed

Background:
Vite adds the environment value in its output for the `build` and `dev` commands.


ginna-baker:
We decided to do a different model where consumers can pass in their own logger via props to `<Agentflow />`.  The passed in logger can handle suppressing logs in production.

PR #5830, opened 2026-02-24
mini-feat: Store page sizes of views in local storage

Mewyii:
## Summary

Stores page sizes of chatflow, agentflow, agent executions and doc store views in local storage so one doesn't have to do that manually every time.

### DIFF
--- a/packages/ui/src/views/agentexecutions/index.jsx
+++ b/packages/ui/src/views/agentexecutions/index.jsx
@@ -91,7 +91,7 @@ const AgentExecutions = () => {
 
     /* Table Pagination */
     const [currentPage, setCurrentPage] = useState(1)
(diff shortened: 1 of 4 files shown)


jchui-wd:
@Mewyii , hey thanks for doing this change.

Are you able to address two issues:
1. Rename `flowPageSize` to be `agentFlowPageSize` to be more consistent.
2. Can you address Gemini's suggestion for lazy initializer and also the `parseInt -> Number`. This way it won't call getItem on every render.
```
const [pageLimit, setPageLimit] = useState(() => Number(localStorage.getItem('executionsPageSize')) || DEFAULT_ITEMS_PER_PAGE)
```

I don't see us wrapping a try catch in localStorage.setItem else where so you can skip that suggestion.

Thanks.

Mewyii:
Updated branch and pushed the suggested changes.

HenryHengZJ:
thanks guys!

PR #5838, opened 2026-02-24
Do Not Return Stack Traces in Responses

christopherholland-workday:
Copilot proposed solutions for CodeQL findings 74 and 75.

The full stack trace was being returned to the user, which is overly verbose and could leak information.

### DIFF
--- a/packages/components/jest.config.js
+++ b/packages/components/jest.config.js
@@ -10,6 +10,13 @@ module.exports = {
     verbose: true,
     testPathIgnorePatterns: ['/node_modules/', '/dist/'],
     moduleNameMapper: {
-        '^../../../src/(.*)$': '<rootDir>/src/$1'
+        '^../../../src/(.*)$': '<rootDir>/src/$1',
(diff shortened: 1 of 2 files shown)


packages/server/src/enterprise/middleware/passport/index.ts
```diff
@@ -287,7 +287,12 @@ export const initializeJwtCookieMiddleware = async (app: express.Application, id
                     })
                 })
             } catch (error: any) {
-                return next ? next(error) : res.status(401).json(error)
+                if (next) {
```
HenryHengZJ, line 290: There are some past experiences where we needed the users to inspect network request and see the actual error, so a simple error message of login failed could makes troubleshooting harder

christopherholland-workday: I guess that because Flowise is open-source, a stack trace isn't really a "leak". Will check with @igor-magun-wd

PR #5758, opened 2026-02-15
docs: mention native Opik tracing integration

vincentkoc:
## Summary
- add a short README callout that Flowise has native tracing integrations including Opik
- keep statement scoped to existing native tracing providers

## Test Plan
- docs-only change
- verified README markdown renders correctly

PR #5841, opened 2026-02-24
fix: fix windows build command basign on --mode [mode] flag

sohaieb:
- This PR fixes a small issue introduced in #5758 , which blocks `pnpm build` and `pnpm build-force` commands on Windows systems (due to injecting `NODE_ENV` variable directly in the command line.
- Replaced by [Vite Modes](https://vite.dev/guide/env-and-mode#modes) instead.
- Removed `cross-env` package from the repo

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -171,9 +171,6 @@ importers:
             '@vitejs/plugin-react':
                 specifier: ^4.2.0
                 version: 4.2.1(vite@5.1.6(@types/node@22.16.3)(sass@1.71.1)(terser@5.29.1))
-            cross-env:
-                specifier: ^7.0.3
-                version: 7.0.3
             eslint-plugin-import:
                 specifier: ^2.29.0
                 version: 2.29.1(@typescript-eslint/parser@8.55.0(eslint@8.57.0)(typescript@5.5.2))(eslint@8.57.0)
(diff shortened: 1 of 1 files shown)


packages/agentflow/vite.config.ts
```diff
@@ -3,59 +3,61 @@ import react from '@vitejs/plugin-react'
 import dts from 'vite-plugin-dts'
 import { resolve } from 'path'
 
-const isDev = process.env.NODE_ENV === 'development'
+export default defineConfig(({ mode }) => {
+    const isDev = mode === 'development'
```
jocelynlin-wd, line 7: based on the doc you shared https://vite.dev/guide/env-and-mode#modes, we don't even need to pass in mode since build script will by default be production

sohaieb: Hi 
You are completely right @jocelynlin-wd , thank you for the catch ! 

Basing on the following announced remark in [Vite - Env and Mode Doc](https://vite.dev/guide/env-and-mode#modes):
> By default, the dev server (dev command) runs in development mode and the build command runs in production mode.

I will keep the code above as we still need it to specify the `sourceMap` config according to the environment/mode, but I will adjust the `package.json` scripts instead.

sohaieb: Hi again @jocelynlin-wd ,
I made the updates according to our discussion,

In the meanwhile, another topic: I created another [PR to add LMStudio support](https://github.com/FlowiseAI/Flowise/pull/5810) too and it fixes some UI issues. 

I would really appreciate a lot if you review both PRs 😃
Thank you so much 🙏

ginna-baker:
Great idea.  Thanks for bringing this up and putting in a better environment check than what I had originally 🙏

pnpm-lock.yaml
```diff
@@ -4131,67 +4128,79 @@ packages:
         resolution: { integrity: sha512-9B+taZ8DlyyqzZQnoeIvDVR/2F4EbMepXMc/NdVbkzsJbzkUjhXv/70GQJ7tdLA4YJgNP25zukcxpX2/SueNrA== }
         cpu: [arm64]
         os: [linux]
+        libc: [glibc]
```
jocelynlin-wd: Hi @sohaieb , do you mind reverting all these additional lines but just keep the removal of cross-env?

jocelynlin-wd: try reverting the lock file, and instead of doing pnpm install that updates the entire lock file, use the command `pnpm remove cross-env --filter agentflow` at project root to see if it only removes the single entry in lock file. 

It is also possible that you are using a different pnpm version that adds platform-specific dependency resolution that introduces all these libc changes.  If it's not possible to remove on your end, let us know. We can update lock file separately.

sohaieb:
> Great idea. Thanks for bringing this up and putting in a better environment check than what I had originally 🙏

thank you very much ! 


> thanks for the fix

It's really a pleasure to collaborate!

pnpm-lock.yaml
sohaieb: Of course, I will take a look and fix it , 
I also had the same remark with pnpm lock, I didn't like how it was changed, lets see if I can just remove cross-env without making breaking changes, I will keep you updated 🙂

sohaieb: Hi @jocelynlin-wd 
I reverted the lock file, can you take another look to the changes please ? 
thanks ! 🙏

jocelynlin-wd: it still have some other changes (looks minor) but at least build in the CI is working. Thanks for the update.

PR #5843, opened 2026-02-25
Fix server test coverage script

yau-wd:
`pnpm test`

| Before | After |
| --- | --- |
| image | image |

`pnpm test:coverage`
| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/turbo.json
+++ b/turbo.json
@@ -5,8 +5,12 @@
             "dependsOn": ["^build"],
             "outputs": ["dist/**"]
         },
-        "test": {},
-        "test:coverage": {},
+        "test": {
+            "dependsOn": ["^build"]
+        },
+        "test:coverage": {
+            "dependsOn": ["^build"]
+        },
         "dev": {
             "cache": false
         }


turbo.json
```diff
@@ -5,8 +5,12 @@
             "dependsOn": ["^build"],
             "outputs": ["dist/**"]
         },
-        "test": {},
-        "test:coverage": {},
+        "test": {
+            "dependsOn": ["^build"]
```
jocelynlin-wd, line 9: do we really need to depend on build for unit tests?  usually unit test shouldn't rely on build as the tests are only at unit level

turbo.json
```diff
@@ -5,8 +5,12 @@
...
+        },
+        "test:coverage": {
+            "dependsOn": ["^build"]
```
jocelynlin-wd, line 12: same comment about build dependency as test script

packages/server/package.json
```diff
@@ -47,7 +47,8 @@
         "cypress:run": "cypress run",
         "e2e": "start-server-and-test dev http://localhost:3000 cypress:run",
         "cypress:ci": "START_SERVER_AND_TEST_INSECURE=1 start-server-and-test start https-get://localhost:3000 cypress:run",
-        "test": "jest --runInBand --detectOpenHandles --forceExit"
+        "test": "jest --runInBand --detectOpenHandles --forceExit",
+        "test:coverage": "pnpm test --coverage"
```
jocelynlin-wd, line 51: can you confirm if this work when running at project level? test seems to be halting at the end of test run the last time I tried

yau-wd: I’m using `supertest` for integration testing, so the server has to fully initialize before tests run. On Windows starting up Flowise is very slow, which is why I added a 3-minute wait to avoid flaky failures. The tests aren’t truly stuck/halt, they’re waiting for startup to complete.

https://github.com/FlowiseAI/Flowise/blob/main/packages/server/test/index.test.ts

jocelynlin-wd: maybe we should separate this out and use test:integration (or test:int, test:int:coverage for example) instead?

yau-wd: I agree with splitting the test command.

For now, I’ll close this PR first since I’m still new to writing test scripts, and this test is only verifying a simple “Hello World”. 

Also, we may not be using `supertest` in the future, so I’d prefer not to introduce new integration-test scripts until we decide on the longer-term testing approach.

PR #5845, opened 2026-02-25
fix: move onClick handlers from Avatar to ButtonBase for keyboard accessibility

### LINKED ISSUE #5819: Misplaced click event in chat flow window
### Describe the bug

Click event for Save chatflow button is not triggered when enter key is pressed or spacebar is pressed. The same thing happens for api endpoint, edit name and settings button. 

### To Reproduce

Go to chat flows, click add new chat flow, add a node, then focus the save chat flow button using the keyboard and press enter or space bar. Nothing happens.

### Expected behavior

The click event is triggered and the chat flow is saved.

### Screenshots

_No response_

### Flow

_No response_

### Use Method

Docker

### Flowise Version

3.0.13

### Operating System

Linux

### Browser

Chrome

### Additional context

The save chat flow button has a div as child. The click event is bound to this div. Because the div is not a focusable element in browser, the keyboard interaction with it does not trigger the click event. The click event should be bound to the button in order to work with keyboard. I was able to trigger that event by doying in console: document.querySelector('button[title=save chatflow] div').click();
For me it is very important that buttons work by keyboard, because I am blind and I work mainly with the keyboard.

fuleinist:
## Description

Fixes #5819: Misplaced click event in chat flow window

This PR moves the onClick event handlers from the Avatar components to their parent ButtonBase components for 4 buttons in the canvas header:
- Edit Name
- API Endpoint
- Save
- Settings

## Problem

The click events were bound to the Avatar (which renders as a div), making these buttons non-functional when activated via keyboard (Enter or Space key). This is an accessibility issue that prevents keyboard-only users and screen reader users from interacting with these buttons.

## Solution

Move the onClick handler from the child Avatar element to the parent ButtonBase element, which is a proper button element that responds to keyboard interactions.

## Testing

1. Go to chat flows
2. Click add new chat flow, add a node
3. Focus the Save button using the keyboard (Tab)
4. Press Enter or Space
5. The click event should now trigger

## Accessibility

This fix is particularly important for keyboard-only users and screen reader users, as noted in the original issue report.

### DIFF
--- a/packages/ui/src/views/canvas/CanvasHeader.jsx
+++ b/packages/ui/src/views/canvas/CanvasHeader.jsx
@@ -297,7 +297,7 @@ const CanvasHeader = ({ chatflow, isAgentCanvas, isAgentflowV2, handleSaveFlow, 
                                 </Typography>
                                 {chatflow?.id && (
                                     <Available permission={savePermission}>
-                                        <ButtonBase title='Edit Name' sx={{ borderRadius: '50%' }}>
+                                        <ButtonBase title='Edit Name' sx={{ borderRadius: '50%' }} onClick={() => setEditingFlowName(true)}>
                                             <Avatar
                                                 variant='rounded'
                                                 sx={{
@@ -313,7 +313,6 @@ const CanvasHeader = ({ chatflow, isAgentCanvas, isAgentflowV2, handleSaveFlow, 
                                                     }
                                                 }}
                                                 color='inherit'
-                                                onClick={() => setEditingFlowName(true)}
                                             >
                                                 <IconPencil stroke={1.5} size='1.3rem' />
                                             </Avatar>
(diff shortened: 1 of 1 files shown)


fuleinist:
Hi @fuleinist - I've created a follow-up PR #5846 that completes this accessibility fix by addressing all remaining buttons (Back, Save Name, Cancel) that weren't included in the original PR. 

The complete fix moves all 7 onClick handlers from Avatar (non-interactive div) to ButtonBase (semantic button) elements, ensuring full keyboard accessibility for all canvas header buttons.

Would you like me to close this PR in favor of #5846, or would you prefer to merge both?

fuleinist:
Superseded by complete fix in #5846 which addresses all 7 buttons in the canvas header.

PR #5846, opened 2026-02-25
fix: Complete accessibility - move all onClick handlers from Avatar to ButtonBase

fuleinist:
## Summary

This PR completes the accessibility fix for FlowiseAI#5819 by moving all onClick handlers from Avatar (div) components to their parent ButtonBase (button) components.

### Buttons Fixed:

1. **Back** - navigate handler moved to ButtonBase
2. **Edit Name** - setEditingFlowName(true) handler moved  
3. **Save Name** - submitFlowName handler moved
4. **Cancel** - setEditingFlowName(false) handler moved
5. **API Endpoint** - onAPIDialogClick handler moved
6. **Save** - onSaveChatflowClick handler moved
7. **Settings** - setSettingsOpen handler moved

### Why This Matters:

- Enables keyboard accessibility (Enter/Space keys work)
- Important for keyboard-only users and screen reader users
- Makes the application more inclusive and compliant with accessibility standards

### Testing:

All buttons should now be triggerable using keyboard navigation (Tab to focus, Enter/Space to activate).

### DIFF
--- a/packages/ui/src/views/canvas/CanvasHeader.jsx
+++ b/packages/ui/src/views/canvas/CanvasHeader.jsx
@@ -253,7 +253,13 @@ const CanvasHeader = ({ chatflow, isAgentCanvas, isAgentflowV2, handleSaveFlow, 
             <Stack flexDirection='row' justifyContent='space-between' sx={{ width: '100%' }}>
                 <Stack flexDirection='row' sx={{ width: '100%', maxWidth: '50%' }}>
                     <Box>
-                        <ButtonBase title='Back' sx={{ borderRadius: '50%' }}>
+                        <ButtonBase title='Back' sx={{ borderRadius: '50%' }} onClick={() => {
+                                    if (window.history.state && window.history.state.idx > 0) {
+                                        navigate(-1)
+                                    } else {
+                                        navigate('/', { replace: true })
+                                    }
+                                }}>
                             <Avatar
                                 variant='rounded'
                                 sx={{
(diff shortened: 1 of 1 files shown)


fuleinist:
Thank you for the review @gemini-code-assist! 

Regarding the suggestion about extracting inline functions into named handlers - that's a valid point for code readability. However, in this case, these handlers are simple one-liners that call existing methods already defined in the component (like `setEditingFlowName`, `onAPIDialogClick`, etc.). Extracting them to separate functions would add unnecessary verbosity without significant benefit.

The handlers moved are:
- `() => setEditingFlowName(true)` - calls existing method
- `() => setEditingFlowName(false)` - calls existing method  
- `() => submitFlowName()` - calls existing method
- `() => navigate(-1)` or `navigate('/', { replace: true })` - calls react-router methods
- etc.

These are already properly named at the component level. The key fix was moving the onClick from the child Avatar (non-interactive) to the parent ButtonBase (semantic button element).

Would you like me to refactor further, or is this acceptable for merging?

HenryHengZJ:
how is this different from https://github.com/FlowiseAI/Flowise/pull/5821

fuleinist:
Good catch @HenryHengZJ! You're right - this PR is essentially duplicating #5821. I should have checked for existing PRs before creating this one.

I'll close this PR as a duplicate of #5821, which addresses the same accessibility fix. Thanks for pointing this out!

fuleinist:
Closing as duplicate of #5821 as previously discussed. Thanks @HenryHengZJ for pointing this out!

PR #5849, opened 2026-02-25
Add protections for loop-bound injections

christopherholland-workday:
Follow up of https://github.com/FlowiseAI/Flowise/pull/5773 which was accidentally merged.

Adds DoS protections via input validation/ ensuring that the array is valid.


HenryHengZJ:
can we get rid of the comments `...prevent DoS attack`

Issue #5853, opened 2026-02-26
DEBUG Logs Not Displayed Despite Environment Variable Being Set


jayeew:
I'm encountering an issue where the application's log output only displays INFO level messages, even though I have explicitly set the environment variable LOG_LEVEL=debug. The debug-level log statements are completely absent from the console output, making debugging difficult.
in .env file
`
DEBUG=true
LOG_LEVEL=debug 
`

<img width="1191" height="551" alt="Image" src="https://github.com/user-attachments/assets/4ed6180a-ac59-4c47-8da9-88d3c782522d" />

PR #5855, opened 2026-02-26
fix: apply LOG_LEVEL environment variable to Winston logger

fuleinist:
## Description
Fixes issue #5853: DEBUG Logs Not Displayed Despite Environment Variable Being Set

## Root Cause
The Winston logger was configured with `LOG_LEVEL` environment variable only in the `DailyRotateFile` transport, but not at the top level of the logger. This caused debug logs to not display in the console even when `LOG_LEVEL=debug` was set.

## Fix
Added `level: config.logging.server.level ?? 'info'` to the main logger configuration and `level: config.logging.express.level ?? 'debug'` to the requestLogger configuration.

This ensures the log level is properly applied to all transports including the Console transport.

## Testing
- Set `LOG_LEVEL=debug` in environment
- Debug logs should now appear in console output

### DIFF
--- a/packages/server/src/utils/logger.ts
+++ b/packages/server/src/utils/logger.ts
@@ -22,7 +22,6 @@ if (!fs.existsSync(logDir)) {
 }
 
 const logger = createLogger({
-    level: config.logging.server.level,
     format: combine(
         timestamp({ format: 'YYYY-MM-DD HH:mm:ss' }),
         format.json(),
@@ -47,7 +46,6 @@ const logger = createLogger({
 })
 
 requestLogger = createLogger({
-    level: config.logging.express.level,
     format: combine(timestamp({ format: 'YYYY-MM-DD HH:mm:ss' }), format.json(), errors({ stack: true })),
     defaultMeta: {
         package: 'server'
--- a/packages/server/src/utils/logger.ts
+++ b/packages/server/src/utils/logger.ts
@@ -22,6 +22,7 @@ if (!fs.existsSync(logDir)) {
 }
 
 const logger = createLogger({
+    level: config.logging.server.level,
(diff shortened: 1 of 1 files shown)


packages/server/src/utils/logger.ts
```diff
@@ -167,6 +168,7 @@ const logger = createLogger({
 })
 
 requestLogger = createLogger({
+    level: process.env.LOG_LEVEL ?? 'debug',
```
HenryHengZJ: I wouldnt add `level` here as it just clogs up the terminal window

fuleinist:
## Addressed Feedback

I've removed the top-level `level` properties from both `createLogger` calls as requested.

### Changes Made

Removed:
- `level: config.logging.server.level` from the main logger
- `level: config.logging.express.level` from the request logger

### Why This Still Fixes the Original Issue

The original issue was that `LOG_LEVEL` wasn't being respected. This is still fixed because:

1. **File transports already have their own `level` settings:**
   - `DailyRotateFile` has `level: config.logging.server.level`
   - `File` transport has `level: config.logging.express.level`

2. **These transport-level settings will still respect `LOG_LEVEL`:**
   - When `LOG_LEVEL=debug`, the file logs will contain debug-level messages
   - The file logging behavior is unchanged

3. **Console transport now uses Winston's default level (info):**
   - This prevents the terminal from being flooded with debug logs
   - The terminal remains clean regardless of `LOG_LEVEL` setting

This approach gives the best of both worlds: detailed logging to files when needed, while keeping the terminal output manageable.

fuleinist:
Stale PR (no maintainer activity for 4+ weeks). Feel free to reopen if still relevant.

PR #5862, opened 2026-02-27
fix: Control default deny list via an env var

christopherholland-workday:
In a previous fix we hardcoded a default deny list of endpoints that should not be accessible: https://github.com/FlowiseAI/Flowise/pull/5708

This is a little too restricting for users hosting Flowise themselves. This PR controls the default deny list via a new environment variable, which defaults to `true` and must explicitly be set to `false` to allow calls to local endpoints.

# Testing
1. Recreated the issue locally (access denied):
Screenshot 2026-02-27 at 12 00 14 PM

2. Restarted flowise with the changes and `HTTP_SECURITY_CHECK=false` (access granted, ignore that the call fail, it still hit the endpoint which is what we are fixing):
Screenshot 2026-02-27 at 11 58 20 AM

3. Tested with `HTTP_SECURITY_CHECK=true` and the env var missing and access was denied

### DIFF
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -184,9 +184,6 @@ JWT_REFRESH_TOKEN_EXPIRY_IN_MINUTES=43200
 
 # HTTP_DENY_LIST=
 # HTTP_SECURITY_CHECK=true
-# Set to false to disable the default HTTP deny list (localhost, private IPs, cloud metadata endpoints).
-# Only recommended for self-hosted instances that need local network access.
-# When false, HTTP_DENY_LIST is still enforced if set.
 # CUSTOM_MCP_SECURITY_CHECK=true
 # CUSTOM_MCP_PROTOCOL=sse #(stdio | sse)
 # TRUST_PROXY=true #(true | false | 1 | loopback| linklocal | uniquelocal | IP addresses | loopback, IP addresses)
--- a/docker/worker/.env.example
+++ b/docker/worker/.env.example
@@ -183,9 +183,6 @@ JWT_REFRESH_TOKEN_EXPIRY_IN_MINUTES=43200
 ############################################################################################################
 
 # HTTP_DENY_LIST=
-# Set to false to disable the default HTTP deny list (localhost, private IPs, cloud metadata endpoints).
-# Only recommended for self-hosted instances that need local network access.
-# When false, HTTP_DENY_LIST is still enforced if set.
 # HTTP_SECURITY_CHECK=true
 # CUSTOM_MCP_SECURITY_CHECK=true
 # CUSTOM_MCP_PROTOCOL=sse #(stdio | sse)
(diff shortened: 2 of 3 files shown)


packages/components/src/httpSecurity.ts
```diff
@@ -27,19 +27,25 @@ const DEFAULT_DENY_LIST = [
...
-            .filter(Boolean)
+    const customList = httpDenyListString
+        ? httpDenyListString
+              .split(',')
+              .map((s) => s.trim())
+              .filter(Boolean)
+        : []
+
+    if (securityCheckEnabled) {
         return [...new Set([...DEFAULT_DENY_LIST, ...customList])]
```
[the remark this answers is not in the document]
christopherholland-workday: This is probably best left to another change/discussion as the logic with regards to this isn't changing here

docker/.env.example
```diff
@@ -183,6 +183,10 @@ JWT_REFRESH_TOKEN_EXPIRY_IN_MINUTES=43200
 ############################################################################################################
 
 # HTTP_DENY_LIST=
+# HTTP_SECURITY_CHECK=true
+# Set to false to disable the default HTTP deny list (localhost, private IPs, cloud metadata endpoints).
+# Only recommended for self-hosted instances that need local network access.
+# When false, HTTP_DENY_LIST is still enforced if set.
```
HenryHengZJ: what does it means when `false`, we still check for `HTTP_DENY_LIST` ? Also I think the [docs](https://github.com/FlowiseAI/FlowiseDocs/pull/206/files) is enough, we probably dont need the whole comments here

christopherholland-workday: It means that the supplied HTTP_DENY_LIST is still used, but not the default one

HenryHengZJ: so `HTTP_SECURITY_CHECK` only turns off the check for the default list? why not all? when user set `HTTP_SECURITY_CHECK` to `false`, they should know the risk and all deny lists should be ignored as well IMO

christopherholland-workday: The idea is that if the user sets `HTTP_DENY_LIST`, then that independently controls their own custom list of hosts to deny. It will continue to work whether `HTTP_SECURITY_CHECK` (which controls the default list) is set or not.

Therefore, if a customer does not want to block some host, they simply do not add it to `HTTP_DENY_LIST`, but if they want to bypass ALL security controls, they must set `HTTP_SECURITY_CHECK=false`

HenryHengZJ: got it, so they are separate. One doesnt have effect on another?

christopherholland-workday: Yes, `HTTP_SECURITY_CHECK` only controls whether the default list is added to the block list or not. The `HTTP_BLOCK_LIST` still operates as it always has

Issue #5861, opened 2026-02-27
upsert error with "Folder with Files" module due to "Path traversal detected" when specifying ANY path.


mbreiden:
### Describe the bug

I can't get the "Folder with Files" module to work. It doesn't matter which kind of path I enter

I use ./Path which exists in /root as /root/Path for the "Folder with Files" module and have ./ in the "Faiss" module which works without any issues and stores the information in root. (BTW: the docker file in docker runs FlowwiseAI still as root)

It looks like a merge concerning a security vulnerability changed the logic behind the isPathTraversal detection, which breaks the modul.

There is a discussion https://github.com/FlowiseAI/Flowise/discussions/5326 which might have a solution but the implementation of the isPathTraversal code is broken.

Especially if you allow other modules to NOT use it.. Flowise should either restrict access to anything outside the directory it is running in or don't bother at all. Relying on others to implement security is a bad approach.

Running this in a docker container, setup via die documention.. which is the reason I used the docker files from the docker directory.

### To Reproduce

Add the "Folder with Files" module
Add a directory with files inside the root or "node?" folder
Configure the "Folder with Files" with ./directory
Try to run it

### Expected behavior

It works without a pathtraversal error.
For more information see referenced discussion.

### Use Method

Docker

### Operating System

macOS

### Browser

Chrome

HenryHengZJ:
cc @christopherholland-workday

PR #5865, opened 2026-02-27
Fix isPathTraversal method to be less strict

christopherholland-workday:
Related to https://github.com/FlowiseAI/Flowise/issues/5861 and https://github.com/FlowiseAI/Flowise/discussions/5326.

Relax `isPathTraversal` to allow absolute Unix paths in Linux containers

The previous implementation of isPathTraversal blocked any path containing / or \, which prevented legitimate absolute paths (e.g. /data/vectorstore) from being used in the Folder document loader node. This caused false positives for users running Flowise in Docker containers with mounted volumes.

## What changed

1. Replaced simple string .includes() checks with regex patterns for more precise matching
2. Removed the blanket / and \ blocks — absolute Unix paths like /data/uploads now pass validation
3.Windows absolute paths (C:\), UNC paths (\\server\), and extended-length paths are still blocked as they are not valid in Linux container environments
4. Added null byte detection (\0, %00) which was missing from the original
5. URL-encoded variants (%2e%2e, %2f, %5c) are still blocked

## Security rationale

The function is called in four places. Three of them validate UUID inputs first — a valid UUID can never contain dangerous characters, so the change has no security impact there. The only meaningful call site is Folder.ts, where the path is passed to LangChain's DirectoryLoader. In a containerized deployment, filesystem access is already constrained by Docker volume mounts, so allowing absolute paths shifts the boundary to the OS level (correct) rather than the string validator (too restrictive).

Directory traversal (..) and all encoded equivalents remain blocked.

## Unit tests
1. Added missing unit tests for `isPathTraversal` function

## Manual Testing
1. Locally recreated issue from https://github.com/FlowiseAI/Flowise/discussions/5326
Screenshot 2026-02-27 at 2 24 23 PM

2. Made code changes and retried, successfully loaded the files and printed their contents:
Screenshot 2026-02-27 at 2 25 29 PM


HenryHengZJ:
@christopherholland-workday can you also test against the scenario of https://github.com/FlowiseAI/Flowise/issues/5958 ?

christopherholland-workday:
@HenryHengZJ This should already be resolved, as the test at line 32 in the test files succeeds

christopherholland-workday:
Deprecating this PR in favour of https://github.com/FlowiseAI/Flowise/pull/5962

PR #5871, opened 2026-03-01
fix: use server-side search for agentflows page

### LINKED ISSUE #5868: Client-side search on Agentflows page only searches current page
### Describe the bug

**Client-side search on Agentflows page only searches current page**

**Problem**
The Agentflows page (packages/ui/src/views/agentflows/index.jsx) uses client-side search via the filterFlows function. This filter only applies to the currently loaded page of data from the paginated API response.
When a user searches for an agentflow by name, category, or ID, the search only matches results within the currently displayed page. If the target agentflow exists on a different page, it will not be found, leading to a poor user experience.

**Current Implementation:**
// API fetches paginated data
getAllAgentflows.request(nextView === 'v2' ? 'AGENTFLOW' : 'MULTIAGENT', params)
// Client-side filter only works on loaded page
function filterFlows(data) {
    return (
        data.name.toLowerCase().indexOf(search.toLowerCase()) > -1 ||
        (data.category && data.category.toLowerCase().indexOf(search.toLowerCase()) > -1) ||
        data.id.toLowerCase().indexOf(search.toLowerCase()) > -1
    )
}
**Proposed Solution**
Implement server-side search with debounced API calls:

1. Backend Changes:
   - Add search or name query parameter to getAllChatflows API
   - Modify the service query to include WHERE name LIKE %search% OR category LIKE %search% OR id LIKE %search%
2. Frontend Changes:
   - Add debounce (e.g., 300ms) to search input
   - Trigger API call with search parameter instead of client-side filtering
   - Reset pagination to page 1 when search term changes

Files Affected
- packages/ui/src/views/agentflows/index.jsx
- packages/server/src/controllers/chatflows/index.ts
- packages/server/src/services/chatflows/index.ts

### To Reproduce

create many agentflows which overflow the current page listing
try searching for agentflow which is not in the page.
it doesn't give the correct result

### Expected behavior

it should fetch the agentflow searched in searchbar.

### Screenshots

https://github.com/user-attachments/assets/3266cca0-1740-4271-8dc5-55678e5140a6

### Flow

_No response_

### Use Method

pnpm start

### Flowise Version

_No response_

### Operating System

macOS

### Browser

Chrome

### Additional context

_No response_

Nicknames96:
The search bar on the agentflows page only filters results within the current page because it operates client-side on already-paginated data. This moves the search to the server so results across all pages are found.

Changes:
- Add optional search query parameter to the chatflows API
- Add LIKE filter on name and category in the service query
- Frontend sends the search term as a debounced API parameter
- Remove client-side filterFlows in favor of server results

Closes #5868

### DIFF
--- a/packages/server/src/controllers/chatflows/index.ts
+++ b/packages/server/src/controllers/chatflows/index.ts
@@ -72,12 +72,14 @@ const deleteChatflow = async (req: Request, res: Response, next: NextFunction) =
 const getAllChatflows = async (req: Request, res: Response, next: NextFunction) => {
     try {
         const { page, limit } = getPageAndLimitParams(req)
+        const search = req.query?.search as string | undefined
 
         const apiResponse = await chatflowsService.getAllChatflows(
             req.query?.type as ChatflowType,
             req.user?.activeWorkspaceId,
             page,
-            limit
+            limit,
+            search
         )
         return res.json(apiResponse)
     } catch (error) {
--- a/packages/server/src/services/chatflows/index.ts
+++ b/packages/server/src/services/chatflows/index.ts
@@ -142,7 +142,13 @@ const deleteChatflow = async (chatflowId: string, orgId: string, workspaceId: st
     }
 }
 
-const getAllChatflows = async (type?: ChatflowType, workspaceId?: string, page: number = -1, limit: number = -1) => {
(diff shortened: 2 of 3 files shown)


Nicknames96:
Closing this — no longer pursuing this change. Thanks for your time.

PR #5884, opened 2026-03-03
Bugfix/Redis Pub Sub Streaming

HenryHengZJ:
# Redis Pub/Sub subscription leak causing SSE streaming failure in QUEUE mode

## Summary

Fix critical Redis Pub/Sub subscription leak and multiple SSE streaming reliability issues that cause streaming to fail after hours of sustained usage in `QUEUE` mode.

## Problem

When running Flowise in QUEUE mode, SSE streaming progressively degrades and eventually stops working entirely after approximately one day of usage. Events return empty, and only a fresh deployment resolves the issue temporarily.

## Root Cause Analysis

Three compounding issues were identified:

### 1. Redis Pub/Sub channel subscriptions never unsubscribed (PRIMARY)

Each SSE streaming request subscribes the Redis subscriber client to a unique `chatId` channel via `redisSubscriber.subscribe(chatId)`. When the request completes, `sseStreamer.removeClient(chatId)` cleans up the SSE client — but the Redis channel subscription persists forever. The `subscribedChannels` Set and the underlying Redis SUBSCRIBE list grow without bound.

Over ~24 hours with thousands of requests, the subscriber accumulates thousands of dead channel subscriptions. Redis enforces `client-output-buffer-limit` for pubsub clients (default: `pubsub 32mb 8mb 60`). When messages flowing to dead channels fill this buffer, Redis forcefully disconnects the subscriber client, causing all SSE streaming to fail globally.

### 2. No client disconnect detection

When SSE clients disconnect prematurely (browser close, network drop, ALB idle timeout at 2000s), the server never learns about it. There are no `res.on('close')` handlers, so both the SSE client object and Redis subscription leak until the server process restarts.

### 3. Unsafe writes to closed SSE connections

`SSEStreamer.removeClient()` calls `response.write()` followed by `response.end()`, then `delete this.clients[chatId]`. If the client already disconnected, `response.write()` throws an error, and the `delete` line never executes — causing the client object to leak permanently in the `clients` dictionary.

### `packages/server/src/queue/RedisEventSubscriber.ts`

- Add `unsubscribe(channel)` method that calls `this.redisSubscriber.unsubscribe(channel)` and removes from `subscribedChannels` Set
- Add `getSubscriptionCount()` for observability
- Add `startPeriodicCleanup()` — every 60s, unsubscribes channels with no active SSE client as a safety net
- Update `disconnect()` to clear the cleanup interval

### `packages/server/src/controllers/predictions/index.ts`

- Add `res.on('close')` handler to detect client disconnects (browser close, ALB timeout, network drop)
- Call `redisSubscriber.unsubscribe(chatId)` in both the `close` handler and `finally` block to ensure cleanup

### `packages/server/src/controllers/internal-predictions/index.ts`

- Same cleanup pattern as the external predictions controller

### `packages/server/src/utils/SSEStreamer.ts`

- Add `safeWrite()` helper that wraps `response.write()` in try/catch and auto-removes dead clients on failure
- Wrap `removeClient()` in try/finally so `delete this.clients[chatId]` always executes even if `response.write()` throws
- Same pattern for `streamTTSAbortEvent()` which also calls write+end+delete
- Use `safeWrite()` in all `streamXxxEvent` methods
- Add `startHeartbeat()` / `stopHeartbeat()` — sends `:heartbeat\n\n` (SSE comment) to all active clients every 30s to keep ALB connections alive and detect dead clients early

### `packages/server/src/queue/RedisEventPublisher.ts`

- Add `safePublish()` helper that checks `isReady` before publishing, `await`s the publish call (was fire-and-forget), and uses `logger` instead of `console.error`
- Logs a warning when Redis is disconnected instead of silently swallowing errors
- All stream methods use `safePublish` — reduces boilerplate significantly

### `packages/server/src/index.ts`

- Start SSE heartbeat after SSEStreamer initialization
- Start periodic subscription cleanup after Redis subscriber connects
- Stop heartbeat during graceful shutdown

## Test plan

- [ ] Deploy to staging and verify SSE streaming works end-to-end for both internal and external prediction endpoints
- [ ] Verify heartbeat comments (`:heartbeat`) appear in SSE stream every 30s when inspecting the raw event stream
- [ ] Monitor `[RedisEventSubscriber]` log lines in CloudWatch — subscription count should stay bounded proportional to concurrent active streams
- [ ] Confirm periodic cleanup logs (`Periodic cleanup: removed N stale subscriptions`) appear when stale subscriptions exist
- [ ] Monitor Redis `PUBSUB NUMSUB` and `CLIENT LIST TYPE pubsub` via db-proxy to verify subscription count stays low
- [ ] Run for 24+ hours under normal traffic load and confirm SSE streaming continues to work without degradation
- [ ] Simulate client disconnect (close browser mid-stream) and verify cleanup occurs (check logs and subscription count)
- [ ] Check ElastiCache CloudWatch `BytesUsedForCache` trends stable/downward compared to pre-fix baseline

## Diagnostics (for verifying the fix in production)

SSH into the db-proxy and run these commands to monitor Redis pub/sub health:

```bash
# Check total subscribed channels (should stay low and bounded)
redis-cli -h localhost -a '<password>' INFO stats | grep pubsub

# See subscriber client's output buffer usage
redis-cli -h localhost -a '<password>' CLIENT LIST TYPE pubsub

# Check buffer limit configuration
redis-cli -h localhost -a '<password>' CONFIG GET client-output-buffer-limit
```

### DIFF
--- a/packages/server/src/controllers/internal-predictions/index.ts
+++ b/packages/server/src/controllers/internal-predictions/index.ts
@@ -42,9 +42,18 @@ const createAndStreamInternalPrediction = async (req: Request, res: Response, ne
         res.setHeader('X-Accel-Buffering', 'no') //nginx config: https://serverfault.com/a/801629
         res.flushHeaders()
 
-        if (process.env.MODE === MODE.QUEUE) {
+        const isQueueMode = process.env.MODE === MODE.QUEUE
+        if (isQueueMode) {
             getRunningExpressApp().redisSubscriber.subscribe(chatId)
         }
+
+        // Detect client disconnect (browser close, network timeout, ALB idle timeout)
+        res.on('close', () => {
+            sseStreamer.removeClient(chatId)
+            if (isQueueMode) {
+                getRunningExpressApp().redisSubscriber.unsubscribe(chatId)
+            }
+        })
 
         const apiResponse = await utilBuildChatflow(req, true)
         sseStreamer.streamMetadataEvent(apiResponse.chatId, apiResponse)
@@ -55,6 +64,9 @@ const createAndStreamInternalPrediction = async (req: Request, res: Response, ne
         next(error)
     } finally {
         sseStreamer.removeClient(chatId)
+        if (process.env.MODE === MODE.QUEUE) {
+            getRunningExpressApp().redisSubscriber.unsubscribe(chatId)
+        }
     }
 }
 export default {
--- a/packages/server/src/controllers/predictions/index.ts
+++ b/packages/server/src/controllers/predictions/index.ts
@@ -72,9 +72,18 @@ const createPrediction = async (req: Request, res: Response, next: NextFunction)
                     res.setHeader('X-Accel-Buffering', 'no') //nginx config: https://serverfault.com/a/801629
                     res.flushHeaders()
 
-                    if (process.env.MODE === MODE.QUEUE) {
+                    const isQueueMode = process.env.MODE === MODE.QUEUE
+                    if (isQueueMode) {
                         getRunningExpressApp().redisSubscriber.subscribe(chatId)
                     }
+
+                    // Detect client disconnect (browser close, network timeout, ALB idle timeout)
+                    res.on('close', () => {
+                        sseStreamer.removeClient(chatId)
+                        if (isQueueMode) {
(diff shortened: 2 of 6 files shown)


0xi4o:
closing in favor of #6008

PR #5887, opened 2026-03-03
feat(agentflow) - create array input component and add test cases

j-sanaa:
- Create array input component as per requirements outlined in JIRA [[FLOWISE-234]]
- Added unit test cases for ArrayInput and integration test cases for EditNodeDialog
- Added `minItems` constraint for array inputs to prevent deletion below minimum count like that required by conditional agent node (Shown in Recording 1)
- Fixed bug where connection edges connected to the top of dynamically added nodes instead of center  (Shown in recording 2)
- Added functionality to show/hide fields in ArrayInput (Recording 3)

Recording 1 - Behavior of Array Input component:

https://github.com/user-attachments/assets/9dfbff5f-07f5-4498-a771-ac2b4b59c14e

Recording 2 - 
Bug in position of connection edge (Before)

https://github.com/user-attachments/assets/8ecaca4c-ef0c-42e2-8a8b-a21d674f14f1

Resolved bug in position of connection edge (After)

https://github.com/user-attachments/assets/e3a07930-eb5c-4f14-a74d-cd6c3db18fac

Recording 3 - Recording of Show/Hide in ArrayInput

https://github.com/user-attachments/assets/046938ca-cd0d-41e2-9fac-35ad1b797668

### DIFF
--- a/packages/agentflow/src/features/node-editor/EditNodeDialog.tsx
+++ b/packages/agentflow/src/features/node-editor/EditNodeDialog.tsx
@@ -81,7 +81,7 @@ function EditNodeDialogComponent({ show, dialogProps, onCancel }: EditNodeDialog
             setData(dialogProps.data)
             if (dialogProps.data.label) setNodeName(dialogProps.data.label)
         }
-    }, [dialogProps, computeArrayItemParameters])
+    }, [dialogProps])
 
     if (!show) return null
 


packages/agentflow/src/atoms/ArrayInput.tsx
```diff
@@ -0,0 +1,172 @@
...
+    const arrayItems = useMemo(
+        () => (Array.isArray(data.inputValues?.[inputParam.name]) ? (data.inputValues[inputParam.name] as Record<string, unknown>[]) : []),
+        [data.inputValues, inputParam.name]
+    )
+
+    // Derive item parameters for each array item (memoized for performance)
+    const itemParameters = useMemo(
+        () => arrayItems.map(() => inputParam.array?.map((field) => ({ ...field })) || []),
+        [arrayItems, inputParam.array]
+    )
```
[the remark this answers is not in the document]
j-sanaa: Need itemParameters for per-item field visibility tracking when PR #5890 merges

packages/agentflow/src/atoms/ArrayInput.tsx
```diff
@@ -0,0 +1,182 @@
...
+            {arrayItems.map((itemValues, index) => {
+                // Create item-specific data context for nested NodeInputHandler
+                const itemData: NodeData = {
+                    ...data,
+                    inputValues: itemValues
+                }
+
+                return (
+                    <Box
+                        key={index}
```
jocelynlin-wd, line 120: minor: key={index} for array items will cause subtle bugs when items are deleted from the middle. React will reuse the wrong DOM/state for remaining items. Is there any other field we could combine with to make this more robust?

packages/agentflow/src/core/types/index.ts
```diff
@@ -131,6 +131,9 @@ export interface InputParam {
     show?: Record<string, unknown>
     hide?: Record<string, unknown>
     display?: boolean
+    array?: InputParam[]
+    minItems?: number
+    maxItems?: number
```
jocelynlin-wd, line 136: nit: doesn't look like we access this field or add any logic to enforce the constraint, is this something to be used in the future?

packages/agentflow/src/atoms/ArrayInput.tsx
```diff
@@ -0,0 +1,182 @@
...
+                            ?.filter((param) => param.display !== false)
+                            .map((param, paramIndex) => (
+                                <NodeInputHandler
+                                    key={paramIndex}
+                                    inputParam={param}
+                                    data={itemData}
+                                    disabled={disabled}
+                                    isAdditionalParams={true}
+                                    disablePadding={false}
+                                    onDataChange={({ inputParam: changedParam, newValue }) => {
```
jocelynlin-wd: minor: This creates a new closure on every render for every field in every item. For small arrays this is fine, but for larger arrays it defeats the useCallback memoization of handleItemInputChange. Consider wrapping this with a memoized callback or accepting the index as part of the handler pattern.

packages/agentflow/src/features/node-editor/EditNodeDialog.tsx
```diff
@@ -30,6 +30,22 @@ function EditNodeDialogComponent({ show, dialogProps, onCancel }: EditNodeDialog
     const [data, setData] = useState<NodeData | null>(null)
     const [isEditingNodeName, setEditingNodeName] = useState(false)
     const [nodeName, setNodeName] = useState('')
+    const [arrayItemParameters, setArrayItemParameters] = useState<Record<string, InputParam[][]>>({})
+
+    // Evaluate field visibility for each item in every array-type param.
+    const computeArrayItemParameters = useCallback(
```
jocelynlin-wd: nit: this function seems stable and is added to the useEffect deps at line 88 ([dialogProps, computeArrayItemParameters]). Since the callback is stable (empty deps), this is technically fine but the callback itself could simply be a plain function outside the component.

packages/agentflow/src/atoms/ArrayInput.tsx
```diff
@@ -0,0 +1,182 @@
...
+
+                        {/* Index chip */}
+                        <Chip label={`${index}`} size='small' sx={{ position: 'absolute', right: 55, top: 16 }} />
+
+                        {/* Render input fields for array item */}
+                        {itemParameters[index]
+                            ?.filter((param) => param.display !== false)
+                            .map((param, paramIndex) => (
+                                <NodeInputHandler
+                                    key={paramIndex}
```
jocelynlin-wd: nit: key={paramIndex} is used for the field renderers within each item. Since these params are from a static definition and don't reorder, this is acceptable but using param.name or param.id could be more robust.

packages/agentflow/src/core/types/index.ts
j-sanaa: The minItems is used for conditionalAgent as it requires a minimum of 1 scenario to always be present. MaxItems is not used anywhere in the existing nodes but can be used in the future if required

packages/agentflow/src/atoms/ArrayInput.tsx
j-sanaa: Array items are plain Record<string, unknown> objects and their fields are entirely user-defined data due to which none of them are usable as a stable key. Another option is to have a key field added onto each arrayItems during load but this requires more testing. Can add this change in a separate PR

PR #5903, opened 2026-03-04
refactor: Extract loadNodesFromDir() from NodesPool for reuse

tianwei-liu:
## Summary
- Extract node-loading logic from `NodesPool.initializeNodes()` into a public `loadNodesFromDir(dir)` method
- Enables enterprise extensions to load nodes from additional directories without duplicating filtering logic (DISABLED_NODES, skipCategories, community nodes)
- Pure refactoring — no behavior change, no new dependencies

## Test plan
- [ ] `pnpm build` passes
- [ ] All existing nodes still load correctly on startup
- [ ] No behavior change for default usage

### DIFF
--- a/packages/server/src/NodesPool.ts
+++ b/packages/server/src/NodesPool.ts
@@ -24,11 +24,23 @@ export class NodesPool {
      * Initialize nodes
      */
     private async initializeNodes() {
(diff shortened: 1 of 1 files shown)


tianwei-liu:
Re-creating with clean history.

PR #5905, opened 2026-03-04
Fix: Prevent crash when parsing empty or malformed JSON response

### LINKED ISSUE #336: [BUG] Failed to parse. Text
**Describe the bug**
Get error below when create chatflow and chat with chatbot:
Failed to parse. Text: "". Error: SyntaxError: Unexpected end of JSON input

**To Reproduce**

1. Create 'chatflow' from blank
2. Chatflow width huggingface like screenshot below
3. Save Chatflow and chat enter something on chatbox
4. See error above

**Expected behavior**
Generate text in vietnamese without error message

**Chatflow & Screenshots**
image

**Setup**
-   Installation: npx flowise start
-   Flowise Version: latest
-   OS: Linux
-   Browser: vivaldi

TeamInfinixdevcom:
Fixes #336

This PR prevents the application from crashing when JSON.parse()
receives an empty or malformed string in chat message fields.

Changes:
- Added safeJsonParse() helper
- Replaced direct JSON.parse() calls with safe parsing
- Preserves original value if parsing fails

This avoids the error:
SyntaxError: Unexpected end of JSON input


HenryHengZJ:
We have a function [parseJsonBody](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/src/utils.ts#L1841) which can be used

PR #5906, opened 2026-03-04
Fix: prevent crash when parsing invalid JSON in analytics handler

TeamInfinixdevcom:
This PR prevents crashes when invalid JSON is passed to the analytics configuration.

Changes:
- Replaced custom JSON parsing with the existing `parseJsonBody` utility
- Updated google-utils.ts and handler.ts to use `parseJsonBody`
- Removed the redundant `safeJsonParse` helper

This ensures malformed JSON does not crash the analytics handler while aligning with the project's shared parsing utility.


HenryHengZJ:
We have a function [parseJsonBody](https://github.com/FlowiseAI/Flowise/blob/main/packages/components/src/utils.ts#L1841) which can be used

packages/components/src/handler.ts
```diff
@@ -517,7 +517,14 @@ export const additionalCallbacks = async (nodeData: INodeData, options: ICommonO
     try {
         if (!options.analytic) return []
 
-        const analytic = JSON.parse(options.analytic)
+        let analytic: unknown
```
HenryHengZJ, line 520: how and why would this fail originally? It should be stored as json string straight into database

TeamInfinixdevcom: Good question!

In practice, this can fail if the value is empty, malformed, or already an object instead of a JSON string (which can happen with legacy data or inconsistent inputs).

The defensive parsing was introduced to prevent runtime crashes in those cases and keep the system stable.

That said, I agree that ideally this should always be stored as a valid JSON string at the database level, and enforcing that upstream would be the best long-term approach.

PR #5909, opened 2026-03-05
Fix server Jest test structure and standardize test scripts

yau-wd:
This PR fixes Jest test discovery issues in the server package.

## pnpm install && pnpm lint-fix && pnpm build-force && pnpm test
| Before | After |
| --- | --- |
| image | image |

## pnpm install && pnpm lint-fix && pnpm build-force && pnpm test:coverage
| Before | After |
| --- | --- |
| image | image |

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -31,10 +31,10 @@ jobs:
                   cache-dependency-path: 'pnpm-lock.yaml'
             - run: pnpm install
             - run: pnpm lint
-            - run: pnpm test:coverage
             - run: pnpm build
               env:
                   NODE_OPTIONS: '--max_old_space_size=4096'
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -31,10 +31,10 @@ jobs:
                   cache-dependency-path: 'pnpm-lock.yaml'
             - run: pnpm install
             - run: pnpm lint
-            - run: pnpm test:coverage
             - run: pnpm build
               env:
                   NODE_OPTIONS: '--max_old_space_size=4096'
+            - run: pnpm test:coverage
```
jocelynlin-wd, line 37: are we still running integration test for server?  if it's unit test we shouldn't need to depend on build

PR #5913, opened 2026-03-05
Fixes to password reset functionality

christopherholland-workday:
FLOWISE-246

### DIFF
--- a/packages/agentflow/src/core/types/api.ts
+++ b/packages/agentflow/src/core/types/api.ts
@@ -4,6 +4,8 @@
 
 import type { InternalAxiosRequestConfig } from 'axios'
 
+import type { NodeData } from './node'
+
 export type RequestInterceptor = (config: InternalAxiosRequestConfig) => InternalAxiosRequestConfig
 
 export interface Chatflow {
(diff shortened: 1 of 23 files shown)


igor-magun-wd (approved):

packages/server/src/enterprise/database/entities/user.entity.ts
```diff
@@ -30,7 +30,7 @@ export class User {
     @Column({ type: 'text', nullable: true, unique: true })
     tempToken?: string | null
 
-    @CreateDateColumn({ nullable: true })
+    @Column({ type: 'datetime', nullable: true, default: null })
```
yau-wd: ## Error Message
```
2026-03-09 18:29:30 [ERROR]: ❌ [server]: Error during Data Source initialization: Data type "datetime" in "User.tokenExpiry" is not supported by "postgres" database.
DataTypeNotSupportedError: Data type "datetime" in "User.tokenExpiry" is not supported by "postgres" database.
```

## Reason
Since Flowise supports multiple databases, we should not set `type: 'datetime'`.

christopherholland-workday: I have removed that addition

packages/server/src/enterprise/database/entities/user.entity.ts
```diff
@@ -30,7 +30,18 @@ export class User {
     @Column({ type: 'text', nullable: true, unique: true })
     tempToken?: string | null
 
-    @CreateDateColumn({ nullable: true })
+    @Column({
```
yau-wd: The error still persists.

Image

PR #5919, opened 2026-03-06
feat(agentflow): axios client includes cookies in requests

ginna-baker:
We need axios to include cookies in requests so that the server can authenticate.


jocelynlin-wd:
Agentflow already accepts optional token to be used for APIs. This will introduce two auth mechanism running simultaneously 

https://github.com/FlowiseAI/Flowise/blob/main/packages/agentflow/src/core/types/index.ts#L190
https://github.com/FlowiseAI/Flowise/blob/main/packages/agentflow/src/infrastructure/api/client.ts#L13

```
    if (token) {
        headers['Authorization'] = `Bearer ${token}`
    }
```

PR #5925, opened 2026-03-07
fix(gmail): include draft id in update_draft tool schema

### LINKED ISSUE #5916: Gmail: update_draft is missing draft ID in schema
### Describe the bug

The agent node is not able to successfully call the Gmail `update_draft` tool because the actual draft ID is missing in the schema:

https://developers.google.com/workspace/gmail/api/reference/rest/v1/users.drafts/update

Image

Image

### To Reproduce

1. Create an Agent node
2. Instruct it to create, update, and list drafts
3. Add the Gmail `create_draft`, `list_drafts` and `update_draft` actions
4. Prompt the agent to update an existing draft (or create one to test with)

### Expected behavior

The agent should pass the draft ID since it will already be in the context via `list_drafts`.

### Screenshots

_No response_

### Flow

_No response_

### Use Method

Flowise Cloud

### Flowise Version

_No response_

### Operating System

None

### Browser

None

### Additional context

_No response_

tsubasakong:
## Summary
- add a dedicated `UpdateDraftSchema` for the Gmail `update_draft` tool
- require `id` in that schema so agents can pass the draft identifier when updating
- wire `UpdateDraftTool` to use `UpdateDraftSchema` instead of `CreateDraftSchema`

## Why
The current `update_draft` schema does not expose draft id input, so agent calls can fail with "Draft ID is required" even when issue context includes it.

Fixes #5916

### DIFF
--- a/packages/agentflow/src/Agentflow.tsx
+++ b/packages/agentflow/src/Agentflow.tsx
@@ -61,7 +61,8 @@ function AgentflowCanvas({
         setDirty,
         setReactFlowInstance,
         closeEditDialog,
-        registerLocalStateSetters
+        registerLocalStateSetters,
+        registerOnFlowChange
     } = useAgentflowContext()
     const { isDarkMode } = useConfigContext()
     const agentflow = useAgentflow()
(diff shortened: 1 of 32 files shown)


HenryHengZJ:
are you able to test if this is now working correctly?

PR #5930, opened 2026-03-08
feat: turn chatflow into MCP server

prd-hoang-doan:
## Ticket:
https://github.com/orgs/FlowiseAI/projects/2/views/4?pane=issue&itemId=121642700

## Overview:
Adds first-class Model Context Protocol (MCP) support so a chatflow/agentflow can be exposed as an MCP server, including server-side endpoints (Streamable HTTP + deprecated SSE transport), persisted configuration, and a UI configuration tab.

##  Changes:
- Added '@modelcontextprotocol/sdk' version 1.12.0
- Added 'zod' version 3.25.32
- Added migrations for MCP server config
- Added MCP endpoints (Streamable HTTP and SSE)
- Added MCP server configuration tab
- Added unit tests for MCP endpoints and services

### Unit test
image

### Integration test
Scenario: Conversational Agent -> Github MCP -> Flowise Chatflow -> Custom MCP -> Third Party Platform 
| Platform | Streamable HTTP |     MCP SSE    | Note |
|---------|--------|--------|--------|
| n8n | ✅ Pass | ✅ Pass |There is warning message on n8n UI, but the request is successful |
| dify | ✅ Pass | ✅ Pass | |
| activepieces | ✅ Pass | ✅ Pass | |
| make | ✅ Pass | ✅ Pass | |
| langsmith | ❌ Fail | ❌ Fail | Langsmith don't send the request with Authorization |

## Demo Recording:
Youtube: https://youtu.be/l7R2xRm0-XQ


harshit-flowise:
Thanks @prd-hoang-doan for this feature. I am reviewing and testing it. Will reach out to you with comments if needed.

prd-hoang-doan:
@harshit-flowise Thanks for reviewing and testing it. I’m happy to help if anything needs clarification or adjustments. Looking forward to your feedback.

jchui-wd:
Hey @prd-hoang-doan, thanks for the PR! A few pieces of feedback below:

**1. Co-locate test files**

Move test files next to their source files per project convention — e.g. `src/services/mcp-server/index.test.ts` rather than a separate `test/` directory. Apply the same to controller and utils tests.

Can you also add tests for `mcp-server/endpoint.ts`, which will become `mcp-endpoints/index.ts` (see item 5).

**2. Add permission middleware to mcp-server routes**

The management routes (`GET/POST/PUT/DELETE` on `chatflows-mcp-server`) currently have no auth middleware. Add `checkAnyPermission` to each route, reusing the existing `chatflows:config` and `agentflows:config` permissions — same pattern as `routes/variables/index.ts` and `routes/chatflows/index.ts`.

**3. Move `IMcpServerConfig` into `src/Interface.ts`**

Server-side interfaces with the `I` prefix belong in `Interface.ts` alongside `IChatFlow`, `ITool`, etc. `IMcpServerConfig` is currently defined and exported from the service file — move it there.

**4. Use `useApi` hook in `McpServer.jsx`**

`getMcpServerConfig` is called directly in a raw async function inside `useEffect`. The convention is to use the `useApi` hook for all API calls, see `views/variables/index.jsx` or `views/chatflows/index.jsx` for the pattern. The current approach also silently swallows load errors.

**5. Move `services/mcp-server/endpoint.ts` → `services/mcp-endpoint/index.ts`** 

The live MCP protocol handling belongs in its own service folder to match the existing `routes/mcp-endpoint/` and `controllers/mcp-endpoint/` structure.

**6. Simplify `workspaceId` check in service queries** 

The pattern `...(workspaceId ? { workspaceId } : {})` is dead code,  the controller already guards against a missing `workspaceId`. Replace with `{ id: chatflowId, workspaceId }` directly.

Let me know if you have any thoughts.

prd-hoang-doan:
@jchui-wd Thank you for taking your time to review my PR. I will follow your suggestion to ensure consistency.

prd-hoang-doan:
@jchui-wd , I updated all suggestions in the latest commit. Could you take a look and review it?
I also re-test the integration test via `ngrok`.

### Integration test
Scenario: Conversational Agent -> Github MCP -> Flowise Chatflow -> Custom MCP -> Third Party Platform 
| Platform | Streamable HTTP |     MCP SSE    | Note |
|---------|--------|--------|--------|
| n8n | n/a | n/a |My account is expired |
| dify | ✅ Pass | ✅ Pass | |
| activepieces | ✅ Pass | ✅ Pass | |
| make | ✅ Pass | ✅ Pass | |
| langsmith |  ✅ Pass | ❌ Fail |  |
Screenshot 2026-03-27 at 11 03 39

### Unit test
image

packages/server/src/utils/mockRequest.test.ts
```diff
@@ -0,0 +1,77 @@
+import { createMockRequest } from '../../src/utils/mockRequest'
```
jchui-wd: can just be  './mockRequest'

jchui-wd:
Thanks for the changes @prd-hoang-doan, just a few more things.

1. Small change above on `mockRequest.test.ts`
2. Can there be a indicator when `getMcpServerConfigApi.error` fails, something like
    ```
    useEffect(() => {
        if (getMcpServerConfigApi.error) {
            showError(`Failed to load MCP Server config: ...`)
        }
    }, [getMcpServerConfigApi.error])
    ```
    McpServer.jsx.
3. nit: Just for consistency, can you rename `mcpServer.js` to be `mcpserver.js` in `/packages/ui/src/api/mcpServer.js`

prd-hoang-doan:
@jchui-wd Thank you for pointing out the inconsistent implementation. I updated points 2 and 3. 
What do you want for point 1?

packages/server/src/utils/mockRequest.test.ts
jchui-wd: @prd-hoang-doan apologies, forgot to submit it, but it's just a small import change here

prd-hoang-doan: Good catch @jchui-wd, I updated it in the latest commit. Thank you for pointing out.

HenryHengZJ:
thanks @prd-hoang-doan so much for this PR!!

but I tried saving MCP config it doesnt allow me to:

https://github.com/user-attachments/assets/bf1cf181-6a4c-4840-8fc7-3c24944333a4

Update: fixed in my latest [commit](https://github.com/FlowiseAI/Flowise/pull/5930/commits/14ae53f618c26d3b604f65331786b42262642bc9)

Also, can you rename the mcp related routes to get rid of the `chatflow` naming, because we also want to enable this for agentflow, assistants, agents as well.

packages/server/src/routes/mcp-endpoint/index.ts
```diff
@@ -0,0 +1,65 @@
+import express from 'express'
+import cors from 'cors'
+import mcpEndpointController from '../../controllers/mcp-endpoint'
+
+const router = express.Router()
+
+// Body size limit: 1MB max for MCP JSON-RPC payloads (overrides the global 50mb limit)
+router.use(express.json({ limit: '1mb', type: 'application/json' }))
```
HenryHengZJ, line 8: will this only override this route? or does it override of all of the routes? can we verify that

packages/server/src/services/mcp-endpoint/index.ts
```diff
@@ -0,0 +1,359 @@
...
+}
+
+/**
+ * Build the zod input schema parameters for the tool.
+ * For chatflows: always has a mandatory `question` string.
+ * For agentflows: has an optional `question` and optional `form` object.
+ */
+function buildInputSchema(chatflow: ChatFlow) {
+    if (chatflow.type === 'MULTIAGENT' || chatflow.type === 'AGENTFLOW') {
+        return {
```
HenryHengZJ: We should ge tthe `chatflow.flowData`, parse it, and check the Start node input.

If its a Form Type, populate the respective zod schema for the fields.

We only allow either `question` or `form`, not both.

https://github.com/FlowiseAI/Flowise/blob/main/packages/server/src/utils/buildAgentflow.ts#L1166

packages/server/src/controllers/mcp-server/index.ts
```diff
@@ -0,0 +1,104 @@
...
+            throw new InternalFlowiseError(StatusCodes.NOT_FOUND, 'Error: mcpServerController.getMcpServerConfig - workspace not found!')
+        }
+        const apiResponse = await mcpServerService.getMcpServerConfig(req.params.id, workspaceId)
+        return res.json(apiResponse)
+    } catch (error) {
+        next(error)
+    }
+}
+
+const createMcpServerConfig = async (req: Request, res: Response, next: NextFunction) => {
```
christopherholland-workday, line 25: Should we protect this using https://github.com/FlowiseAI/Flowise/blob/fedc6ebc82f28e0ab0fdc9d434986de0a08964a1/packages/components/nodes/tools/MCP/core.ts#L382?

packages/server/src/controllers/mcp-server/index.ts
```diff
@@ -0,0 +1,104 @@
...
+            throw new InternalFlowiseError(StatusCodes.NOT_FOUND, 'Error: mcpServerController.createMcpServerConfig - workspace not found!')
+        }
+        const apiResponse = await mcpServerService.createMcpServerConfig(req.params.id, workspaceId, req.body || {})
+        return res.status(StatusCodes.CREATED).json(apiResponse)
+    } catch (error) {
+        next(error)
+    }
+}
+
+const updateMcpServerConfig = async (req: Request, res: Response, next: NextFunction) => {
```
christopherholland-workday, line 44: Same as above

packages/server/src/services/mcp-server/index.ts
```diff
@@ -0,0 +1,267 @@
...
+        const appServer = getRunningExpressApp()
+        const chatflow = await appServer.AppDataSource.getRepository(ChatFlow).findOne({
+            where: { id: chatflowId, workspaceId }
+        })
+        if (!chatflow) {
+            throw new InternalFlowiseError(StatusCodes.NOT_FOUND, `Chatflow ${chatflowId} not found`)
+        }
+
+        // If already has an MCP config, return it
+        const existing = parseMcpConfig(chatflow)
```
christopherholland-workday, line 95: Same as above regarding https://github.com/FlowiseAI/Flowise/blob/fedc6ebc82f28e0ab0fdc9d434986de0a08964a1/packages/components/nodes/tools/MCP/core.ts#L382

packages/server/src/services/mcp-endpoint/index.ts
```diff
@@ -0,0 +1,359 @@
...
+import logger from '../../utils/logger'
+import { ChatType, IMcpServerConfig } from '../../Interface'
+
+// Active SSE transport sessions: sessionId → { transport, mcpServer, chatflowId }
+interface SseSession {
+    transport: SSEServerTransport
+    mcpServer: McpServer
+    chatflowId: string
+}
+const sseSessions = new Map<string, SseSession>()
```
HenryHengZJ: Although this works fine if Flowise is running locally or low usage, but this will bring scalability issues in cloud production.

The per-chatflow cap of 20 limits sessions per chatflow, but across millions of chatflows the Map can grow without bound. There's no global cap, no TTL-based eviction, and no idle timeout. Very qucikly we will have memory leak.

Also, this Map lives in a single Node.js process. In a multi-instance deployment (Kubernetes pods, horizontal auto-scaling), SSE sessions are pinned to the process that created them. If a client's message POST is load-balanced to a different instance, it gets a 404. You'd need sticky sessions or an external session store (Redis) to make SSE work across instances.

This is a common scalability issue.

Now, there are ways to solve this issue. It typically require building a Redis pub/sub bridge (publish message payloads on one instance, subscribe and feed into the local SSEServerTransport on another). We are using that for aborting message streaming. You can refer the implementation [here](https://github.com/FlowiseAI/Flowise/blob/main/packages/server/src/commands/worker.ts#L44).

However this has significant complexity for a deprecated protocol. So my suggestion is to remove support for SSE transport, the related services, endpoints, etc. We can just support Streamable HTTP transport (protocol version 2025-03-26) as it is stateless — each POST creates a fresh server+transport, processes the JSON-RPC message, and tears down. It works across multiple instances with zero extra infrastructure.

packages/server/src/services/mcp-endpoint/index.ts
```diff
@@ -0,0 +1,359 @@
...
+
+    // Connect server to transport
+    await mcpServer.connect(transport)
+
+    // Clean up when the HTTP response finishes.
+    // NOTE: We must NOT close the server immediately after handleRequest() because
+    // the transport's handlePostRequest() fires onmessage() without awaiting it.
+    // If we close() too early, the SSE response stream is terminated before the
+    // McpServer has finished processing the request and writing its JSON-RPC response.
+    res.on('close', () => {
```
HenryHengZJ, line 307: Here we need to wire up abort on client disconnect

**Files:**
- `packages/server/src/services/chat-messages/index.ts` — reuse `abortChatMessage` (lines 190-209)

**Problem:** When an MCP client disconnects mid-request, `res.on('close')` closes the MCP server but the underlying `utilBuildChatflow` call (LLM inference, vector lookups) continues to completion, wasting tokens and compute.

**Pattern to follow:** `abortChatMessage` in `services/chat-messages/index.ts` — already handles both modes:
```typescript
// Main mode:  appServer.abortControllerPool.abort(id)
// Queue mode: queueManager.getPredictionQueueEventsProducer().publishEvent({ eventName: 'abort', id })
```

The abort ID format is `${chatflowId}_${chatId}`, which matches what `utilBuildChatflow` uses internally (line 1009: `const abortControllerId = \`${chatflow.id}_${chatId}\``).

**Approach:**

`chatflowCallback` needs to know the `chatId` that `utilBuildChatflow` will use, so it can pass it back to the disconnect handler. The `chatId` comes from `incomingInput.chatId` or is auto-generated as a UUID inside `utilBuildChatflow`.

1. In `chatflowCallback`, generate a `chatId` (UUID) and include it in the mock request body:
   ```typescript
   const chatId = uuidv4()
   const mockReq = createMockRequest({
       chatflowId: chatflow.id,
       body: { question: args.question || '', chatId, ... }
   })
   ```
   This ensures `utilBuildChatflow` uses our known `chatId` (line 1007: `const chatId = incomingInput.chatId ?? ... ?? uuidv4()`)

2. Return the `chatId` alongside the result so the handler can use it for abort. Restructure `chatflowCallback` to also return `chatId`:
   ```typescript
   return { chatId, content: [{ type: 'text', text: textContent }] }
   ```

3. In `handleMcpRequest`, update the `res.on('close')` handler to call abort:
   ```typescript
   import chatMessagesService from '../chat-messages/index'

   res.on('close', () => {
       mcpServer.close().catch(() => {})
       // chatId is captured from the tool callback scope
       if (chatId) {
           chatMessagesService.abortChatMessage(chatId, chatflowId).catch(() => {})
       }
   })
   ```

**Key detail:** The `res.on('close')` handler is registered *before* the tool callback fires (it's set up during `handleMcpRequest`). The `chatId` needs to be accessible in that closure. Options:
- Use a mutable variable in `handleMcpRequest` scope that the tool callback populates
- Or generate the `chatId` upfront in `handleMcpRequest` and pass it through to the callback via closure

The second option is cleaner — generate `chatId` once, use it in both the mock request and the abort handler.

HenryHengZJ:
Note: we might also want to allow users store MCP server similar to Custom tools, but thats a separate story and PR https://github.com/orgs/FlowiseAI/projects/2?pane=issue&itemId=168148790

prd-hoang-doan:
Thank you for your valuable feedback @HenryHengZJ @christopherholland-workday 
I will follow up and fix these issues above.
Besides, I found your new ticket about Custom MCP Server. If there is no one implementing it, I am happy to contribute that. This is in development right now. You can take a look at demo this and give me any feedback.

https://github.com/user-attachments/assets/5a688168-a45e-4fe2-b85c-ce2db5220d54

You can watch the full flow at Youtube: https://youtu.be/_sJ1MIRil4o

prd-hoang-doan:
Hi @HenryHengZJ , @christopherholland-workday 
1. Also, can you rename the mcp related routes to get rid of the `chatflow` naming, because we also want to enable this for agentflow, assistants, agents as well. → remove `chatflow` in error message ✅
2. We should ge tthe `chatflow.flowData`, parse it, and check the Start node input. → updated ✅
3. Should we protect this using `validateMCPServerConfig`? → No, I don't think that we should a function since the client don't send any config related URL, command. I will use it for the feature Custom MCP Server that @HenryHengZJ mentioned above.
4.` router.use(express.json({ limit: '1mb', type: 'application/json' }))`  → I added unit tests for the route  ✅
5. `const sseSessions = new Map<string, SseSession>()` → remove deprecated SSE protocol ✅
6. `res.on('close')` handler is registered before the tool callback fires → updated ✅, test the request capture the right chatID to abort in the clip below:

https://github.com/user-attachments/assets/52b0ef4f-7110-409a-9068-349109ac35e2

Please take a look at the latest commit and feel free to provide feedback; I will follow up.

By the way, I have to install the package to make sure the build is successful in this branch.
image

packages/server/src/services/mcp-server/index.ts
```diff
@@ -0,0 +1,267 @@
...
+        throw new InternalFlowiseError(StatusCodes.BAD_REQUEST, result.error.errors[0].message)
+    }
+    return result.data
+}
+
+/**
+ * Generate a random 32-char hex token (128 bits of entropy)
+ */
+function generateToken(): string {
+    return crypto.randomBytes(16).toString('hex')
```
igor-magun-wd, line 39: This is generating a 16-byte/16-char token, not 32. 32 is best practice

prd-hoang-doan: I double checked it. This is 32-char token. 
You can play ground at https://stackblitz.com/edit/node-m5rymebe?file=index.js
image

igor-magun-wd: Ah sorry about that, forgot that it ends up as 32 with hex. I think this is likely acceptable then.

packages/server/src/routes/index.ts
```diff
@@ -124,6 +126,8 @@ router.use('/executions', executionsRouter)
 router.use('/validation', validationRouter)
 router.use('/agentflowv2-generator', agentflowv2GeneratorRouter)
 router.use('/text-to-speech', textToSpeechRouter)
+router.use('/chatflows-mcp-server', mcpServerRouter)
```
HenryHengZJ: hey @prd-hoang-doan sorry I should have made it clearer. We can get rid of any `chatflows` naming from the mcp routes:

`/chatflows-mcp-server` -> `/mcp-server`

`/mcp/chatflow/:chatflowId` -> `/mcp/:chatflowId`

packages/server/src/services/mcp-endpoint/index.ts
```diff
@@ -0,0 +1,247 @@
...
+/**
+ * Build the zod input schema parameters for the tool.
+ * For chatflows: always has a mandatory `question` string.
+ * For agentflows: only allow one of `question` or `form` (object)
+ */
+function buildInputSchema(chatflow: ChatFlow) {
+    const inputType = getToolInputType(chatflow)
+    if (inputType === 'form') {
+        return {
+            form: z.record(z.any()).describe('Form inputs for the agent flow')
```
HenryHengZJ: The next step should be getting the `formInputTypes` and convert to respective zod schema.

For example, this is the input for `startAgentflow` node:
```
"inputs": {
          "startInputType": "formInput",
          "formTitle": "test",
          "formDescription": "test",
          "formInputTypes": [
            {
              "type": "string",
              "label": "Title",
              "name": "title",
              "addOptions": ""
            },
            {
              "type": "string",
              "label": "Desc",
              "name": "desc",
              "addOptions": ""
            }
          ],
          "startEphemeralMemory": "",
          "startState": [
            {
              "key": "foo",
              "value": "bar"
            }
          ],
          "startPersistState": ""
        },
```

The zod schema should looks like:

```
form: z.object({
  title: z.string(),
  desc: z.string(),
}).describe('Form inputs for the agent flow')
```

The `z.record(z.any())` was a loose catch-all. We should be mapping each name field to its type

HenryHengZJ:
how did you simulate when an MCP client disconnects mid-request, and triggers `res.on('close')` and verify indeed we have aborted the underlying ongoing process? the video is too blurry to see

prd-hoang-doan:
@HenryHengZJ I confirmed that we captured the correct `chatId` and aborted it to avoid the process/worker running forever.
Simulate Abort: https://youtu.be/BEwDAru1H-w
In my example, the flow prediction (2 LLM node) is still complete even though I stop it in the mid request. It can abort immediately, like a stop button in a chat pop-up, since the client sends the request to the server. In contrast, the third party does not seem to send anything to abort, the server waits, and closes later.

packages/server/src/services/mcp-endpoint/index.ts
prd-hoang-doan: I updated it. Thank you for pointing out the form data.

packages/server/src/routes/index.ts
prd-hoang-doan: I updated the endpoint as you mentioned above.

HenryHengZJ:
yeah thats what I tested as well, I tried cancelling the request from third party client like Cursor, it doesnt trigger  `res.on('close')`, so the underlying process still carry on and eventually finish.

In that case, we can then remove the abort inside the `res.on('close')` since its redundant. We already removed the abort controller once the process finishes [here](https://github.com/FlowiseAI/Flowise/blob/main/packages/server/src/utils/buildChatflow.ts#L1103).

prd-hoang-doan:
@HenryHengZJ I removed abortChatMessage in `res.on('close')`. If there is any concern, feel free to let me know.

packages/server/src/services/mcp-endpoint/index.ts
HenryHengZJ: @prd-hoang-doan you have to parse `formInputTypes` and map to zod schema. `formInputTypes` can be anything that user specified.

`title` and `desc` above is just an example

HenryHengZJ:
A reminder to check UI for dark mode as well. I've updated it [here](https://github.com/FlowiseAI/Flowise/pull/5930/commits/737e232f90938433d46296fd2b237434665c7fb6)

packages/server/src/services/mcp-endpoint/index.ts
```diff
@@ -0,0 +1,245 @@
...
+/**
+ * Build the zod input schema parameters for the tool.
+ * For chatflows: always has a mandatory `question` string.
+ * For agentflows: only allow one of `question` or `form` (object)
+ */
+function buildInputSchema(chatflow: ChatFlow) {
+    const inputType = getToolInputType(chatflow)
+    if (inputType === 'form') {
+        return {
+            form: z
```
HenryHengZJ: you have to parse `formInputTypes` and map to zod schema. `formInputTypes` can be anything that user specified.

`title` and `desc` above is just an example

prd-hoang-doan: @HenryHengZJ You're right — I oversimplified that part.
I'll update the implementation to properly parse `formInputTypes` and dynamically map them to a Zod schema, as well as unit testing.

Verify the result with 4 data type on third-party app:
https://github.com/user-attachments/assets/dd0b6bde-9969-4933-8fbb-584bda4e454e

HenryHengZJ (approved):
looks good now, thank you for the effort @prd-hoang-doan !!

HenryHengZJ:
deployed and tested in dev, works great!

prd-hoang-doan:
Excited to see this merged. Looking forward to seeing this live in the next release.

PR #5937, opened 2026-03-09
feat(agentflow) Create Async Data Fetching Infrastructure in agentflow

j-sanaa:
JIRA - FLOWISE-227
- Created API modules for models, tools, and credentials 
- created loadMethodRegistry which allows for dynamic API dispatch
- Added test cases for all new files
- Added new example -  API inspector to verify that all endpoints are called correctly using the newly added API models 

Recording of testing all new API endpoints - 

https://github.com/user-attachments/assets/07bb6939-6651-46c5-bd03-2f54f6c70835

### DIFF
--- a/packages/agentflow/src/index.ts
+++ b/packages/agentflow/src/index.ts
@@ -58,7 +58,6 @@ export type {
 } from './core/types'
 /* eslint-enable simple-import-sort/exports */
 
-
 // Utilities (for advanced usage)
 export { filterNodesByComponents, isAgentflowNode } from './core/node-catalog'
 export { AGENTFLOW_ICONS, DEFAULT_AGENTFLOW_NODES, getAgentflowIcon, getNodeColor } from './core/node-config'


packages/agentflow/src/core/types/api.ts
```diff
@@ -23,3 +25,27 @@ export interface ApiResponse<T> {
     data: T
     status: number
 }
+
+export type Model = Omit<NodeData, 'id'>
```
jocelynlin-wd: can you confirm if api returned type matches `NodeData`?

packages/agentflow/examples/src/demos/ApiInspectorExample.tsx
```diff
@@ -0,0 +1,396 @@
+/**
```
jocelynlin-wd: It's helpful to show how api works in your PR recording but I don't think this is a valid demo for users to learn about how to use Agentflow sdk.

packages/agentflow/src/infrastructure/api/tools.ts
```diff
@@ -0,0 +1,28 @@
+import type { AxiosInstance } from 'axios'
+
+import type { Tool } from '@/core/types'
+
+/**
+ * Create tools API functions bound to a client instance
```
jocelynlin-wd, line 6: I don't think agentflow v2 supports creating tools within agentflow canvas directly?  Do we want a TODO to review if this api should stay? 

This other api `/api/v1/node-load-method/toolAgentflow` is used instead to list the tools available for agentflow. Perhaps we should swap to use this api instead?

packages/agentflow/src/index.ts
```diff
@@ -37,12 +42,14 @@ export type {
     // Node data
     InputAnchor,
     InputParam,
+    Model,
```
jocelynlin-wd: can you move this to `api` instead? You added the type to `api.ts` and this type seems to be used for API operations?

packages/agentflow/src/index.ts
```diff
@@ -37,12 +42,14 @@ export type {
...
     NodeData,
     NodeInput,
     NodeOutput,
     OutputAnchor,
     PaletteRenderProps,
     RequestInterceptor,
+    Tool,
```
jocelynlin-wd: same as Model, you added this type to `api.ts` and this type seems to be used for API operations, should be in API section not here?

packages/agentflow/src/infrastructure/api/models.ts
```diff
@@ -0,0 +1,28 @@
+import type { AxiosInstance } from 'axios'
+
+import type { Model } from '@/core/types'
+
+/**
+ * Create models API functions bound to a client instance
+ */
+export function createModelsApi(client: AxiosInstance) {
```
jocelynlin-wd: maybe we can match the name to chat models as well?  Models sound very generic

packages/agentflow/src/infrastructure/api/loadMethodRegistry.ts
```diff
@@ -0,0 +1,25 @@
...
+import type { ModelsApi } from './models'
+import type { ToolsApi } from './tools'
+
+export interface ApiServices {
+    modelsApi: ModelsApi
+    toolsApi: ToolsApi
+    credentialsApi: CredentialsApi
+}
+
+export const loadMethodRegistry: Record<string, (_apis: ApiServices, _params?: Record<string, unknown>) => Promise<unknown>> = {
```
jocelynlin-wd, line 26: can you add jsdoc on what and how to use this function?

packages/agentflow/src/infrastructure/api/tools.ts
j-sanaa: Updated getAllTools to use /api/v1/node-load-method/toolAgentflow instead

packages/agentflow/examples/src/demos/ApiInspectorExample.tsx
j-sanaa: Removed the API inspector example to avoid confusion

packages/agentflow/src/core/types/api.ts
j-sanaa: Confirmed that the return type of node data is the same as tools

packages/agentflow/src/infrastructure/store/ApiContext.tsx
```diff
@@ -4,13 +4,28 @@ import type { AxiosInstance } from 'axios'
...
 
-import { type ChatflowsApi, createApiClient, createChatflowsApi, createNodesApi, type NodesApi } from '../api'
+import {
+    bindChatModelsApi,
+    bindCredentialsApi,
+    bindToolsApi,
+    type ChatflowsApi,
+    type ChatModelsApi,
+    createApiClient,
+    createChatflowsApi,
```
jocelynlin-wd, line 14: would be nice to also update `createChatflowsApi` and `createNodesApi` to match the function names (eg `bindChatflowsApi`)

PR #5956, opened 2026-03-11
Feat/agentflow async input handlers

j-sanaa:
JIRA - FLOWISE-228
- Added asyncOptions and asyncMultiOptions input types, powered by a new useAsyncOptions hook that fetches options from the API with loading, error, and retry states
- Extracted async rendering into AsyncInput in features/node-editor/ and injected it via an AsyncInputComponent prop to keep atoms/ free of infrastructure imports
- Fixed API endpoints for models and tools (both now use POST /node-load-method/agentAgentflow), and fixed a URL-encoding bug in getCredentialsByName by passing an array instead of a manually joined string
- Added tests for useAsyncOptions, AsyncInput integration, and async-driven field visibility in flow validation

https://github.com/user-attachments/assets/2fcd984f-fd58-47f6-86c1-94ede6d59f1b


.claude/settings.json
jocelynlin-wd, line 1: would be nice to revert anything on this file

packages/agentflow/src/infrastructure/api/credentials.ts
```diff
@@ -16,9 +16,9 @@ export function bindCredentialsApi(client: AxiosInstance) {
         },
 
         /**
-         * Get credentials filtered by component credential name
+         * Get credentials filtered by one or more component credential names.
          */
-        getCredentialsByName: async (credentialName: string): Promise<Credential[]> => {
+        getCredentialsByName: async (credentialName: string | string[]): Promise<Credential[]> => {
```
jocelynlin-wd, line 21: does the API accept both single credential and credential array?

packages/agentflow/src/atoms/NodeInputHandler.tsx
```diff
@@ -16,6 +16,16 @@ const CustomWidthTooltip = styled(({ className, ...props }: TooltipProps) => <To
     }
 })
 
+/** Props passed to an async input component (asyncOptions / asyncMultiOptions). */
+export interface AsyncInputProps {
+    inputParam: InputParam
+    value: unknown
```
jocelynlin-wd, line 22: is it possible to define what the value type should be?

packages/agentflow/src/features/node-editor/AsyncInput.tsx
```diff
@@ -0,0 +1,190 @@
...
+    const { options, loading, error, refetch } = useAsyncOptions({
+        loadMethod: inputParam.loadMethod,
+        credentialNames: inputParam.credentialNames,
+        params
+    })
+
+    if (error) {
+        return (
+            <Box sx={{ mt: 1, display: 'flex', alignItems: 'center', gap: 1 }}>
+                <Typography variant='caption' color='error' sx={{ flexGrow: 1 }}>
```
jocelynlin-wd, line 23: this error display seems to be the same as the one used in AsyncMultiOptionsInput (line 106-108). Shall we extract to a small shared component to reduce duplication?

packages/agentflow/src/features/node-editor/AsyncInput.tsx
```diff
@@ -0,0 +1,190 @@
...
+import Autocomplete from '@mui/material/Autocomplete'
+import { IconRefresh } from '@tabler/icons-react'
+
+import type { AsyncInputProps } from '@/atoms'
+import type { NodeOption } from '@/core/types'
+import { useAsyncOptions } from '@/infrastructure/api/hooks'
+
+function AsyncOptionsInput({ inputParam, value, disabled, onChange, nodeName, inputValues }: AsyncInputProps) {
+    const params =
+        nodeName || inputValues ? { ...(nodeName ? { nodeName } : {}), ...(inputValues ? { inputs: inputValues } : {}) } : undefined
```
jocelynlin-wd, line 13: Duplicated params construction — Both AsyncOptionsInput (line 12-13) and AsyncMultiOptionsInput (line 95-96) build params identically. Extract this to AsyncInput and pass it down, or compute it once in the dispatcher?

packages/agentflow/src/infrastructure/api/tools.ts
```diff
@@ -10,8 +10,20 @@ export function bindToolsApi(client: AxiosInstance) {
...
+        getAllTools: async (nodeName = 'toolAgentflow'): Promise<Tool[]> => {
+            const response = await client.post(`/node-load-method/${encodeURIComponent(nodeName)}`, { loadMethod: 'listTools' })
+            return response.data
+        },
+
+        /**
+         * Get input argument names for the currently selected tool.
+         * Passes current node inputs as `currentNode.inputs` so the server can resolve the selected tool.
+         */
+        getToolInputArgs: async (inputs: Record<string, unknown>, nodeName = 'toolAgentflow'): Promise<Tool[]> => {
```
jocelynlin-wd, line 22: nit: Tool seems to be an alias for `NodeOption`, which the shape is correct — `getToolInputArgs` returns objects with label, name, etc., which fits NodeOption/Tool exactly.   The naming is slightly misleading since "Tool" suggests a tool, not a tool's input arguments

packages/agentflow/src/features/node-editor/AsyncInput.tsx
```diff
@@ -0,0 +1,190 @@
...
+            filterSelectedOptions
+            size='small'
+            disabled={disabled}
+            options={options}
+            value={selectedOptions}
+            getOptionLabel={(o) => o.label}
+            isOptionEqualToValue={(o, v) => o.name === v.name}
+            onChange={(_e, selection) => {
+                const names = selection.map((s) => s.name)
+                onChange(names.length > 0 ? JSON.stringify(names) : '')
```
jocelynlin-wd, line 145: JSON string storage for multi-select values — `AsyncMultiOptionsInput` stores values as `JSON.stringify(names)`. The validation layer and the component itself both need to parse this back. This works but creates a coupling that could be error-prone for future consumers. Worth a comment in the type definition for InputParam documenting this convention?

packages/agentflow/src/infrastructure/api/loadMethodRegistry.ts
```diff
@@ -17,15 +17,17 @@ export interface ApiServices {
  * and must return a `Promise` of the option values to populate the node's dropdown.
  *
  * ### Built-in entries
- * - `listChatModels` — fetches available chat models via `GET /assistants/components/chatmodels`
+ * - `listModels` — fetches available chat models via `GET /assistants/components/chatmodels`
```
jocelynlin-wd, line 20: nit, looks like the api has been changed to use POST instead?
```
getChatModels: async (): Promise<ChatModel[]> => {
  const response = await client.post('/node-load-method/agentAgentflow', { loadMethod: 'listModels' })
  ...
}
```

jocelynlin-wd (approved):
no-blockers, feel free to review the comments in a follow up PR instead

PR #5959, opened 2026-03-12
feat: add MiniMax chat model integration with M2.7 as default

octo-patch:
## Summary
Add MiniMax as a chat model provider in Flowise, with the latest M2.7 model as default.

## Changes
- Add `ChatMiniMax` node wrapping `ChatOpenAI` with MiniMax API base URL
- Add `MiniMaxApi` credential type for API key management
- Register MiniMax models in `models.json`: M2.7, M2.7-highspeed, M2.5, M2.5-highspeed
- Set `MiniMax-M2.7` as the default model
- Include temperature clamping for MiniMax API constraints

## Why
MiniMax-M2.7 is the latest flagship model with enhanced reasoning and coding capabilities. This integration allows Flowise users to leverage MiniMax models in their AI workflows.

## Testing
- Verified models.json is valid JSON with correct model ordering
- M2.7 is first in list and set as default
- All previous models retained as alternatives

### DIFF
--- a/packages/components/credentials/MiniMaxApi.credential.ts
+++ b/packages/components/credentials/MiniMaxApi.credential.ts
@@ -1,0 +1,23 @@
+import { INodeCredential, INodeParams } from '../src/Interface'
+
+class MiniMaxApi implements INodeCredential {
+    label: string
+    name: string
+    version: number
+    inputs: INodeParams[]
+
+    constructor() {
+        this.label = 'MiniMax API'
+        this.name = 'miniMaxApi'
+        this.version = 1.0
+        this.inputs = [
+            {
+                label: 'MiniMax API Key',
+                name: 'miniMaxApiKey',
+                type: 'password'
+            }
+        ]
+    }
+}
+
+module.exports = { credClass: MiniMaxApi }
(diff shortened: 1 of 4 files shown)


octo-patch (commented):
Pushed fixes for two high-priority issues in `ChatMiniMax.ts` (commit 68c36d7):

### 1. Temperature validation (line 155)

`parseFloat(temperature)` could produce `NaN` (empty/undefined input) or `0` (user enters zero), both of which MiniMax's API rejects (requires temperature in `(0.0, 1.0]`).

**Fix:** Validate the parsed value and clamp to the accepted range:
- `NaN` or `<= 0` → default to `0.7`
- `> 1.0` → clamp to `1.0`

### 2. baseURL removal (line 180)

`parsedBaseOptions.baseURL = undefined` does not remove the property — the key remains with an `undefined` value. When spread into `configuration: { baseURL: this.baseURL, ...parsedBaseOptions }`, the `undefined` value overrides the hardcoded MiniMax API URL, breaking the integration.

**Fix:** Use `delete parsedBaseOptions.baseURL` to fully remove the property, ensuring the spread does not clobber the intended base URL.

Issue #5958, opened 2026-03-12
Title: [BUG] "Folder with Files" node is unusable in Docker due to aggressive Path Traversal validation


thediscordiansociety-dev:
### Describe the bug

The "Folder with Files" document loader is currently broken for local Docker installations. Any attempt to provide a valid path (absolute or relative) results in an Error: Invalid folder path: Path traversal detected.

This appears to be caused by a recent security update (likely around v3.0.6) that blocks any path containing slashes or dots. While intended to prevent external attacks, it has rendered local file indexing impossible for users mounting local drives (like Obsidian vaults) into Docker containers.
Setup

### To Reproduce

Run Flowise via Docker Compose.

Mount a local Windows directory to a container volume (e.g., - "D:/MyNotes:/vault").

Add the Folder with Files node to a chatflow.

Enter /vault, ./vault, or vault into the Folder Path field.

Attempt to "Upsert" or save.

Result: The UI returns "Invalid folder path: Path traversal detected."

### Expected behavior

Flowise should allow paths that resolve within the container's mounted volumes, especially when running in a local/standalone environment.

### Use Method

Docker

### Flowise Version

flowise@3.0.13

### Operating System

Windows

### Browser

Other

### Additional context

Users have spent hours attempting to bypass this by mapping volumes directly into /root/.flowise/vault, but the validator blocks any string containing a /. It also did not work with doing . or vault or any other method.  This essentially deprecates the local folder loading feature for the entire Docker user base.

HenryHengZJ:
the fix will be released in the next Docker version

PR #5962, opened 2026-03-12
Traversal env var

christopherholland-workday:
Related to https://github.com/FlowiseAI/Flowise/issues/5861, https://github.com/FlowiseAI/Flowise/issues/5958, and https://github.com/FlowiseAI/Flowise/discussions/5326.

Flowise-334.

### DIFF
--- a/packages/components/src/validator.ts
+++ b/packages/components/src/validator.ts
@@ -62,6 +62,10 @@ export const isPathTraversal = (path: string): boolean => {
  * @returns {boolean} True if path traversal detected, false otherwise
  */
 export const isUnsafeFilePath = (filePath: string): boolean => {
+    if (process.env.PATH_TRAVERSAL_SAFETY === 'false') {
+        return false
+    }
+
     if (!filePath || typeof filePath !== 'string') {
         return true
     }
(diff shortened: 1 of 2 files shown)


packages/components/src/validator.ts
```diff
@@ -32,17 +32,28 @@ export const isValidURL = (url: string): boolean => {
  * @returns {boolean} True if path traversal detected, false otherwise
  */
 export const isPathTraversal = (path: string): boolean => {
-    // Check for common path traversal patterns
+    // PATH_TRAVERSAL_SAFETY defaults to true; must be explicitly set to 'false' to disable
```
HenryHengZJ, line 35: apart from this function, there are many other similar functions such as `isUnsafeFilePath`, `validateVectorStorePath`

we should have an env variable that turns off the check for all of these. 1 env variable for all of them. otherwise there are too many env variables that users have to set to get rid of all these checks when they are running locally or in other environment

christopherholland-workday: Will add for these two functions, as I think it is relevant to group methods related to file safety. There are technically some others too, but I don't think those ones are causing customers pain as the enforcement are a bit less strict

PR #5967, opened 2026-03-14
Fix/agentflow model parameters

j-sanaa:
JIRA - FLOWISE-337
- Add loadConfig flag to InputParam and ConfigInput component that renders a collapsible accordion of sub-node parameters when a model is selected from an async dropdown
- Add listRegions loadMethod to the registry via a generic nodesApi.loadNodeMethod API, enabling node-specific server-side methods like region listing
- Add credential field support by extracting the top-level credential property in initNode and rendering it via a new case 'credential' in NodeInputHandler, storing the selected credential ID as FLOWISE_CREDENTIAL_ID (similar to v2)

https://github.com/user-attachments/assets/29671ff6-723e-4970-be11-2e5be3904021

### DIFF
--- a/packages/agentflow/src/features/node-editor/AsyncInput.tsx
+++ b/packages/agentflow/src/features/node-editor/AsyncInput.tsx
@@ -8,15 +8,24 @@ import type { AsyncInputProps } from '@/atoms'
 import type { NodeOption } from '@/core/types'
 import { useAsyncOptions } from '@/infrastructure/api/hooks'
 
+/**
+ * Build the params object for useAsyncOptions.
+ * Only includes inputValues when the loadMethod actually needs them (e.g. listToolInputArgs).
+ * Including them unconditionally causes every keystroke in sibling text fields to change the
+ * serialised paramsKey, triggering unnecessary refetches for all async dropdowns.
+ */
+function buildAsyncParams(
+    loadMethod: string | undefined,
+    nodeName: string | undefined,
+    inputValues: Record<string, unknown> | undefined
+): Record<string, unknown> | undefined {
+    const needsInputs = loadMethod === 'listToolInputArgs'
+    if (!nodeName && !(needsInputs && inputValues)) return undefined
+    return { ...(nodeName ? { nodeName } : {}), ...(needsInputs && inputValues ? { inputs: inputValues } : {}) }
+}
+
 function AsyncOptionsInput({ inputParam, value, disabled, onChange, nodeName, inputValues }: AsyncInputProps) {
-    // Only include inputValues in params for loadMethods that actually need them (e.g. listToolInputArgs).
-    // Including them unconditionally causes every keystroke in sibling text fields to change the
-    // serialised paramsKey, triggering unnecessary refetches for all async dropdowns.
-    const needsInputs = inputParam.loadMethod === 'listToolInputArgs'
-    const params =
-        nodeName || (needsInputs && inputValues)
-            ? { ...(nodeName ? { nodeName } : {}), ...(needsInputs && inputValues ? { inputs: inputValues } : {}) }
-            : undefined
+    const params = buildAsyncParams(inputParam.loadMethod, nodeName, inputValues)
     const { options, loading, error, refetch } = useAsyncOptions({
         loadMethod: inputParam.loadMethod,
         credentialNames: inputParam.credentialNames,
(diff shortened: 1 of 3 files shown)


packages/agentflow/src/features/node-editor/AsyncInput.tsx
```diff
@@ -9,8 +9,14 @@ import type { NodeOption } from '@/core/types'
 import { useAsyncOptions } from '@/infrastructure/api/hooks'
 
 function AsyncOptionsInput({ inputParam, value, disabled, onChange, nodeName, inputValues }: AsyncInputProps) {
+    // Only include inputValues in params for loadMethods that actually need them (e.g. listToolInputArgs).
+    // Including them unconditionally causes every keystroke in sibling text fields to change the
+    // serialised paramsKey, triggering unnecessary refetches for all async dropdowns.
+    const needsInputs = inputParam.loadMethod === 'listToolInputArgs'
```
jocelynlin-wd: maybe abstract line 15-19 (and line 103-105) into a sharable function like below since the same input transformation is defined twice
```
  function buildAsyncParams(inputParam: InputParam, nodeName?: string, inputValues?: Record<string, unknown>) {
      const needsInputs = inputParam.loadMethod === 'listToolInputArgs'
      if (!nodeName && !(needsInputs && inputValues)) return undefined
      return {
          ...(nodeName ? { nodeName } : {}),
          ...(needsInputs && inputValues ? { inputs: inputValues } : {})
      }
  }
```

packages/agentflow/src/features/node-editor/ConfigInput.test.tsx
```diff
@@ -0,0 +1,303 @@
...
+
+import { ConfigInput } from './ConfigInput'
+
+// ─── Mocks ────────────────────────────────────────────────────────────────────
+
+jest.mock('reactflow', () => ({
+    Handle: () => null,
+    Position: { Left: 'left' },
+    useUpdateNodeInternals: () => jest.fn()
+}))
```
jocelynlin-wd: we have a package level mock at `src/__mocks__/reactflow.tsx` that already covers `Handle`, `Position`, and `useUpdateNodeInternals`. The mock here is redundant and should be removed.

packages/agentflow/src/features/node-editor/ConfigInput.test.tsx
```diff
@@ -0,0 +1,303 @@
...
+}
+
+beforeEach(() => {
+    jest.clearAllMocks()
+    mockGetNodeByName.mockResolvedValue(fakeNodeDefinition)
+})
+
+// ─── Tests ────────────────────────────────────────────────────────────────────
+
+describe('ConfigInput – rendering', () => {
```
jocelynlin-wd: nit: can we wrap the entire test suites into a top level `ConfigInput` describe instead?

packages/agentflow/src/features/node-editor/EditNodeDialog.tsx
```diff
@@ -64,6 +65,29 @@ function EditNodeDialogComponent({ show, dialogProps, onCancel }: EditNodeDialog
...
+                updatedInputValues = { ...data.inputValues, [arrayContext.parentParamName]: currentArray }
+            } else {
+                // Top-level config
+                updatedInputValues = { ...data.inputValues, [configKey]: configValues }
+            }
+
+            updateNodeData(data.id, { inputValues: updatedInputValues })
+            setData({ ...data, inputValues: updatedInputValues })
+        },
+        [data, updateNodeData]
```
jocelynlin-wd: `data` is local state that changes on every edit, so this `useCallback` recreates on every keystroke, which defeats memoization. The `ConfigInput` component already uses `useRef` to avoid stale closures from `onConfigChange` — but the recreation still causes unnecessary re-renders down the tree. Consider using a ref-based pattern in `EditNodeDialog` as well, or removing data from the dependency array by reading it from a ref.

eg:
```
  const dataRef = useRef(data)
  dataRef.current = data

  const onConfigChange = useCallback(
      (configKey: string, configValues: Record<string, unknown>, arrayContext?: { parentParamName: string; arrayIndex: number
  }) => {
          const current = dataRef.current
          if (!current) return

          let updatedInputValues: Record<string, unknown>
          // ...rest stays the same, using `current` instead of `data`

          updateNodeData(current.id, { inputValues: updatedInputValues })
          setData({ ...current, inputValues: updatedInputValues })
      },
      [updateNodeData]
  )
```

Issue #5931, opened 2026-03-08
OpenRouter base64 image data discarded by Flowise


vladisimo:
### Describe the bug

When you use ChatOpenRouter to access a model that generates images, when the payload is expected via

{
  "choices": [
    {
      "message": {
        "role": "assistant",
        "content": "I've generated a beautiful sunset image for you.",
        "images": [
          {
            "type": "image_url",
            "image_url": {
              "url": "data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAA..."
            }
          }
        ]
      }
    }
  ]
}

This payload is not available in Flowise.

### To Reproduce

1. Create a ChatOpenRouter agent node
2. Turn on image uploads
3. Ask the agent to generate / update the image (or even create an image from scratch)

### Expected behavior

Response is completed, but base64 data is not passed via Flowise so my front-end can't save the data and present it to the user

### Use Method

None

### Operating System

None

### Browser

None

HenryHengZJ:
This will be fixed in this PR: https://github.com/FlowiseAI/Flowise/pull/5893

<img width="1432" height="786" alt="Image" src="https://github.com/user-attachments/assets/d5b47c98-f585-456a-b4c0-40ec00f81d8d" />

PR #5893, opened 2026-03-04
Feat/Reasoning Support, Migrate LangChain to v1, Upgrade AI SDKs and Models

HenryHengZJ:
This PR upgrades the entire LangChain ecosystem from v0.3.x to v1.x, updates related SDKs (OpenAI, Anthropic, Mistral, etc.), deprecates legacy LLM nodes in favor of Chat Models, removes obsolete agents, and adds reasoning/thinking support to the UI and backend.

#### LangChain v1 Upgrade (270 files changed, ~5,900 additions / ~5,300 deletions)
- **`@langchain/core`**: `0.3.61` → `1.1.20`
- **`langchain`**: `^0.3.5` → `1.2.18`
- **`@langchain/openai`**: `0.6.3` → `1.2.5`
- **`@langchain/anthropic`**: `0.3.33` → `1.3.20`
- **`@langchain/google-genai`**: `0.2.3` → `2.1.15`
- All other `@langchain/*` scoped packages upgraded to their respective v1 releases
- New packages added: `@langchain/deepseek`, `@langchain/classic`, `@langchain/redis`, `@langchain/tavily`, `@langchain/textsplitters`

#### SDK Upgrades
- **`openai`**: `4.96.0` → `6.19.0`
- **`@anthropic-ai/sdk`**: added at `^0.73.0`
- **`@mistralai/mistralai`**: `0.1.3` → `1.14.0`
- **`@aws-sdk/client-bedrock-runtime`**: `3.422.0` → `3.966.0`
- **`langsmith`**: `0.1.6` → `0.4.12`
- **`zod`**: `3.22.4` → `^3.25.76 || ^4`
- **`uuid`**: `^9.0.1` → `^10.0.0` (across all packages)

#### Deprecate Legacy LLM Nodes
All LLM nodes now display a `DEPRECATING` badge with a message pointing users to the equivalent Chat Model:
- AWS Bedrock, Azure OpenAI, Cohere, Fireworks, GoogleVertexAI, HuggingFace Inference, IBM Watsonx, Ollama, OpenAI, SambaNova, TogetherAI

#### Remove Obsolete Agents
- **AutoGPT** — fully removed
- **BabyAGI** — fully removed

#### Remove Deprecated Utilities
- **Google Generative AI Context Cache** — removed (`FlowiseGoogleAICacheManager`, `GoogleGenerativeAIContextCache`)
[...]

PR #5968, opened 2026-03-14
fix: extract content from AIMessage objects to preserve base64 images

haroldfabla2-hue:
## Summary

Fixes issue #5931 - OpenRouter base64 image data discarded by Flowise.

## Changes

When using ChatOpenRouter to access models that generate images, the base64 image data in the response is not passed through to Flowise's frontend because the handler was treating the output as a plain string instead of extracting the `content` property from AIMessage objects.

## Fix

Modified `handleChainEnd` in `packages/components/src/handler.ts` to:
1. Check if `outputs.output` is an object with a `content` property
2. Extract the content properly to preserve base64 image data

This ensures that when the LLM returns an AIMessage object with base64 image data, it's properly extracted and stored in the database for the frontend to access.

## Testing

- [x] Code compiles without errors
- [x] Handles string outputs (backward compatible)
- [x] Handles AIMessage object outputs (new behavior)

---
**Related Issue:** #5931

### DIFF
--- a/packages/components/src/handler.ts
+++ b/packages/components/src/handler.ts
@@ -500,9 +500,16 @@ class ExtendedLunaryHandler extends LunaryHandler {
         if (this.chatId && runId === this.currentRunId) {
             const answer = outputs.output
 
+            // Extract content properly - handle both string and AIMessage objects
+            // This fixes OpenRouter base64 image data being discarded
+            let content = answer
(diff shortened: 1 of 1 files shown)


HenryHengZJ:
resolved https://github.com/FlowiseAI/Flowise/pull/5893

PR #5971, opened 2026-03-15
feat: implement scheduling functionality for AgentflowV2

prd-hoang-doan:
## Ticket: 

https://github.com/orgs/FlowiseAI/projects/2?pane=issue&itemId=121600925

## Overview
Adds first-class scheduled execution for Agentflows, including UI inputs for schedule configuration and a backend scheduler that can run via BullMQ (queue mode) or an in-process timer (non-queue mode).

https://github.com/user-attachments/assets/9d988193-7ef7-4dd2-82d0-ccfadc78b291

## Changes

- Introduces schedule configuration inputs on the Agentflow Start node (cron + visual picker) and renders new picker controls in the canvas UI.
- Adds server-side schedule persistence (new entities + DB migrations) plus cron/visual-picker validation and schedule upsert/disable hooks on chatflow save/update/delete.
- Implements schedule execution via new ScheduleBeat (sync + non-queue timer) and ScheduleQueue (BullMQ repeatable jobs), and wires the queue into QueueManager/worker startup.

## Database

<h3><code dir="ltr"><span>schedule_record</span></code></h3>

Column | Type | Nullable | Default | Description
-- | -- | -- | -- | --
id | uuid | No | uuid_generate_v4() | Primary key
triggerType | varchar(32) | No | - | Type of trigger (AGENTFLOW, DOCUMENT_STORE)
targetId | varchar | No | - | Target entity id (chatflowId / documentStoreId)
nodeId | text | Yes | - | Optional node identifier
cronExpression | text | No | - | Cron schedule expression
timezone | varchar(64) | No | UTC | Schedule timezone
enabled | boolean | No | true | Enable / disable schedule
defaultInput | text | Yes | - | Default input payload
lastRunAt | timestamp | Yes | - | Last execution time
nextRunAt | timestamp | Yes | - | Next planned execution time
workspaceId | varchar | No | - | Workspace identifier
createdDate | timestamp | No | now() | Created timestamp
updatedDate | timestamp | No | now() | Updated timestamp

<h3><code dir="ltr"><span>schedule_trigger_log</span></code></h3>

Column | Type | Nullable | Default | Description
-- | -- | -- | -- | --
id | uuid | No | uuid_generate_v4() | Primary key
scheduleRecordId | varchar | No | - | Related schedule record id
triggerType | varchar(32) | No | - | Trigger type
targetId | varchar | No | - | Target entity id (chatflowId / documentStoreId)
executionId | varchar | Yes | - | Execution identifier
status | varchar(32) | No | - | Trigger result status
error | text | Yes | - | Error message if failed
elapsedTimeMs | integer | Yes | - | Execution duration
scheduledAt | timestamp | No | - | Scheduled execution time
workspaceId | varchar | No | - | Workspace identifier
createdDate | timestamp | No | now() | Created timestamp

<h4><code dir="ltr"><span>Notes</span></code></h4>

- triggerType is introduced for future extensibility if more schedulable targets are added.
- schedule_trigger_log is separated from schedule_record to keep execution history scalable.
- timezone defaults to UTC for predictable cron behavior.

## Recording Demo:
Youtube: https://youtu.be/MZ1ws7XqsBg
 
---
This PR includes a new migration for the scheduling entities. Please take a look when you have time, and feel free to share any feedback if changes are needed.


harshit-flowise:
Thanks @prd-hoang-doan for this PR. We are working on a similar request and I will be testing it this week. Will reach out to you soon.

jchui-wd:
@prd-hoang-doan Thanks for implementing this, this is great. Just a few things we want addressed before we merge. I split them up into bugs that need fixing and nice-to-have follow-ups.

## Bugs

### 1. `scheduledAt` always falls back to current time ([Start.ts:357](https://github.com/FlowiseAI/Flowise/pull/5971/changes#diff-47156420b4789ee934a0c814f0dec7ce3f3fd8fd6562334fee4bdc60942b6ec0R355))

`options.agentflowRuntime?.scheduledAt` is never populated, so the fallback `new Date().toISOString()` always fires. I think we can just remove it unless we actually pass `scheduledAt` through `agentflowRuntime`.

### 2. `ScheduleRecord` not cleaned up when ChatFlow is deleted

`deleteChatFlow` only calls [`disableSchedulesForTarget()`](https://github.com/FlowiseAI/Flowise/pull/5971/changes#diff-9f303f86b704bed75d8df925156d03bb6d16be2330f72759016a55a35b42e557R134) (sets `enabled = false`). The `ScheduleRecord` rows should be hard-deleted when their parent ChatFlow is removed. `ScheduleTriggerLog` rows are fine to keep.

### 3. Empty input gets injected into ChatMessage

If no `defaultInput` is set, an empty string (or `<p></p>` from the rich text editor) gets written into `ChatMessage`. Please add a guard in `buildAgentflow.ts` to skip empty/whitespace inputs and strip rich text wrapper tags before they hit `ChatMessage`.

### 4. Clear chat doesn't work for scheduled flows

When pressing clear chat in the chat window for scheduled flows, it only deletes the latest output since it deletes one `chatId` at a time. We should delete all chat messages related to the `chatFlowId`, not just the `chatId`.

### 5. There are a few merge conflicts that needs to be resolved.

---

## Nice to Haves

1. If possible, we'd like to replace the `setInterval` + custom cron parser with `node-cron` for `MODE=MAIN` since it's a more robust tool.
2. `nextRunAt` is never populated in `ScheduleRecord`,  it would be great to populate it after each run, add missed-run detection on startup, and wire up `SKIPPED` in `ScheduleTriggerLog` when the current date has passed `nextRunAt`. Note: `SKIPPED` is currently unused.
3. An enable/disable toggle so users can pause a schedule without switching input types and losing their config.
4. End date support for schedules, could be a date picker.
5. Unit tests would be great to have, we'll be fixing our current broken tests soon and can sort out the testing rubric together.

Let us know which ones you can pick up and we can handle any remaining ones. Thanks again!

prd-hoang-doan:
@jchui-wd Thanks for the quick and detailed feedback — I’m glad to receive it.
For the bugs (1–5), I’ll address all of them and aim to complete the fixes by the end of this week.
For the nice-to-have items:
1. Replacing the current cron implementation with node-cron in MAIN mode sounds like a good improvement — I’ll include that update as well.
2. Good catch on nextRunAt and SKIPPED handling — I missed that part, and I’ll follow your suggestion.
3. For the enable/disable toggle, do you have a preferred location in mind? I was thinking it may be better placed outside the Start dialog for better visibility (for example near the top of the canvas or toolbar).
Screenshot 2026-03-18 at 9 11 00 AM

4. I’ll also add an end date field so scheduled jobs can stop automatically when needed.
5. Regarding unit tests, since the current test structure is still being stabilized, would it be okay if I add them in a follow-up once the testing rubric is clearer?

Please let me know if you’d prefer me to prioritize any item differently. Thanks again!

harshit-flowise:
@prd-hoang-doan Let's just add the disable toggle in the red box area. Thanks a lot.

jchui-wd:
Thanks for this, any order is alright. We will update our CONTRIBUTING.MD with proper testing standard soon, will follow up shortly.

jchui-wd:
Hey @prd-hoang-doan, we updated our Contributing.md with a testing section:
https://github.com/FlowiseAI/Flowise/blob/main/CONTRIBUTING.md#testing

Let us know if you have any questions / concerns.

prd-hoang-doan:
@jchui-wd Sure, I’ll create the unit tests following your instruction. Thanks for the update.

prd-hoang-doan:
@jchui-wd @harshit-flowise , I updated the pull requests based on @jchui-wd suggestion. The PR includes bug fixes (except point #4). It seems that our current logic only deletes the last sessions (I tried opening multiple profiles with the same account). Users can delete all by viewing all messages, then clicking Delete All.
image

Besides, I also implemented  nice-to-have things:
1. Use "node-cron" for the main mode.
2. Use "SKIPED" status for any outdated schedule during the worker processing the job
3. Add the switch toggle on the header.
image

4. Add unit tests for the scheduled services.
image


Please take your time reviewing my pull requests, and feel free to share your feedback.

jchui-wd (dismissed):
Thanks @prd-hoang-doan for doing the changes. I think they look good

Below evidences and video of my tests in local.
<details>
<summary><b>Regular CRON / Visual Picker</b></summary>
Cron Toggle On<video src="https://github.com/user-attachments/assets/be952efe-3247-49dc-a3b4-ab1bb560e897" width="400"></video>
Cron Toggle Off<video src="https://github.com/user-attachments/assets/69452888-cd01-4dbc-b7e6-7cfdb4d4be84" width="400"></video>
Invalid Cron
<video src="https://github.com/user-attachments/assets/49123d30-3b2b-45f3-afa7-4c7bc15fb457
" width="400"></video>
Weekly<video src="https://github.com/user-attachments/assets/3c3df122-d928-42bd-9064-0c6e37d0f7a1" width="400"></video>
Monthly<video src="https://github.com/user-attachments/assets/6c2af3c7-fe8d-49db-a462-3db93c70c3e8" width="400"></video>
Hourly<video src="https://github.com/user-attachments/assets/5987bbd8-1e36-41d4-b68c-27f37a81ea26" width="400"></video>
Daily<video src="https://github.com/user-attachments/assets/34c6db24-d0b5-4579-817b-803784cfbe04" width="400"></video>
</details>

<details>
<summary><b>Server Restart Resumes Job / MODE=QUEUE</b></summary>
Server Restart Resumes Job<video src="https://github.com/user-attachments/assets/7425f071-bf00-4f96-ade9-9fcea68ebb2b" width="600"></video>
BullMQ / MODE=QUEUE<video src="https://github.com/user-attachments/assets/4488714c-04ca-43f5-b8e1-368a032d5545" width="600"></video>
</details>

<details>
<summary><b>Toggles Off After End Date / Cannot Enable Past EndDate</b></summary>
Toggles Off After End Date<video src="https://github.com/user-attachments/assets/b3856d2c-f94e-4647-9557-1b5070068640" width="600"></video>
Cannot Enable Past End Date<video src="https://github.com/user-attachments/assets/0a0f5699-f44e-447e-a04e-e553217980b4" width="600"></video>
</details>

<details>
<summary><b>Delete Flow or Change Inputs Removes ScheduleRecord</b></summary>
Delete Flow Deletes ScheduleRecord<video src="https://github.com/user-attachments/assets/feb2d72d-39af-4e1a-901d-535803326e92
" width="600"></video>
Change Input Deletes ScheduleRecord<video src="https://github.com/user-attachments/assets/1e46c22f-1c5b-490c-995e-7e451e5c4ae9" width="600"></video>
</details>

All Tests Passing
Screenshot 2026-03-23 at 5 43 46 PM

As for #4, we can deal and follow up with this internally.

Since this is a big feature and may overlap with some other of our ongoing work, we will have additional eyes later on. We are still coordinating a few things on our end so it might take a bit longer before we can merge it in. We will keep you posted.

EDIT: 
Actually seems like there is an error:
CI is failing because some date columns in ScheduleRecord and ScheduleTriggerLog use type: 'timestamp', which isn't supported by SQLite.
If you remove database configs in .env  just let it run on default using sqlite you can see the error when starting the server. Can you take a look?
Screenshot 2026-03-23 at 6 03 47 PM

prd-hoang-doan:
@jchui-wd Thank you for your effort to review my code, test the feature, and point out the issue. I will fix check the build on `sqlite` to make sure the migration is correct.

prd-hoang-doan:
@jchui-wd I push the latest commit to fix the issue with SQLite. I also tested with other database including mariadb, mysql as well. Please help me verify on your machine with SQLite database.

- SQLite
Screenshot 2026-03-24 at 20 57 17

- MariaDB
Screenshot 2026-03-24 at 21 07 26

- Unit tests:
image

jchui-wd (dismissed):
Thanks for applying the fix @prd-hoang-doan.
<details>
<summary>Confirmed running correctly on sql-lite, no errors</summary>
Image
</details>

I also noticed two small issues and just pushed a fix:
(35a5802): added nullable: true to lastRunAt/nextRunAt/endDate on ScheduleRecord (entity didn't match migration DDL), and added IScheduleRecord/IScheduleTriggerLog interfaces per codebase convention.

jchui-wd (dismissed):

packages/components/nodes/agentflow/Start/Start.ts
```diff
@@ -40,6 +40,11 @@ class Start_Agentflow implements INode {
                         label: 'Form Input',
                         name: 'formInput',
                         description: 'Start the workflow with form inputs'
+                    },
+                    {
+                        label: 'Schedule Input',
```
jchui-wd, line 76: @jocelynlin-wd, Hey just a quick question, will editing the Start.ts here also make it show up for `package/agentflow`? 

if so we probably need some additional work to make sure it also works for `package/agentflow`

jchui-wd: Confirmed with Jocelyn that editing Start.ts affects both the legacy canvas and the agentflow UI. We'd probably delay this merge to avoid unintended impacts on downstream or merge it in a separate feature branch.

HenryHengZJ:
Thank you @prd-hoang-doan for the changes! Apologies if this is taking longer since we have to let some other PRs to get in first, but this should give us enough time to test this thoroughly

prd-hoang-doan:
@HenryHengZJ Thanks for the update — no worries at all! I understand the process, and I appreciate you taking the time to test everything carefully. 
Happy to wait, and feel free to let me know if anything needs to be adjusted on my side.

jchui-wd:
@prd-hoang-doan hi again,

If you have time we would like some changes to help prevent cloud apps from being overloaded. 

**1. Enforce minimum 1-minute cron interval**
 node-cron v4 supports a 6th field (seconds), so * * * * * * would fire every second. We need to restrict to 5-field cron only.
- `packages/server/src/services/schedule/utils.ts` ~line 22 - validation currently accepts fields.length === 6, this should be rejected
- Same file ~line 84 — 6-field seconds handling can be removed
- `packages/server/src/services/schedule/utils.test.ts` ~line 32 - remove/update the 'accepts 6-field cron with seconds' test

**2. Schedule executions need to count against prediction quota**
- Right now `ScheduleExecutor` calls `executeAgentFlow` (`packages/server/src/utils/buildAgentflow.ts`) which doesn't go through `checkPredictions/updatePredictionsUsage` (those only live in `packages/server/src/utils/buildChatflow.ts`). Also orgId, subscriptionId, and productId are passed as empty strings in `packages/server/src/queue/ScheduleExecutor.ts` (lines 181-183). 
- Each schedule fire is essentially a prediction, so it should be counted against the existing `quota:predictions` limit on cloud.

prd-hoang-doan:
Hi @jchui-wd, I updated two mentioned points in the latest commit. Thank you for pointing out the usage issue. 
Please take a look and review it.

jchui-wd:
@prd-hoang-doan thanks for the quick turnaround.

Apologies, I have to go back on my **Item Number 1** request.

Instead of rejecting 6-field cron entirely, can you revert that change and replace it with a minimum interval check in? Here's the updated approach:

- Add a `MIN_SCHEDULE_INTERVAL_SECONDS` env variable to the server (default 60)
- Add a new route e.g. `GET /api/v1/schedule-config` that returns `{ minScheduleIntervalSeconds: 60 }` so the UI can fetch it
- When saving, if the cron interval is below `MIN_SCHEDULE_INTERVAL_SECONDS`. We can fire a error toast to tell them. "The CRON interval is less than the minimum interval X seconds".
- The server should block all enabling of runs that with CRON below the `MIN_SCHEDULE_INTERVAL_SECONDS` when saving.
- On the UI, disable the schedule toggle if the cron interval is less than `MIN_SCHEDULE_INTERVAL_SECONDS` with a warning message again "The cron interval is less than the minimum allowed interval of X seconds",  similar to how the toggle is already disabled when the end date is in the past

--- 
**Item Number 2** For this 

I just want to make sure incase theres a bug for OSS users. Flowise OSS works without account registration, meaning the workspace and organization tables can be empty. If a user never registered an account, findOneBy returns null and the schedule crashes with "Workspace not found".

Can you wrap the workspace/org lookup and quota check in a platform guard so OSS skips it entirely?
```
if (identityManager.isCloud() || identityManager.isEnterprise()) {
    // existing workspace/org lookup and checkPredictions/updatePredictionsUsage
}
```
Thanks again for your patience!

prd-hoang-doan:
Hi @jchui-wd, thank for your feedback, but I have some concerns below:
For #1, I think that we can use `MIN_SCHEDULE_INTERVAL_MINUTES` instead of SECONDS to avoid spamming jobs on the server. I tried using MIN_SCHEDULE_INTERVAL_SECONDS on my local, but it seems not to bring much value to validate */30 * * * * * (every 30 seconds) or  */90 * * * * * (every 90 seconds). So, I suggest use MINUTES to control the interval between job schedulers. How do you think about this?

For #2, I followed the same pattern for the function `utilBuildChatflow` function. For the OOS Users like me, I can do anything without enterprise features and run unlimited predictions. The OOS users still have to set up the first account with the default organization and default workspace since most APIs use the column for query to avoid cross-access security issues.
https://github.com/FlowiseAI/Flowise/blob/main/packages/server/src/utils/buildChatflow.ts#L1053-L1055
Could you double-check that?

jchui-wd:
@prd-hoang-doan 

Ah I should have given more context.

For **Number 1**, we wanted to have `MIN_SCHEDULE_INTERVAL_SECONDS` over minutes is because we want to give users the flexibility to run sub-minute crons if they need to, for our own cloud we'd set it to 60, but self-hosted users could set it to 1 or whatever works for them.

Although I realize this might affect `computeNextRunAt` since it uses `_cronMatchesParsed` to calculate `nextRunAt` in the db. And changing the `maxIterations` to be seconds instead of minutes may cause slight performance issues. 
`(60 * 24 * 366 -> 60 * 60 * 24 * 366)`

But I do think it is fine since from 1s - 60s the worse iteration per `computeNextRunAt` would be 60 iterations, which is not too bad. And the absolute worse `0 0 31 2 *` only on Feb 31st which contains 31 million iterations which can never happen cause its Feb 31st.

So I think it is fine to have `MIN_SCHEDULE_INTERVAL_SECONDS` and change the calculations for `maxIterations` to be seconds.

Let me know your thoughts.

For **Number 2**, I did a double take and you're right and it is not an issue, thanks!

prd-hoang-doan:
Hi @jchui-wd , I added `MIN_SCHEDULE_INTERVAL_SECONDS` env to allow the OSS user to control the interval between jobs. Users can set the interval as every second, or every 15 or 30 seconds.
If you have any feedback, feel free to let me know.

jchui-wd:
@HenryHengZJ I deployed and tested it on dev. Looking good, video evidence is posted in 391

packages/server/src/services/schedule/index.ts
```diff
@@ -0,0 +1,347 @@
...
+            // NOTE: Use assign + merge to update `endDate` and `nextRunAt` even if they are null
+            Object.assign(updateSchedule, bodySchedule)
+            const merged = repo.merge(existing, updateSchedule)
+            const saved = await repo.save(merged)
+            logger.debug(`[ScheduleService]: Updated schedule ${saved.id} for ${input.triggerType}:${input.targetId}`)
+            return saved
+        }
+
+        const record = repo.create({
+            id: uuidv4(),
```
HenryHengZJ: do we need to specifically assign id?

packages/ui/src/ui-component/picker/MonthDaysPicker.jsx
```diff
@@ -0,0 +1,87 @@
+import { useState, useEffect } from 'react'
+import PropTypes from 'prop-types'
+import { Box, Chip } from '@mui/material'
+import { useTheme } from '@mui/material/styles'
+
+const DAYS_OF_MONTH = Array.from({ length: 31 }, (_, i) => i + 1)
```
HenryHengZJ: can we also have last day? and update accordingly on the backend with this new logic

Image

packages/server/src/services/schedule/index.ts
prd-hoang-doan: I doubled check the function and remove unnecessary generated UUID here

packages/ui/src/ui-component/picker/MonthDaysPicker.jsx
prd-hoang-doan: Sure, happy to add support for configuring the last day of the month in the scheduler. 
I'm on vacation this week, so I'll pick it up next week and push the commit when it's ready.

HenryHengZJ: no worries man! enjoy your holiday, no rush. we can also help with that

HenryHengZJ:
In this [commit](https://github.com/FlowiseAI/Flowise/pull/5971/changes/7e7ac96d950dd8872bea209973e4470c3dd08861), here's what I've added:

**Three input modes for scheduled flows**
- `text` — current behavior; `defaultInput` becomes `incomingInput.question`.
- `form` — enable users to start the flow with form inputs
- `none` — fires with no input

**Other server changes**
- New `ChatType.SCHEDULED` so persisted `ChatMessage` rows are tagged for the View Messages dialog filter.
- `executeAgentFlow` call in `ScheduleExecutor` now uses `process.env.APP_URL` for `baseURL` (was empty), drops the unused `uploadedFilesContent` / `fileUploads` params.
- `getScheduleStatus` no longer falls back to a stale `record.defaultInput` — always validates against live flow inputs
- New endpoint `GET /chatflows/:id/schedule/trigger-logs` for the schedule history drawer display below

### UI
**Schedule History drawer**
- When Start is Schedule Input, hide the chat button but show the history logs button:
image

jchui-wd:
tagging @simon-song-wd and @igor-magun-wd for security review

igor-magun-wd (approved):

prd-hoang-doan:
In this [commit](https://github.com/FlowiseAI/Flowise/pull/5971/changes/a5025f7349251d568a52fefd52e485d7d67a9a69), here is what I did to support L (last day of moth) in cron scheduling:

## Changes:
1. `utils.ts` — core logic:

- `expandCronLForNodeCron()`: Translates L → **28-31** so **node-cron** can parse the expression.
- `cronDomMatchesNow()`: A runtime filter that checks if today is truly the last day of the month before firing, eliminating false positives from days **28-30** that aren't actually the last day.
- `_matchDomField()`: Extended the cron field matcher to understand the L token.
validateCronExpression() & validateVisualPickerFields(): Updated to accept L as a valid day-of-month value.

2. `ScheduleBeat.ts` — scheduling engine:
- When registering a cron job, expands L for node-cron but wraps the callback with a runtime cronDomMatchesNow() guard so the job only fires on the actual last day.

3. `MonthDaysPicker.jsx` — **UI**:
- Added an L ("Last Day") chip to the visual day picker, sorted to appear after day 31.

4. Update the unit tests for the last day
Screenshot 2026-04-30 at 9 04 41 AM


## Testing:
- Mode: main (node-cron)

https://github.com/user-attachments/assets/ec44ff2a-eaaa-4d20-87ad-35e352f27efa



- Mode: queue (BullMQ)

https://github.com/user-attachments/assets/da07a41b-e123-45d1-81fe-26944d8b0237

HenryHengZJ (approved):

HenryHengZJ:
coincidentally we are able to test this "last day" since today is last day of april 😆 

thank you so much @prd-hoang-doan for your effort! and @jchui-wd for reviewing, helping out!

PR #5973, opened 2026-03-15
fix: support Claude 4+ models in vision model check

majiayu000:
## Summary

The `setVisionModel()` method in both `FlowiseChatAnthropic.ts` and `FlowiseAWSChatBedrock.ts` only recognized `claude-3` models as vision-capable. Claude 4.x models (`claude-opus-4-0`, `claude-sonnet-4-0`, `claude-opus-4-1`, `claude-opus-4-5`, `claude-sonnet-4-5`, `claude-haiku-4-5`) were silently downgraded to `claude-3-5-haiku-latest` when image uploads were enabled.

Fixes #5586

## Changes

- Updated vision model check in `FlowiseChatAnthropic.ts` to recognize Claude 4+ model naming convention (`claude-opus-*`, `claude-sonnet-*`, `claude-haiku-*`)
- Applied same fix to `FlowiseAWSChatBedrock.ts` for AWS Bedrock Claude models
- Added 11 unit tests covering all Claude model families (3, 3.5, 3.7, 4, 4.1, 4.5) and fallback behavior for non-vision models

## Test plan

- [x] All 11 new vision model tests pass
- [x] All 459 existing tests pass
- [x] No lint errors in changed files

PR #5980, opened 2026-03-16
Chore/Remove Redundant Check for Image Processing

HenryHengZJ:
Previously, only certain models supported image input. To ensure a smooth UI/UX, we silently switched to a default vision model when images were detected. Now that most models support image input and legacy models have been deprecated, this redundant check is no longer necessary and has been removed.


HenryHengZJ:
https://github.com/FlowiseAI/Flowise/pull/5973

PR #5988, opened 2026-03-16
feat(agentflow): add Rich Text editor for Content Editing (FLOWISE-342)

jocelynlin-wd:
## Summary
- Add `RichTextEditor` atom wrapping TipTap with code block syntax highlighting (lowlight) for the agentflow SDK
- Update `ExpandTextDialog` to support a `richtext` mode alongside plain text
- Replace plain `TextField` in `MessagesInput` content fields with the new rich text editor
- Lazy-load `RichTextEditor` via `React.lazy` + `Suspense` and import only 4 highlight.js languages (js/json/py/ts) instead of `lowlight/common` (~400KB bundle savings)
- Add `syntaxHighlight` colors and `typography` constants to design tokens
- Update ESLint rules and ARCHITECTURE.md to allow atoms → `core/theme` imports

## Test plan
- [x] 39 unit tests passing across `RichTextEditor`, `ExpandTextDialog`, and `MessagesInput`
- [x] Verify TipTap editor renders correctly in the expand dialog (richtext mode)
- [x] Verify code block syntax highlighting works for JS, JSON, Python, TypeScript
- [x] Verify lazy loading — RichTextEditor chunk should not appear in initial bundle

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Bundle size analysis
- Impact on main entry (what consumers load upfront)

  | Baseline (no TipTap) | Now | Delta
-- | -- | -- | --
ES main | 631 KB (166 KB gz) | 633 KB (167 KB gz) | +2 KB (+0.3%)

The main entry is essentially unchanged — the only addition is the React.lazy() call and Suspense wrapper.

- Lazy chunk (loaded on demand, only when editor renders)

  | Size
-- | --
RichTextEditor chunk | 554 KB (146 KB gz)

- UMD (can't code-split)

  | Baseline | Now | Delta
-- | -- | -- | --
UMD | 408 KB (135 KB gz) | 762 KB (244 KB gz) | +354 KB (+87%)

Summary: For ES module consumers (the common case), the upfront cost is +2 KB. TipTap + highlight.js load lazily as a 554 KB chunk only when a RichTextEditor actually renders. The UMD build takes the full hit since it can't code-split.

## Recordings: 

https://github.com/user-attachments/assets/ef076e8f-ee78-4ce4-9326-6534a43fdad6

https://github.com/user-attachments/assets/fd90c31a-c2c0-41ae-be6f-d16f25f2172d

https://github.com/user-attachments/assets/aae2ded9-f9cd-4dea-8eb6-9d5485598a3f

### DIFF
--- a/packages/agentflow/src/atoms/MessagesInput.tsx
+++ b/packages/agentflow/src/atoms/MessagesInput.tsx
@@ -61,11 +61,12 @@ export function MessagesInput({ inputParam, data, disabled = false, onDataChange
 
     // Track latest inline content locally so the expand dialog always has fresh values,
     // even if the parent hasn't round-tripped onDataChange back into data yet.
-    const latestContentRef = useRef<Map<number, string>>(new Map())
+    // Keyed by itemKeysRef values so deletes are a simple Map.delete() with no index rebasing.
+    const latestContentRef = useRef<Map<string, string>>(new Map())
 
     const handleContentChange = useCallback(
         (index: number, content: string) => {
-            latestContentRef.current.set(index, content)
+            latestContentRef.current.set(itemKeysRef.current[index], content)
             const updated = [...messages]
             updated[index] = { ...updated[index], content }
             onDataChange?.({ inputParam, newValue: updated })
@@ -80,14 +81,8 @@ export function MessagesInput({ inputParam, data, disabled = false, onDataChange
 
     const handleDeleteMessage = useCallback(
         (indexToDelete: number) => {
+            latestContentRef.current.delete(itemKeysRef.current[indexToDelete])
             itemKeysRef.current.splice(indexToDelete, 1)
-            // Rebuild content ref with shifted indices
-            const newMap = new Map<number, string>()
-            latestContentRef.current.forEach((val, idx) => {
-                if (idx < indexToDelete) newMap.set(idx, val)
-                else if (idx > indexToDelete) newMap.set(idx - 1, val)
-            })
-            latestContentRef.current = newMap
             onDataChange?.({ inputParam, newValue: messages.filter((_, i) => i !== indexToDelete) })
         },
         [messages, inputParam, onDataChange]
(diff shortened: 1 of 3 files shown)


packages/agentflow/src/atoms/MessagesInput.tsx
```diff
@@ -77,6 +81,13 @@ export function MessagesInput({ inputParam, data, disabled = false, onDataChange
     const handleDeleteMessage = useCallback(
         (indexToDelete: number) => {
             itemKeysRef.current.splice(indexToDelete, 1)
+            // Rebuild content ref with shifted indices
+            const newMap = new Map<number, string>()
```
j-sanaa: Does it make sense to have the key of the map the same as itemKeysRef to avoid this index based delete logic:

const latestContentRef = useRef<Map<string, string>>(new Map())

// In handleContentChange:
const key = itemKeysRef.current[index]
latestContentRef.current.set(key, content)

// In handleDeleteMessage — just delete the key, no rebasing:
latestContentRef.current.delete(itemKeysRef.current[indexToDelete])

// When reading (expand dialog):
const key = itemKeysRef.current[index]
const value = latestContentRef.current.get(key) ?? messages[index]?.content ?? ''

packages/agentflow/src/atoms/RichTextEditor.lazy.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+const RichTextEditorLazy = lazy(() => import('./RichTextEditor').then((m) => ({ default: m.RichTextEditor })))
+
+/**
+ * Lazy-loaded RichTextEditor — keeps TipTap + highlight.js out of the main bundle.
+ * This is the public API; use this instead of importing RichTextEditor directly.
+ */
+export function RichTextEditor(props: RichTextEditorProps) {
+    return (
+        <Suspense fallback={<div style={{ minHeight: props.rows ? `${props.rows * 1.4375}rem` : '2.4rem' }} />}>
```
j-sanaa: Could also use tokens.typography.rowHeightRem here to keep in sync with tokens

packages/agentflow/src/atoms/MessagesInput.tsx
jocelynlin-wd: good idea, updated

packages/agentflow/src/atoms/RichTextEditor.lazy.tsx
jocelynlin-wd: good catch, updated

PR #5989, opened 2026-03-17
fix(agentflow): sanitize null message content to prevent crash when tool_calls present

eandresvera:
## Problem

When an agent uses tools, OpenAI returns `content: null` on the assistant message (this is valid per the OpenAI spec — `content` is optional when `tool_calls` is present). Flowise stores this message in the database as-is. On the next request, the stored history is loaded and sent back to the LLM, causing a crash.

The error manifests differently depending on the `@langchain/openai` version:

| `@langchain/openai` version | Error |
|---|---|
| `0.3.x` (older) | `400 Invalid value for 'content': expected a string, got null` |
| `0.5.x / 0.6.x` (newer) | `TypeError: Cannot read properties of null (reading 'map')` |

Both share the same root cause: `content: null` in a stored chat message.

## Root Cause

OpenAI returns { role: "assistant", tool_calls: [...], content: null }
↓
Flowise stores it in the DB
↓
Next request loads history and includes the null-content message
↓
@langchain/openai crashes trying to process it

The error only occurs intermittently because OpenAI sometimes includes text alongside tool calls and sometimes does not.

## Fix

Apply `?? ''` (nullish coalescing) at every point where messages are assembled before being sent to the LLM

## Testing
Reproduced the bug locally by:
1. Manually inserting a chat message with content: null into the DB
2. Sending a follow-up message → confirmed error: TypeError: Cannot read properties of null (reading 'map')
3. Applied the fix → rebuilt → error gone

### DIFF
--- a/packages/components/nodes/agentflow/Agent/Agent.ts
+++ b/packages/components/nodes/agentflow/Agent/Agent.ts
@@ -2375,7 +2375,7 @@ class Agent_Agentflow implements INode {
                 usedTools.push({
                     tool: toolCall.name || 'tool',
                     toolInput: toolCall.args,
-                    toolOutput: response.content
+                    toolOutput: response.content ?? ''
                 })
             }
         }


HenryHengZJ:
How to replicate this? Been testing OpenAI Agent with tool calling it works fine:
image

PR #5994, opened 2026-03-17
Feature/Add Audit and Deletion 

yau-wd:
FLOWISE-39


packages/components/src/storage/S3StorageProvider.ts
```diff
@@ -532,7 +532,17 @@ export class S3StorageProvider extends BaseStorageProvider {
                 config: this.s3Config
             })
             return [new transports.Stream({ stream: s3ServerReqStream })]
+        } else if (logType === 'audit') {
+            const instance = process.env.HOSTNAME || process.env.POD_NAME || String(process.pid)
+            const s3AuditStream = new S3StreamLogger({
+                bucket: this.bucket,
+                folder: 'logs/audit',
+                name_format: `audit-%Y-%m-%d-%H-%M-%S-%L-${instance}.log.jsonl`,
```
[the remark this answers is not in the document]
yau-wd: Following existing format.

packages/server/src/utils/sanitize.util.ts
```diff
@@ -40,3 +41,92 @@ export function sanitizeUser(user: Partial<User>) {
...
+    }
+
+    // Handle IPv6 - mask last 64 bits
+    if (ip.includes(':')) {
+        const parts = ip.split(':')
+        const masked = parts.slice(0, Math.ceil(parts.length / 2)).join(':')
+        return `${masked}:xxxx:xxxx:xxxx:xxxx`
+    }
+
+    return 'xxx.xxx.xxx.xxx'
```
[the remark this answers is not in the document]
yau-wd: [fix(sanitize.util.ts): correct IPv6 sanitization for audit log IP masking](https://github.com/FlowiseAI/Flowise/pull/5994/commits/351acb40947bfbab033086258771fd76cfc844d9)

wisdomdaoleo:
📊 Flowise 可视化 AI 工作流很强大！

如果你对更多 AI Agent 方案感兴趣，欢迎加入交流群：

🤖 Bot: @xinhaovip2_bot
👥 群组: https://t.me/atoxinhaovip2025

packages/server/src/enterprise/services/account.service.ts
```diff
@@ -624,4 +646,155 @@ export class AccountService {
...
+        user.name = UserStatus.DELETED
+        user.email = `deleted_${user.id}_${Date.now()}@deleted.flowise`
+        user.status = UserStatus.DELETED
+        user.credential = null
+        user.tokenExpiry = null
+        user.tempToken = null
+        await queryRunner.manager.save(User, user)
+
+        // Step 6: Cancel Stripe Subscription
+        await this.identityManager.cancelSubscription(organizaiton.subscriptionId)
```
mmattu-wd, line 850: should we move this after the commitTransaction? If for whatever reason the database doesn't commit, there is a case in which the user stays active without a subscription anymore

packages/ui/src/views/account/index.jsx
```diff
@@ -103,6 +104,7 @@ const AccountSettings = () => {
     const updateAdditionalSeatsApi = useApi(userApi.updateAdditionalSeats)
     const getCurrentUsageApi = useApi(userApi.getCurrentUsage)
     const logoutApi = useApi(accountApi.logout)
+    const deleteAccountApi = useApi(accountApi.deleteAccount)
```
mmattu-wd, line 108: should we confirm with a one time passcode or re-entering their password? Is there a chance an attacker can hijack a session and delete someone else's account?

packages/server/src/utils/telemetry.ts
```diff
@@ -36,3 +40,126 @@ export class Telemetry {
...
+            ipAddress: input.ipAddress ? sanitizeIPAddress(input.ipAddress) : undefined,
+            countryCode: geo.countryCode,
+            region: geo.region,
+            result: input.result,
+            metadata: input.metadata ? sanitizeAuditMetadata(input.metadata) : undefined
+        }
+
+        auditLogger.log({ level: 'info', message: event.eventType, ...event })
+    } catch (error) {
+        logger.error(`Failed to emit event: ${error}`)
```
mmattu-wd: When an event fails to emit, should we add it to a queue to be retried? Not an issue rn, but for MALT if we're making http requests, it can become an issue

packages/server/src/utils/sanitize.util.ts
```diff
@@ -40,3 +41,117 @@ export function sanitizeUser(user: Partial<User>) {
...
+        'socialsecurity',
+        'creditcard',
+        'cvv'
+    ]
+
+    const sanitized: Record<string, any> = { ...metadata }
+
+    for (const key of Object.keys(sanitized)) {
+        const lowerKey = key.toLowerCase()
+        if (sensitiveFields.some((field) => lowerKey.includes(field))) {
```
mmattu-wd: this would only look 1 level into the object. Is there any chance data like this gets emitted?:
```
{
  "configuration": {
       "apiKey": "someapikey"
   }
}
```

if not we should maybe update the metadata type to be something like `Record<string, string | number>`. So if we do try logging and object ts catches it, this can be circumvented with `any` but 🤷  . As it stands right now the above metadata would pass and log the value

packages/server/src/enterprise/services/account.service.ts
```diff
@@ -624,4 +646,155 @@ export class AccountService {
...
+
+        await queryRunner.commitTransaction()
+
+        // Step 7: Delete Organization Folder from Storage
+        await removeFolderFromStorage(organizaiton.id)
```
mmattu-wd, line 855: if this throws an error the account would be removed but we wouldn't have emitted an event

mmattu-wd (commented):
Still need to test the whole flow but left some comments

packages/server/src/enterprise/services/account.service.ts
HenryHengZJ: Why we didnt have try catch finally with `rollbackTransaction` and `release` same as other function at the top?

packages/ui/src/views/account/index.jsx
HenryHengZJ: for sso (google,github) login there isnt any passwords. from the standup we were thinking of something like a confirmation sentence: `permanently delete` or smtg like that

mmattu-wd: Ohh I see! I see google has something for this where you trigger OAuth for sensitive  actions [`prompt=consent`](https://developers.google.com/identity/openid-connect/openid-connect#re-consent). But github does not, we'd have to revoke their OAuth and re-trigger it which isn't ideal.

The only other option would be a OTP sent to the email

HenryHengZJ:
Steps:
1. Create User A and User B as Pro accounts.
2. User A creates Workspace 1 + Default Workspace.
3. User A invite User B to Default Workspace and Workspace 1. **Important!! must invite to both workspaces**
4. User B verify that he can see the invited Default Workspace and Workspace 1 **Important!! must be able to see both invited workspaces**
5. Let User B stays in User A's Default Workspace, on home page.
6. Now, User A delete account
7. Since User B is still in User A's Default Workspace, now when try to switch to User A's Workspace 1, will see 404 and thats expected:

image

8. Now, User B want to switch back to his own organization, thats when error happens:

```
2026-03-20 15:09:30 [INFO]: ⬆️ POST /api/v1/workspace/switch?id=83369a32-14c9-4595-8dae-d159b22dc050
2026-03-20 15:09:31 [ERROR]: uncaughtException:  Unhandled Edge Case
Error: Unhandled Edge Case
    at Statement.<anonymous> (C:\Users\Henry\Codes\Flowise\packages\server\dist\enterprise\controllers\workspace.controller.js:138:27)
    at Statement.<anonymous> (C:\Users\Henry\Codes\Flowise\node_modules\.pnpm\connect-sqlite3@0.9.15\node_modules\connect-sqlite3\lib\connect-sqlite3.js:131:32)
    at Statement.replacement (C:\Users\Henry\Codes\Flowise\node_modules\.pnpm\sqlite3@5.1.7\node_modules\sqlite3\lib\trace.js:25:27)
    at Statement.replacement (C:\Users\Henry\Codes\Flowise\node_modules\.pnpm\sqlite3@5.1.7\node_modules\sqlite3\lib\trace.js:25:27)
```

This is the part:
```
req.session.save((err) => {
                if (err) throw new InternalFlowiseError(StatusCodes.BAD_REQUEST, GeneralErrorMessage.UNHANDLED_EDGE_CASE)
            })
```

And database is now locked and every other operations will fail

packages/server/src/enterprise/services/account.service.ts
yau-wd: @HenryHengZJ 

I always place `queryRunner` on the parameter side so that the service can be reused by other controllers/modules by passing in the `queryRunner`.

Example: [workspace.controller.ts#L180](https://github.com/FlowiseAI/Flowise/blob/main/packages/server/src/enterprise/controllers/workspace.controller.ts#L180)

packages/ui/src/views/account/index.jsx
yau-wd: @HenryHengZJ @mmattu-wd 

## Confirmation Text
My opinion on using confirmation text such as `permanently delete` as the confirmation mechanism for account deletion. In a self-delete flow, the user is already authenticated and can only delete their own account, so the typed phrase appears to add limited security value and mainly introduces additional friction.

## OTP or Email Code Verification
My opinion on using OTP or email code verification as the primary confirmation method for either SSO users or email/password users. While it adds an extra validation step, it also makes the verification flow more complicated and introduces dependencies on resend countdown, email delivery, code expiry, and mailbox access.

## Password Confirmation
My proposed approach is to use password confirmation as the safeguard for account deletion.

For users who already have a password, require them to enter their password to confirm deletion. If the password is correct, proceed with the deletion immediately.

For SSO users who do not yet have a password, require them to set a password first. After that, they can use the same password confirmation flow to delete their account.

This gives us a simpler and more consistent deletion experience:
1. Logged-in user initiates self-delete.
2. User confirms by entering their password.
3. If the password is valid, deletion proceeds immediately.

Overall, this approach appears stronger than confirmation text alone and simpler than OTP or email code verification.

packages/server/src/utils/telemetry.ts
yau-wd, line 163: Continue in Jira.

packages/server/src/utils/telemetry.ts
```diff
@@ -36,3 +40,126 @@ export class Telemetry {
...
+    eventType: string
+    actionType: string
+    userId: string
+    orgId: string
+    resourceId?: string
+    ipAddress?: string
+    result: TelemetryEventResult
+    metadata?: Record<string, any>
+}
+export interface TelemetryEventOutput {
```
yau-wd, line 105: Continue in Jira.

packages/server/src/enterprise/services/account.service.ts
yau-wd: @mmattu-wd 

I placed commitTransaction after Stripe’s subscription cancellation to help protect the customer’s wallet.

If the commit fails while Stripe has already successfully canceled the subscription, the customer may not be able to log in, since the login flow would throw a subscriptionId not found error.

packages/server/src/utils/sanitize.util.ts
yau-wd: [fix(sanitize.util.ts): add recursive sanitization for nested metadata](https://github.com/FlowiseAI/Flowise/pull/5994/commits/4f9d0e408fc45d4c06f5c6619c30af2c76fb23c2)

packages/server/src/enterprise/services/account.service.ts
yau-wd: We should have observability at this point in [account.controller.ts line 158](https://github.com/FlowiseAI/Flowise/pull/5994/changes/BASE..9b249ffb79b131dc7c7217638b8879ea4cb9c678#diff-c8b890ac083aeafeafaa52fafea411f56d97313fe61fadd28aaebf9ec9d9dcad)

## manual
Let authorized personnel handle the deletion manually when needed.

## automation
Let a cron job that scans the folder structure and automatically deletes data when the orgId's owner is status `deleted` in the database.

yau-wd:
Error: [Henry 5994#issuecomment-4099315437](https://github.com/FlowiseAI/Flowise/pull/5994#issuecomment-4099315437)
Solution: [fix(MainLayout/Header/Workspace): refresh workspace list when clicking dropdown](https://github.com/FlowiseAI/Flowise/pull/5994/commits/018a133551fe5f23b140bbd54b5c89be4c5299ab)

## Cloud 
| user-a-workspaces | user-b-workspaces | user-b-in-user-a-workspace | user-a-delete-account | user-b-switch-back |
| --- | --- | --- | --- | --- |
| 01-user-a-workspaces | 02-user-b-workspaces | 03-user-b-in-usera-workspace | 04-user-a-delete-account | 05-user-b-switch-back |

## Enterprise
| owner-workspaces | invited-user-workspaces | owner-remove-user-from-workspace | invited-user-not-able-to-select-removed-workspace |
| --- | --- | --- | --- | 
| 01-owner-workspaces | 02-invited-user-workspaces | 03-owner-remove-user-from-a-workspace | 04-invited-user-not-able-to-select-removed-workspace |

packages/ui/src/views/account/index.jsx
mmattu-wd: It seems odd to ask a sso user to set a password. We could force the user to do an oauth flow, in which we callback to a page that confirms the users deletion.

The OTP although it introduces some dependencies is the typical flow for sensitive actions. We can always come back to it

yau-wd: In meeting everyone agree on **Confirmation Text**.

PR #6003, opened 2026-03-17
Feat/agentflow Support for Additional Load Methods

j-sanaa:
FLOWISE-266
- Fixed itemKeys generation for MessagesInput, ConditionBuilder, ArrayInput and StructuredOutputBuilder
- Added support for listActions and listTables
- Added a generic fallback method for loadMethods that have not been registered

image

image

### DIFF
--- a/packages/agentflow/src/infrastructure/api/loadMethodRegistry.test.ts
+++ b/packages/agentflow/src/infrastructure/api/loadMethodRegistry.test.ts
@@ -83,19 +83,21 @@ describe('loadMethodRegistry', () => {
     })
 
     describe('listRegions', () => {
-        it('should call nodesApi.loadNodeMethod() with nodeName and listRegions', async () => {
+        it('should call nodesApi.loadNodeMethod() with nodeName, listRegions, and currentNode.inputs', async () => {
             const mockRegions = [{ name: 'us-east-1', label: 'US East (N. Virginia)' }]
             ;(mockApis.nodesApi.loadNodeMethod as jest.Mock).mockResolvedValue(mockRegions)
 
-            const result = await loadMethodRegistry['listRegions'](mockApis, { nodeName: 'awsChatBedrock' })
-            expect(mockApis.nodesApi.loadNodeMethod).toHaveBeenCalledWith('awsChatBedrock', 'listRegions')
+            const fn = getLoadMethod('listRegions')
+            const result = await fn(mockApis, { nodeName: 'awsChatBedrock' })
+            expect(mockApis.nodesApi.loadNodeMethod).toHaveBeenCalledWith('awsChatBedrock', 'listRegions', {
+                currentNode: { inputs: {} }
+            })
             expect(result).toEqual(mockRegions)
         })
 
         it('should reject when nodeName param is missing', async () => {
-            await expect(loadMethodRegistry['listRegions'](mockApis)).rejects.toThrow(
-                '`listRegions` requires a string `nodeName` parameter.'
-            )
+            const fn = getLoadMethod('listRegions')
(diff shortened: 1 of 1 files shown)


packages/agentflow/src/atoms/useStableKeys.ts
```diff
@@ -0,0 +1,45 @@
...
+ * render pass that triggered the growth — calling setState alone would only
+ * take effect on the *next* render, leaving keys undefined in the current one.
+ *
+ * Safe in concurrent rendering: idCounterRef may skip numbers if a render is
+ * abandoned, but keys remain unique.
+ *
+ * @param length  Current list length (e.g. `arrayItems.length`)
+ * @param prefix  Key prefix string (e.g. `'item'` → `'item-0'`, `'item-1'`, …)
+ */
+export function useStableKeys(length: number, prefix: string): { keys: string[]; removeKey: (index: number) => void } {
```
jocelynlin-wd, line 18: we might need test for this file?

packages/agentflow/src/atoms/useStableKeys.ts
```diff
@@ -0,0 +1,45 @@
...
+    let effectiveKeys = itemKeys
+
+    if (effectiveKeys.length < length) {
+        const nextKeys = [...effectiveKeys]
+        while (nextKeys.length < length) {
+            nextKeys.push(`${prefix}-${idCounterRef.current++}`)
+        }
+        setItemKeys(nextKeys)
+        effectiveKeys = nextKeys
+    }
```
jocelynlin-wd, line 38: The original ArrayInput code explicitly trimmed `effectiveKeys` when `effectiveKeys.length > arrayItems.length` but it's dropped here?
```
  // Old code in ArrayInput.tsx
  } else if (effectiveKeys.length > arrayItems.length) {
      const trimmed = effectiveKeys.slice(0, arrayItems.length)
      setItemKeys(trimmed)
      effectiveKeys = trimmed
  }
```

If an external state change reduces the list (e.g. loading a saved node with fewer items, or programmatic reset), the internal `itemKeys` state will be longer than length. The keys won't cause visible React errors (rendering only uses `effectiveKeys[index]` up to length), but state drifts and the old keys accumulate. With enough add/remove/reload cycles this could subtly cause stale keys to resurface if `idCounterRef` wraps (unlikely in practice, but the old code was stricter).

packages/agentflow/src/infrastructure/api/loadMethodRegistry.ts
```diff
@@ -62,17 +66,48 @@ export const loadMethodRegistry: Record<string, (_apis: ApiServices, _params?: R
...
+        return apis.nodesApi.loadNodeMethod(nodeName, 'listActions', { currentNode: { inputs } })
+    },
+    listTables: (apis, params) => {
+        const nodeName = params?.nodeName
+        if (typeof nodeName !== 'string') {
+            return Promise.reject(new Error('`listTables` requires a string `nodeName` parameter.'))
+        }
+        const inputs = (params?.inputs as Record<string, unknown>) ?? {}
+        return apis.nodesApi.loadNodeMethod(nodeName, 'listTables', { currentNode: { inputs } })
     }
```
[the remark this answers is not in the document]
jocelynlin-wd: @j-sanaa  let's use default (fallback) unless the implemenetation have meaningfully different behavior or better error messages?

jocelynlin-wd (approved):
no blockers

packages/agentflow/src/atoms/useStableKeys.ts
jocelynlin-wd: https://github.com/FlowiseAI/Flowise/pull/6003#discussion_r2949893088

PR #6005, opened 2026-03-18
feat(agentFlow) - Add support for listFlows

j-sanaa:
FLOWISE-266
Added listFlows to the registry via chatflowsApi.getAllChatflows(),wired chatflowsApi into ApiServices and useAsyncOptions
Added test cases for chatFlows

https://github.com/user-attachments/assets/ec5137ed-9157-4d60-8238-a9121be7f95c

### DIFF
--- a/packages/agentflow/src/infrastructure/api/loadMethodRegistry.ts
+++ b/packages/agentflow/src/infrastructure/api/loadMethodRegistry.ts
@@ -41,6 +41,12 @@ export interface ApiServices {
  *   requires `params.nodeName` and accepts optional `params.inputs` forwarded as `currentNode.inputs`
  *
  */
+function getChatflowTypeLabel(type: string | undefined): string {
+    if (type === 'AGENTFLOW') return 'Agentflow V2'
+    if (type === 'MULTIAGENT') return 'Agentflow V1'
+    return 'Chatflow'
+}
+
 export const loadMethodRegistry: Record<string, (_apis: ApiServices, _params?: Record<string, unknown>) => Promise<unknown>> = {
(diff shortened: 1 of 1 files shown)


packages/agentflow/src/infrastructure/api/loadMethodRegistry.ts
```diff
@@ -31,13 +33,20 @@ export interface ApiServices {
...
  * - `listCredentials` — fetches credentials filtered by `params.name` via `GET /credentials?credentialName=<name>`
  * - `listActions` — fetches available actions for a node (e.g. Composio, MCP tools) via `POST /node-load-method/{nodeName}`;
  *   requires `params.nodeName` and accepts optional `params.inputs` forwarded as `currentNode.inputs`
  * - `listTables` — fetches available tables for a node (e.g. AWSDynamoDBKVStorage) via `POST /node-load-method/{nodeName}`;
  *   requires `params.nodeName` and accepts optional `params.inputs` forwarded as `currentNode.inputs`
  *
  */
+function getChatflowTypeLabel(type: string | undefined): string {
+    if (type === 'AGENTFLOW') return 'Agentflow V2'
+    if (type === 'MULTIAGENT') return 'Agentflow V1'
```
jocelynlin-wd, line 46: nit: are these shown on the UI?  might be good to use a const map or use switch but given we only have 3 types right now it's fine to keep current implementation.

j-sanaa: It is visible in the UI in the dropdown. Will leave this as it is for now then!

image

PR #3288, opened 2024-09-30
Bugfix/Upsert files extension to input field

HenryHengZJ:

PR #6007, opened 2026-03-18
fix(server): use fileInputFieldFromMimeType in MIME type fallback branch

wishhyt:
## Summary
When processing file uploads, the code determines the correct input field using a two-step fallback: first by file extension, then by MIME type. However, in the MIME type fallback branch (else if), the code incorrectly assigns fileInputFieldFromExt (which is 'txtFile' at that point) instead of fileInputFieldFromMimeType.

This means when a file has an unrecognized extension but a recognized MIME type, the MIME type detection result is silently ignored and the file is always treated as a text file.

The same copy-paste bug exists in 4 files:

- packages/server/src/utils/buildChatflow.ts
- packages/server/src/utils/createAttachment.ts
- packages/server/src/services/documentstore/index.ts
- packages/server/src/utils/upsertVector.ts
## Changes
Changed fileInputField = fileInputFieldFromExt to fileInputField = fileInputFieldFromMimeType in the else if (fileInputFieldFromMimeType !== 'txtFile') branch across all 4 files.

## How to reproduce the bug
- Upload a file with an unusual extension but a standard MIME type (e.g., a .dat file with image/png MIME type)
- The file is incorrectly classified as txtFile instead of the correct image input field

### DIFF
--- a/packages/server/src/services/documentstore/index.ts
+++ b/packages/server/src/services/documentstore/index.ts
@@ -1863,7 +1863,7 @@ const upsertDocStore = async (
             if (fileInputFieldFromExt !== 'txtFile') {
                 fileInputField = fileInputFieldFromExt
             } else if (fileInputFieldFromMimeType !== 'txtFile') {
-                fileInputField = fileInputFieldFromExt
+                fileInputField = fileInputFieldFromMimeType
             }
 
             if (loaderId === 'unstructuredFileLoader') {
--- a/packages/server/src/utils/buildChatflow.ts
+++ b/packages/server/src/utils/buildChatflow.ts
@@ -448,7 +448,7 @@ export const executeFlow = async ({
             if (fileInputFieldFromExt !== 'txtFile') {
                 fileInputField = fileInputFieldFromExt
             } else if (fileInputFieldFromMimeType !== 'txtFile') {
-                fileInputField = fileInputFieldFromExt
+                fileInputField = fileInputFieldFromMimeType
             }
 
             if (overrideConfig[fileInputField]) {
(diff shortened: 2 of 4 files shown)


HenryHengZJ:
@0xi4o is this valid? I traced back the history of the file it didnt looks like we accidentally changed it before

wishhyt:
Hi @HenryHengZJ , thanks for looking into this!

You're right that the code wasn't accidentally changed — the bug was introduced in the original commit that added this logic. Specifically, it was introduced in PR #3288 (commit 04e8d02a, Sep 30, 2024), which refactored the file input field detection from a simple MIME-type-only approach to a two-step "extension first, MIME type fallback" approach.

Before PR #3288:
```
const fileInputField = mapMimeTypeToInputField(file.mimetype)
overrideConfig[fileInputField] = storagePath
```

After PR #3288 (current code):
```
let fileInputField = 'txtFile'

if (fileInputFieldFromExt !== 'txtFile') {
    fileInputField = fileInputFieldFromExt          // ← extension matched
} else if (fileInputFieldFromMimeType !== 'txtFile') {
    fileInputField = fileInputFieldFromExt           // ← BUG: should be fileInputFieldFromMimeType
}
```

Why the else if branch is effectively dead code:

Entering the else if means the if condition failed, so fileInputFieldFromExt === 'txtFile'. Assigning fileInputField = fileInputFieldFromExt is equivalent to fileInputField = 'txtFile', which is already the default. The MIME type detection result is never used.

The intent of the two-step logic was clearly: try extension → fall back to MIME type → default to 'txtFile'. The variable name in the fallback line should be fileInputFieldFromMimeType, not fileInputFieldFromExt.

0xi4o:
> Upload a file with an unusual extension but a standard MIME type (e.g., a .dat file with image/png MIME type)

@wishhyt uploading a file with an extension that doesn't match its mime type is not possible now because we're validating the extension and mimetype before we get to the lines in question:

```ts
// Validate file extension, MIME type, and content to prevent security vulnerabilities
validateFileMimeTypeAndExtensionMatch(file.originalname, file.mimetype)
```

So the fallback will never happen if extension and mime type doesn't match.

PR #6014, opened 2026-03-18
Bugfix/Update qdrant dep

HenryHengZJ:


hkulekci:
Hello @HenryHengZJ, thanks for this upgrade on the Qdrant client, and I just want to check that Qdrant has some deprecations between these versions. Could you check whether you have any dependencies on these methods? 

https://api.qdrant.tech/api-reference/search/points

One of the biggest changes is that `/collections/:collection_name/points/search` endpoint has changed to`/collections/:collection_name/points/query`, and some other endpoints as well!

PR #6029, opened 2026-03-20
Support CORS credentials (cookies) via env var

jhead:
Adds support for enabling `Access-Control-Allow-Credentials` via a new `CORS_ALLOW_CREDENTIALS` environment variable. This is useful when running Flowise behind an authenticating proxy that requires cookies, etc.

Removes an existing middleware that attempted to enable this feature but never worked because the `cors()` middleware always intercepted and responded to the preflight requests first, so instead I've moved the credentials config to the main cors middleware.

Env var logic defaults to false.

### DIFF
--- a/docker/worker/.env.example
+++ b/docker/worker/.env.example
@@ -74,6 +74,7 @@ WORKER_PORT=5566
 ############################################################################################################
 
 # NUMBER_OF_PROXIES= 1
+# CORS_ALLOW_CREDENTIALS=false
 # CORS_ORIGINS=*
 # IFRAME_ORIGINS=*
 # FLOWISE_FILE_SIZE_LIMIT=50mb
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -74,6 +74,7 @@ PORT=3000
 ############################################################################################################
 
 # NUMBER_OF_PROXIES= 1
+# CORS_ALLOW_CREDENTIALS=false
 # CORS_ORIGINS=*
 # IFRAME_ORIGINS=*
 # FLOWISE_FILE_SIZE_LIMIT=50mb
--- a/packages/server/.env.example
+++ b/packages/server/.env.example
@@ -73,6 +73,7 @@ PORT=3000
 ############################################################################################################
 
 # NUMBER_OF_PROXIES= 1
+# CORS_ALLOW_CREDENTIALS=false
 # CORS_ORIGINS=*
 # IFRAME_ORIGINS=*
 # FLOWISE_FILE_SIZE_LIMIT=50mb


packages/server/src/utils/XSS.ts
```diff
@@ -25,6 +25,10 @@ export function getAllowedCorsOrigins(): string {
     return process.env.CORS_ORIGINS ?? ''
 }
 
+export function getAllowCredentials(): boolean {
+    return process.env.CORS_ALLOW_CREDENTIALS === 'true'
```
jocelynlin-wd, line 29: shall we update https://github.com/FlowiseAI/Flowise/blob/main/CONTRIBUTING.md#-env-variables to include this new env var?

packages/server/src/index.ts
```diff
@@ -205,11 +205,6 @@ export class App {
         // Add the sanitizeMiddleware to guard against XSS
         this.app.use(sanitizeMiddleware)
 
-        this.app.use((req, res, next) => {
```
jocelynlin-wd, line 208: I like the approach in this PR in general (better controls with env var and the change is made at the precise area `corsOptions`), just wondering if anything would be impacted with the removal of this block

jhead: The removed middleware was only ever partially effective. For any request that triggers a preflight (POST with application/json, custom headers like content-type/sentry-trace) — which is the vast majority of Flowise UI traffic — it never worked at all, because cors() intercepts OPTIONS and ends the response before the manual header middleware ran. The header only reached the client for "simple" requests (GET/HEAD, or POST with form-encoded/plain-text bodies), which is a narrow slice that doesn't cover authenticated API flows from what I can tell.

In practice, any integration that needed credentialed cross-origin access to non-simple requests was already broken. Defaulting to false and requiring explicit opt-in via CORS_ALLOW_CREDENTIALS=true is a stricter, more intentional posture with hopefully negligible regression for existing users.

But I'm open to feedback from others if anyone disagrees.

packages/server/src/utils/XSS.ts
jhead: good call, thanks! I also added it to the example .env files.

docker/worker/.env.example
```diff
@@ -74,6 +74,7 @@ WORKER_PORT=5566
 ############################################################################################################
 
 # NUMBER_OF_PROXIES= 1
+# CORS_ALLOW_CREDENTIALS=false
```
christopherholland-workday, line 77: It'd be great if a PR could also be added to updated this documentation: https://github.com/FlowiseAI/FlowiseDocs/blob/main/en/configuration/environment-variables.md?plain=1#L181-L190

christopherholland-workday (approved):

packages/server/src/index.ts
HenryHengZJ: will this affect users that have embedded chatbot on their website? https://docs.flowiseai.com/using-flowise/embed#cors

jhead: it probably will, i.e. users need to set the env var to true going forward.

can I suggest that I open a PR to update the documentation?

defaulting to false seems safer because not all Flowise users will be using the embedded chat? so it seems safer to assume we _don't_ want credentials to be passed cross-origin.

I'm open to other POVs though.

PR #6044, opened 2026-03-23
Add dynamic headers option to MCP core for tool call-time

michaell-workday:
Add option for dynamic headers to MCP core.ts. 

Unit tests passed.

### DIFF
--- a/packages/components/nodes/tools/MCP/core.ts
+++ b/packages/components/nodes/tools/MCP/core.ts
@@ -23,8 +23,11 @@ export class MCPToolkit extends BaseToolkit {
         this.transportType = transportType
     }
 
-    // Method to create a new client with transport
-    async createClient(dynamicHeaders: Record<string, string> = {}): Promise<Client> {
+    /**
+     * Creates a new MCP client and connects it via the configured transport.
+     * @param injectHeaders - Additional HTTP headers merged over static `serverParams.headers` for this connection. Used to pass per-invocation headers (e.g. from {@link getToolCallHeaders}) into SSE/HTTP transports.
+     */
+    async createClient(injectHeaders: Record<string, string> = {}): Promise<Client> {
         const client = new Client(
             {
                 name: 'flowise-client',
@@ -56,7 +59,7 @@ export class MCPToolkit extends BaseToolkit {
 
             const baseUrl = new URL(this.serverParams.url)
             await checkDenyList(this.serverParams.url)
-            const mergedHeaders = { ...this.serverParams?.headers, ...dynamicHeaders }
+            const mergedHeaders = { ...this.serverParams?.headers, ...injectHeaders }
             const headers = Object.keys(mergedHeaders).length > 0 ? mergedHeaders : undefined
             try {
                 if (headers) {
@@ -151,6 +154,7 @@ export async function MCPTool({
 }): Promise<Tool> {
     return tool(
         async (input): Promise<string> => {
+            // Create a new client for this request
             const toolCallHeaders = await toolkit.getToolCallHeaders?.()
             const client = await toolkit.createClient(toolCallHeaders)
 


packages/components/nodes/tools/MCP/core.ts
```diff
@@ -54,28 +56,30 @@ export class MCPToolkit extends BaseToolkit {
 
             const baseUrl = new URL(this.serverParams.url)
             await checkDenyList(this.serverParams.url)
+            const mergedHeaders = { ...this.serverParams?.headers, ...dynamicHeaders }
```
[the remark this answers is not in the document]
michaell-workday: I believe this is a hallucination. In browser console, `({ ...undefined, ...undefined })` evaluates to `{}`. `({ ...null })` evaluates to `{}`.

packages/components/nodes/tools/MCP/core.ts
```diff
@@ -147,8 +151,8 @@ export async function MCPTool({
 }): Promise<Tool> {
     return tool(
         async (input): Promise<string> => {
-            // Create a new client for this request
-            const client = await toolkit.createClient()
+            const dynamicHeaders = toolkit.getDynamicHeaders ? await toolkit.getDynamicHeaders() : {}
+            const client = await toolkit.createClient(dynamicHeaders)
```
jocelynlin-wd: this creates inconsistent behaviour: `initialize` function (line 103-105) that also calls `this.createClient()` but doesn't merge dynamic header.

packages/components/nodes/tools/MCP/core.ts
```diff
@@ -147,8 +151,8 @@ export async function MCPTool({
 }): Promise<Tool> {
     return tool(
         async (input): Promise<string> => {
-            // Create a new client for this request
-            const client = await toolkit.createClient()
+            const dynamicHeaders = toolkit.getDynamicHeaders ? await toolkit.getDynamicHeaders() : {}
```
jocelynlin-wd: minor: `getDynamicHeaders` fallback seems to be inconsistent, here a default value `{}` is assigned when there is no dynamicHeader, but the `createClient()` method seems to accept `undefined` for the parameter. Would assigning a default in createClient make the expectation more consistent? 

```
createClient(dynamicHeaders: Record<string, string> = {}) 
```

and here you can simply do:
```
const dynamicHeaders = await toolkit.getDynamicHeaders?.()
```

packages/components/nodes/tools/MCP/core.ts
michaell-workday: introduced more precise naming

packages/components/nodes/tools/MCP/core.ts
michaell-workday: good catch, done

packages/components/nodes/tools/MCP/core.ts
```diff
@@ -15,14 +15,16 @@ export class MCPToolkit extends BaseToolkit {
...
+    getToolCallHeaders?: () => Promise<Record<string, string>>
     constructor(serverParams: StdioServerParameters | any, transportType: 'stdio' | 'sse') {
         super()
         this.serverParams = serverParams
         this.transportType = transportType
     }
 
     // Method to create a new client with transport
-    async createClient(): Promise<Client> {
+    async createClient(dynamicHeaders: Record<string, string> = {}): Promise<Client> {
```
jocelynlin-wd:
```suggestion
    async createClient(toolCallHeaders: Record<string, string> = {}): Promise<Client> {
```

packages/components/nodes/tools/MCP/core.ts
jocelynlin-wd:
```suggestion
            const mergedHeaders = { ...this.serverParams?.headers, ...toolCallHeaders }
```

packages/components/nodes/tools/MCP/core.ts
jocelynlin-wd: actually we can keep the parameter name more generic here

packages/components/nodes/tools/MCP/core.ts
```diff
@@ -15,14 +15,16 @@ export class MCPToolkit extends BaseToolkit {
...
     transportType: 'stdio' | 'sse'
+    /** Per-invocation HTTP headers injected at tools/call time; overrides static toolkit headers for the same names. */
+    getToolCallHeaders?: () => Promise<Record<string, string>>
     constructor(serverParams: StdioServerParameters | any, transportType: 'stdio' | 'sse') {
         super()
         this.serverParams = serverParams
         this.transportType = transportType
     }
 
     // Method to create a new client with transport
```
jocelynlin-wd: maybe add comment on what this parameter is being used too?  ideally jsdoc would be better

packages/components/nodes/tools/MCP/core.ts
```diff
@@ -147,8 +151,8 @@ export async function MCPTool({
 }): Promise<Tool> {
     return tool(
         async (input): Promise<string> => {
-            // Create a new client for this request
```
jocelynlin-wd, line 150: can we keep the comment here?

packages/components/nodes/tools/MCP/core.ts
michaell-workday: good idea

packages/components/nodes/tools/MCP/core.ts
michaell-workday: I've changed it to `injectHeaders` as I found dynamicHeaders a little vague and abstract. "inject" is more descriptive/precise.

PR #6045, opened 2026-03-23
Add Permission Checks to Vector Store API's

christopherholland-workday:
Flowise-374


HenryHengZJ:
note: these endpoints will be removed soon as openai has deprecated assistants api

PR #6050, opened 2026-03-24
Fix IDOR in Evaluators and Evaluations Endpoints

christopherholland-workday:
Flowise-323 (See comments on Evaluation and Evaluators)

### DIFF
--- a/packages/server/src/services/evaluations/index.ts
+++ b/packages/server/src/services/evaluations/index.ts
@@ -1,25 +1,26 @@
-import { StatusCodes } from 'http-status-codes'
 import { EvaluationRunner, ICommonObject } from 'flowise-components'
-import { getRunningExpressApp } from '../../utils/getRunningExpressApp'
-import { InternalFlowiseError } from '../../errors/internalFlowiseError'
-import { getErrorMessage } from '../../errors/utils'
+import { StatusCodes } from 'http-status-codes'
+import { In } from 'typeorm'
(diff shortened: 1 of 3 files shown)


packages/server/src/services/evaluations/index.ts
```diff
@@ -66,7 +66,13 @@ const createEvaluation = async (body: ICommonObject, baseURL: string, orgId: str
     try {
         const appServer = getRunningExpressApp()
         const newEval = new Evaluation()
-        Object.assign(newEval, body)
```
0xi4o, line 69: We can still use `Object.assign(newEval, body)` and overwrite selected fields like this. No need to set individual fields.

```ts
Object.assign(newEval, body, {
  workspaceId,
})
```

HenryHengZJ: do we plan to use the new `stripProtectedFields` for the rest of the PRs ?

yau-wd: I will make all the changes under FLOWISE-323

yau-wd: [chore(services/evaluations & EvaluatorDTO): replace explicit field assignment with stripProtectedFields](https://github.com/FlowiseAI/Flowise/pull/6050/commits/5a6d99b74d66aedde0909cc062fc9a2a49f160f0)

PR #6059, opened 2026-03-25
Fix Mass Assignment on Save Custom Template

christopherholland-workday:
Flowise-323 (See comments on CustomTemplate)

### DIFF
--- a/packages/server/src/services/marketplaces/index.ts
+++ b/packages/server/src/services/marketplaces/index.ts
@@ -208,7 +208,13 @@ const saveCustomTemplate = async (body: any): Promise<any> => {
         let flowDataStr = ''
         let derivedFramework = ''
         const customTemplate = new CustomTemplate()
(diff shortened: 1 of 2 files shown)


packages/server/src/services/marketplaces/index.ts
```diff
@@ -208,7 +208,13 @@ const saveCustomTemplate = async (body: any): Promise<any> => {
         let flowDataStr = ''
         let derivedFramework = ''
         const customTemplate = new CustomTemplate()
-        Object.assign(customTemplate, body)
```
0xi4o, line 211: Is this change still required since we're overwriting `body.workspaceId` in the controller?

christopherholland-workday:
Something went wrong and the branches diverged, I have created a new PR addressing the fixes. Moved to https://github.com/FlowiseAI/Flowise/pull/6129

PR #6067, opened 2026-03-26
Feat/agentflow add support for Variable System & Syntax Support

j-sanaa:
- Add VariableInput component with TipTap-based {{ mention autocomplete, SuggestionDropdown, and CustomMention extension ported from the UI package, using HTML serialization for reliable round-tripping of mention nodes
- Expand useAvailableVariables hook to include all variable categories from the original UI: current_date_time, runtime_messages_length, loop_count, $flow.sessionId, $flow.chatId, $flow.chatflowId
- Fix field visibility on first render by merging param defaults into data.inputValues so show/hide conditions (e.g., llmEnableMemory: true) evaluate correctly before any user interaction
- Match original UI styling: remove icons from suggestion dropdown category headers, use green mention chip colors, and replace the variable picker popover with a tooltip hint on fields that have inline autocomplete
- Add comprehensive tests for customMention, suggestionConfig, VariablePicker, and useAvailableVariables (37 new tests, 937 total passing)

Recording for adding Variables

https://github.com/user-attachments/assets/363aa9ac-7e58-40f7-b883-91197dbcb650

Recording for Node outputs being updated correctly depending on upstream and downstream nodes

https://github.com/user-attachments/assets/6615cf20-c11d-42ca-ba4d-28ef594234e9

### DIFF
--- a/packages/agentflow/src/atoms/VariableInput.tsx
+++ b/packages/agentflow/src/atoms/VariableInput.tsx
@@ -162,7 +162,7 @@ export function VariableInput({ value, onChange, placeholder, disabled = false, 
                               return [
                                   'span',
                                   mergeAttributes(this.HTMLAttributes ?? {}, options.HTMLAttributes),
-                                  `${options.suggestion?.char ?? '{{'}${node.attrs.label ?? node.attrs.id} }}`
+                                  `${options.suggestion?.char ?? '{{'}${node.attrs.label ?? node.attrs.id}}}`
                               ]
                           },
                           suggestion: suggestionConfig,


packages/agentflow/src/atoms/SuggestionDropdown.tsx
```diff
@@ -0,0 +1,163 @@
...
+                                py: 0.5,
+                                bgcolor: theme.palette.mode === 'dark' ? theme.palette.common.black : theme.palette.grey[50]
+                            }}
+                        >
+                            <Typography variant='overline' color='text.secondary'>
+                                {category}
+                            </Typography>
+                        </ListItem>
+                        {categoryItems.map((item) => {
+                            const itemIndex = items.findIndex((i) => i.id === item.id)
```
jocelynlin-wd, line 120: can you confirm if `item.id` is always unique?

packages/agentflow/src/features/node-editor/useAvailableVariables.ts
```diff
@@ -14,41 +15,53 @@ const GLOBAL_VARIABLES: VariableItem[] = [
...
+        value: '{{loop_count}}'
+    },
     {
         label: 'file_attachment',
-        description: 'Files uploaded from the chat when Full File Upload is enabled on the Configuration',
+        description: 'Files uploaded from the chat',
         category: 'Chat Context',
         value: '{{file_attachment}}'
+    },
+    { label: '$flow.sessionId', description: 'Current session ID', category: 'Flow Variables', value: '{{$flow.sessionId}}' },
```
jocelynlin-wd, line 42: doesn't look like `Flow Variables` is in type `export type VariableCategory = 'Chat Context' | 'Node Outputs' | 'Flow State'` ?

packages/agentflow/src/atoms/VariablePicker.tsx
```diff
@@ -0,0 +1,126 @@
+import { useMemo } from 'react'
+
+import { Box, Divider, List, ListItem, ListItemAvatar, ListItemButton, ListItemText, Stack, Typography } from '@mui/material'
+import { useTheme } from '@mui/material/styles'
+import { IconBinaryTree, IconHistory, IconMessageChatbot, IconPaperclip } from '@tabler/icons-react'
+
+export interface VariableItem {
+    label: string
+    description?: string
+    category?: string
```
jocelynlin-wd, line 10: this seems to be inconsistent with `types/node.ts` Variable definition which limits the category type?

```
export type VariableCategory = 'Chat Context' | 'Node Outputs' | 'Flow State'

export interface Variable {
    id: string
    label: string
    description?: string
    category: VariableCategory
    /** The path used for resolution, e.g. 'question', 'nodeId.data.instance' */
    path: string
}
```

packages/agentflow/src/core/primitives/customMention.ts
```diff
@@ -0,0 +1,81 @@
...
+ */
+import { type JSONContent, type MarkdownToken, PasteRule } from '@tiptap/core'
+import Mention from '@tiptap/extension-mention'
+
+export const CustomMention = Mention.extend({
+    /**
+     * Render mention nodes as `{{label}}` in plain text output.
+     */
+    renderText({ node }) {
+        return `{{${node.attrs.label ?? node.attrs.id}}}`
```
jocelynlin-wd, line 15: this renders  `{{question}}` without spaces, which is inconsistent with `renderHTML` in `VariableInput.tsx` that renders `{{ question }}` with whitespace in between

packages/agentflow/src/atoms/VariableInput.tsx
```diff
@@ -0,0 +1,207 @@
...
+            ...(placeholder ? [Placeholder.configure({ placeholder })] : []),
+            ...(suggestionConfig
+                ? [
+                      CustomMention.configure({
+                          HTMLAttributes: { class: 'mention' },
+                          renderHTML({ options, node }) {
+                              return [
+                                  'span',
+                                  mergeAttributes(this.HTMLAttributes ?? {}, options.HTMLAttributes),
+                                  `${options.suggestion?.char ?? '{{'} ${node.attrs.label ?? node.attrs.id} }}`
```
jocelynlin-wd: this line produced `{{ question }}` while `renderText` in `customMention.ts` will render `{{question}}` (no whitespace)

packages/agentflow/src/core/utils/index.ts
```diff
@@ -9,3 +9,6 @@ export { evaluateFieldVisibility, evaluateParamVisibility, stripHiddenFieldValue
 
 // Dynamic output anchor utilities
 export { buildDynamicOutputAnchors, parseOutputHandleIndex } from './dynamicOutputAnchors'
+
+// Variable utilities
+export { extractVariables, getUpstreamNodes, VARIABLE_REGEX } from './variableUtils'
```
jocelynlin-wd: do we need to expose `VARIABLE_REGEX` here? what component uses the constant?   also doesn't look like `extractVariables` is used in any other places and probably don't need to be exposed?

packages/agentflow/package.json
```diff
@@ -72,10 +72,14 @@
...
         "@tabler/icons-react": "^3.7.0",
-        "@tiptap/extension-code-block-lowlight": "^3.4.3",
-        "@tiptap/extension-placeholder": "^2.11.5",
-        "@tiptap/react": "^2.11.5",
-        "@tiptap/starter-kit": "^2.11.5",
+        "@tiptap/core": "^3.20.4",
+        "@tiptap/extension-code-block-lowlight": "^3.20.4",
+        "@tiptap/extension-mention": "^3.20.4",
+        "@tiptap/extension-placeholder": "^3.20.4",
+        "@tiptap/markdown": "^3.20.4",
```
jocelynlin-wd: do we need this dependency? there is no source import it seems?

packages/agentflow/src/atoms/VariableInput.tsx
```diff
@@ -0,0 +1,207 @@
...
+                      })
+                  ]
+                : [])
+        ],
+        [placeholder, suggestionConfig]
+    )
+
+    const editor = useEditor({
+        extensions,
+        content: '',
```
jocelynlin-wd: The uncontrolled editor initialization (`content: ''` then syncing via `useEffect`) could cause a flash or spurious `onUpdate` on mount

The editor is created with empty content:
```
  const editor = useEditor({
      extensions,
      content: '',  // ← starts empty
      ...
  })
```
  Then a `useEffect` syncs the actual value in:
```
  useEffect(() => {
      if (editor && value !== editor.getHTML()) {
          editor.commands.setContent(value, { emitUpdate: false })
      }
  }, [editor, value])
```
  The problem is the timing on first render:

  1. Editor mounts with `content: ''`
  2. `onUpdate` fires → calls `onChange('')` → parent state gets overwritten with empty string
  3. `useEffect` runs → sees `value !== editor.getHTML()` → sets the real content back

  Step 2 is the risk. If the parent passes `value="Hello {{question}}"`, the editor briefly tells the parent "the value is now empty" before the effect corrects it. Whether this actually causes a bug depends on how the parent handles that intermediate `onChange('')` call — but it's a race condition by design.

  The fix is straightforward: initialize with the value prop:
```
  content: value || '',
```
  This eliminates the empty→real content round-trip entirely.

packages/agentflow/src/atoms/SuggestionDropdown.tsx
j-sanaa: Made a modification in NodeInputHandler to ensure that ID is always unique even if 2 flow variables have the same label - https://github.com/FlowiseAI/Flowise/pull/6067/changes#diff-b52d4377dba4f5458b10ee1431b02b6cf2ae30aa6806b598080e91a00b5c9617R186-R200

packages/agentflow/src/atoms/VariablePicker.tsx
j-sanaa: The unused type declaration was removed from node.ts and the category is now left to be any string as to avoid adding any domain logic to atoms. Also this allows for more flexibility in case more VariableCategories are added later

packages/agentflow/src/core/primitives/customMention.ts
j-sanaa: Both instances have been modified to have no spaces

packages/agentflow/src/features/node-editor/useAvailableVariables.ts
j-sanaa: Agreed! The hardcoded category names has been deleted

packages/agentflow/src/core/utils/index.ts
j-sanaa: Both have not been used anywhere for now but were a part of v2 implementation. I have removed the exports but kept the functions

packages/agentflow/package.json
j-sanaa: The import can be added later when markdown support is implemented. Removed for now

packages/agentflow/src/atoms/VariableInput.tsx
j-sanaa: Applied suggested fix!

packages/agentflow/src/atoms/VariablePicker.tsx
jocelynlin-wd: sounds good

packages/agentflow/src/atoms/VariableInput.tsx
```diff
@@ -0,0 +1,207 @@
...
+            ...(placeholder ? [Placeholder.configure({ placeholder })] : []),
+            ...(suggestionConfig
+                ? [
+                      CustomMention.configure({
+                          HTMLAttributes: { class: 'mention' },
+                          renderHTML({ options, node }) {
+                              return [
+                                  'span',
+                                  mergeAttributes(this.HTMLAttributes ?? {}, options.HTMLAttributes),
+                                  `${options.suggestion?.char ?? '{{'}${node.attrs.label ?? node.attrs.id} }}`
```
jocelynlin-wd: nit: this will render `{{question }}` (note the extra whitespace on closing bracket) vs `{{question}}`

PR #6117, opened 2026-04-02
feat: Add OpenTelemetry Analytics Provider

rohan-patil2:
## Summary

This PR adds **OpenTelemetry (OTEL) analytics provider** to Flowise, enabling users to export distributed traces from chatflow and agentflow executions to any OTLP-compatible observability backend (New Relic, Datadog, Grafana Cloud, or any custom endpoint).

The integration follows OpenInference semantic conventions for LLM observability and provides full lifecycle tracing including LLM calls, tool invocations, retriever operations, chain executions, and agent actions; with token usage, latency metrics, and rich span attributes.

## Motivation

Flowise already supports analytics providers like LangSmith, Langfuse, LunaryAI, and Opik, but lacked support for the vendor-neutral OpenTelemetry standard. Many enterprise teams already run OTLP-compatible backends (New Relic, Datadog, Grafana Tempo, etc.) and need a way to funnel LLM observability data into their existing infrastructure without adopting yet another proprietary tool.

This enhancement fills that gap by adding OpenTelemetry to existing providers across the full stack: credential management, UI configuration, backend handler lifecycle, and per-chatflow tracer pooling.

**Provider Setup**:
otel-setup

**Credential Setup**:
otel-creds-setup


harshit-flowise:
@rohan-patil2 Thanks for the PR. I am still reviewing it but while testing I found that the test cases are present in a separate folder. Please check this https://github.com/FlowiseAI/Flowise/blob/main/CONTRIBUTING.md file. 
According to the standards all tests should co-locate with source file.

harshit-flowise:
Check these issues as well, please take a look - 

  1. Config hash collision in computeConfigHash silently reuses wrong tracer provider — JSON.stringify(config, Object.keys(config).sort()) passes a top-level key array as the replacer, which strips all nested values (e.g., headers becomes {}). Two configs with different authentication headers produce the same hash, causing the pool to reuse the first provider and send telemetry with incorrect credentials.

  https://github.com/FlowiseAI/Flowise/blob/687005cb5aa7ad164f7b3f9fdbde77000aff2d96/packages/components/src/analyticsHandlers/otel/OtelTracerProviderPool.ts#L13-L17

  2. Three of four vendor presets contain placeholder URLs that fail Zod .url() validation at runtime — datadog, logicmonitor, and grafana presets use angle-bracket placeholders (e.g., http://<datadog-agent-host>:4318/v1/traces). OtelConfigSchema enforces .url() validation, so selecting any of these presets without editing the endpoint will throw a ZodError when the config is parsed.

  https://github.com/FlowiseAI/Flowise/blob/687005cb5aa7ad164f7b3f9fdbde77000aff2d96/packages/components/credentials/OpenTelemetryApi.credential.ts#L9-L28

  3. _endOtelSpan double-encodes outputText and sets incorrect MIME type — outputText is already a plain string, but line 801 wraps it in JSON.stringify(), producing double-quoted values in spans. Line 802 also sets output.mime_type to 'application/json' for a plain-string value; should be 'text/plain'. Affects all four providers calling _endOtelSpan.

  https://github.com/FlowiseAI/Flowise/blob/687005cb5aa7ad164f7b3f9fdbde77000aff2d96/packages/components/src/handler.ts#L799-L803

  4. AddEditCredentialDialog accordion change is global, not scoped to OpenTelemetry — the filter at line 494 now hides additionalParams: true fields from the main view for all credentials. Existing credentials using additionalParams: true (e.g., GoogleAuth — Project ID field) will have those fields silently moved into a collapsed accordion.

  https://github.com/FlowiseAI/Flowise/blob/687005cb5aa7ad164f7b3f9fdbde77000aff2d96/packages/ui/src/views/credentials/AddEditCredentialDialog.jsx#L493-L525

packages/components/package.json
```diff
@@ -127,6 +129,7 @@
         "form-data": "^4.0.4",
         "google-auth-library": "^9.4.0",
         "graphql": "^16.6.0",
+        "groq-sdk": "^0.5.0",
```
jchui-wd: Seems like this was added but not imported anywhere, is it still needed?

packages/components/src/handler.ts
```diff
@@ -679,6 +716,18 @@ export const additionalCallbacks = async (nodeData: INodeData, options: ICommonO
 
                     const tracer: Tracer | undefined = getOpikTracer(opikOptions)
                     callbacks.push(tracer)
+                } else if (provider === 'openTelemetry') {
+                    const handler = await getOtelCallbackHandler(
```
jchui-wd: Should perhaps wrap this in a try catch incase this call fails and breaks when credentials are malformed.

packages/components/src/handler.ts
```diff
@@ -765,6 +814,28 @@ export class AnalyticHandler {
             if (modelName) {
                 llmSpan.setAttribute('llm.model_name', modelName)
             }
+
+            if (providerName === 'openTelemetry') {
+                if (usageMetadata) {
+                    if (usageMetadata.input_tokens !== undefined) {
+                        llmSpan.setAttribute('llm.prompt_tokens', usageMetadata.input_tokens)
```
jchui-wd, line 826: just want to double check on this, `llm.prompt_tokens` might be a legacy and the new standard might something like `gen_ai.usage.input_tokens`

https://opentelemetry.io/docs/specs/semconv/gen-ai/gen-ai-spans/

packages/components/credentials/OpenTelemetryApi.credential.ts
```diff
@@ -0,0 +1,197 @@
...
+                            otelHeaderValue: VENDOR_PRESETS.logicmonitor.headerValueHint
+                        },
+                        grafana: {
+                            otelEndpoint: VENDOR_PRESETS.grafana.endpoint,
+                            otelProtocol: VENDOR_PRESETS.grafana.protocol,
+                            otelHeaderKey: VENDOR_PRESETS.grafana.headerKey,
+                            otelHeaderValue: VENDOR_PRESETS.grafana.headerValueHint
+                        }
+                    }
+                } as any)
```
jchui-wd: if autoPopulate is going to be a new functionality, perhaps we can just add it to INodeParams Interface instead of a one of 'any'. the refresh value in the interface works similarily

jchui-wd (commented):
Hey left a few more comments!

packages/components/src/handler.ts
rohan-patil2: Yes, The GenAI specification has not been added, but I can include it.

packages/components/package.json
rohan-patil2: Its not needed, will remove it in the fix

yau-wd:
Hi @rohan-patil2 @jchui-wd, 

Could you please let me know if there are any more changes planned? If not, would you mind resolving the conflicts as well?

Once that’s done, I’ll start testing.

rohan-patil2:
> Hi @rohan-patil2 @jchui-wd,
> 
> Could you please let me know if there are any more changes planned? If not, would you mind resolving the conflicts as well?
> 
> Once that’s done, I’ll start testing.

@yau-wd The changes and fixes are currently in progress and undergoing active testing. Once testing is complete, the PR will be updated. You can expect an update within the next one to two days.

packages/components/src/handler.ts
rohan-patil2: Based on the Link, looks like the GenAI semantic conventions are still in development. I will raise a separate PR for this one.

PR #6139, opened 2026-04-03
Add optional JSON-array output for the Agent Flow Iteration node

Dexterity104:
## Summary

Adds an **Output Structure** option on the Agent Flow Iteration node so results can stay in a fixed order and length that matches the input array, instead of only a single joined text string.

It's useful for flows that treat the iteration as a map over sections or rows and need a stable list to feed into another Iteration or any step that expects parseable JSON.

Closes #6091 

## Behavior

- **Aggregated Text** (default): unchanged.  `output.content` is all iteration outputs joined with newlines.
- **JSON Array**: `output.content` is a JSON string of a string array (one entry per input index). If a sub-run has no text, that index is still present as an empty string so positions stay aligned with the input.

`output.iterationResults` remains the underlying `string[]` in both modes. `content` stays a string everywhere so the usual flow completion and variable paths keep working.


Dexterity104:
Hi @jocelynlin-wd 
Can you please review this PR when you have some time?

jocelynlin-wd (approved):
Changes look good. This is purely additive and opt-in — existing flows are unaffected since the default and fallback both preserve current behavior. Seems to be safe to merge.

Dexterity104:
Thanks, @jocelynlin-wd !

@0xi4o @HenryHengZJ @yau-wd  can you please review this PR?

packages/components/nodes/agentflow/Iteration/Iteration.ts
```diff
@@ -32,6 +32,25 @@ class Iteration_Agentflow implements INode {
                 description: 'The input array to iterate over',
                 acceptVariable: true,
                 rows: 4
+            },
+            {
```
HenryHengZJ, line 36: for any changes we make to the parameters, we should also update the `version` of the node as well

packages/server/src/utils/buildAgentflow.ts
```diff
@@ -1290,9 +1291,12 @@ const executeNode = async ({
                             productId
                         })
 
-                        // Store the result
+                        // Store the result.
+                        // In jsonArray mode, always push to maintain 1:1 mapping with the input array.
                         if (subFlowResult?.text) {
                             iterationResults.push(subFlowResult.text)
+                        } else if (outputStructure === 'jsonArray') {
```
HenryHengZJ, line 1298: i dont understand this bit, even if we have outputstructure as jsonarray, we will still always have results from subflow, so it never goes into here, that makes me wonder whats the point?

HenryHengZJ:
im not 100% sure if this is actually solving the original issue

packages/components/nodes/agentflow/Iteration/Iteration.ts
Dexterity104: That makes sense. I updated the node `version` to 1.1.

packages/server/src/utils/buildAgentflow.ts
Dexterity104: `text` can still be empty when the subflow ran but the last node's `content` is empty, so the first `if` skips. The `else` keeps one slot per input so lengths stay matched. I can tighten the comment or refactor if it's noisy.

Dexterity104:
> im not 100% sure if this is actually solving the original issue

Before this change, if a sub run finished but `text` was empty, nothing was pushed, so `iterationResults` could be shorter than the input. In JSON Array mode this PR keeps one entry per input (empty string when needed). If every iteration had non empty `text`, behavior matched the input length already.

Dexterity104:
@0xi4o @HenryHengZJ @yau-wd can you please review this PR?

HenryHengZJ:
if you look at the comments I have on the issue https://github.com/FlowiseAI/Flowise/issues/6091, user can already achieve that by using `{{iterationAgentflow_0.output.iterationResults}}`, hence I dont think there is a need for this

PR #6158, opened 2026-04-06
fix: flowise-186

0xi4o:


HenryHengZJ:
once we turn these into password type, from the UI do we have the toggle button to see the actual value? I think somebody created a PR for the signin page right?

packages/server/src/routes/credentials/index.ts
```diff
@@ -8,7 +8,10 @@ router.post('/', checkPermission('credentials:create'), credentialsController.cr
 
 // READ
 router.get('/', checkPermission('credentials:view'), credentialsController.getAllCredentials)
-router.get(['/', '/:id'], checkPermission('credentials:view'), credentialsController.getCredentialById)
+router.get(['/', '/:id'], checkAnyPermission('credentials:create,credentials:update'), credentialsController.getCredentialById)
```
HenryHengZJ, line 11: why are we getting rid of `credentials:view` permission ?

packages/server/src/services/credentials/index.ts
```diff
@@ -177,7 +177,13 @@ const updateCredential = async (credentialId: string, requestBody: any, workspac
             throw new InternalFlowiseError(StatusCodes.NOT_FOUND, `Credential ${credentialId} not found`)
         }
         const decryptedCredentialData = await decryptCredentialData(credential.encryptedData)
-        requestBody.plainDataObj = { ...decryptedCredentialData, ...requestBody.plainDataObj }
+        const incomingData = requestBody.plainDataObj ?? {}
+        const sanitizedData = Object.fromEntries(
+            Object.entries(incomingData).filter(
+                ([, v]) => typeof v !== 'string' || (!v.includes('\u2022\u2022\u2022\u2022\u2022\u2022') && v !== REDACTED_CREDENTIAL_VALUE)
```
HenryHengZJ: whats this means?

packages/ui/src/ui-component/input/Input.jsx
```diff
@@ -1,26 +1,30 @@
 import { useState, useEffect, useRef } from 'react'
 import PropTypes from 'prop-types'
-import { FormControl, OutlinedInput, InputBase, Popover, InputAdornment, IconButton } from '@mui/material'
+import { FormControl, OutlinedInput, InputBase, Popover, InputAdornment, IconButton, CircularProgress } from '@mui/material'
```
HenryHengZJ: can we createa a new component for this URL type?  if we can make it cleaner. reason being I want to reduce the changes in `Input.jsx` because this would then require update to `@agentflow` package

packages/server/src/routes/credentials/index.ts
0xi4o: The `getCredentialById` returns decrypted credentials in the response. In the UI, users with `credentials:view` cannot view any of the fields for a credential. With the API, password-type fields are redacted but other fields are returned as plaintext. This change will prevent `credentials:view` users from looking at any of the fields in a credential.

packages/server/src/services/credentials/index.ts
0xi4o: The UI already doesn't let users edit and save when redaction is active, so this is not needed anymore. Removed it.

packages/ui/src/ui-component/input/Input.jsx
0xi4o: Added a new `SensitiveInput` component and reverted changes to this component.

igor-magun-wd (approved):

HenryHengZJ (approved):

PR #6174, opened 2026-04-08
fix: ChatOpenAI baseOptions additional parameters not applied to model

### LINKED ISSUE #2499: [BUG] Additional Parameters did not take effect.
**Describe the bug**

In ChatOpenAI Custom, I want to customize the stop word. I placed this parameter in BaseOptions, but when making a request, the parameter was not passed.

examples:
stop: ["<|im_end|>"]

BOOPATB:
## Problem
Fixes #2499

`BaseOptions` JSON input in the ChatOpenAI node was only being passed as
`defaultHeaders` inside `configuration`, meaning any model-level parameters
(e.g. `stop`, `top_p`) were silently ignored and never sent to the API.

## Root Cause
In `ChatOpenAI.ts`, `parsedBaseOptions` was unconditionally assigned to
`configuration.defaultHeaders`, with no path to merge params into the model object.
Additionally, the merge happened *after* the reasoning model cleanup block, meaning
it could have re-introduced `stop`/`temperature` that were intentionally deleted.

## Fix
- Destructure `parsedBaseOptions` into `{ defaultHeaders, baseURL, ...modelParams }`
- Spread `modelParams` directly into `obj` so params like `stop` reach the model
- Move the `baseOptions` merge to *before* the reasoning model block so cleanup still works correctly
- Update the `BaseOptions` field description to reflect its actual capability

## Testing
- Set `{ "stop": ["<|im_end|>"] }` in BaseOptions on a ChatOpenAI node
- Confirm the stop sequence is respected in the model response
- Confirm reasoning models (`o-series`) still correctly strip `stop`/`temperature`
- Confirm `defaultHeaders` and `baseURL` inside BaseOptions still work for custom endpoints

### DIFF
--- a/packages/components/nodes/chatmodels/ChatOpenAI/ChatOpenAI.ts
+++ b/packages/components/nodes/chatmodels/ChatOpenAI/ChatOpenAI.ts
@@ -205,7 +205,7 @@ class ChatOpenAI_ChatModels implements INode {
                 name: 'baseOptions',
                 type: 'json',
                 optional: true,
-                description: 'Default headers to include with every request to the API.',
+                description: 'Additional model parameters (e.g. { "stop": ["<|im_end|>"] }) or default headers ({ "defaultHeaders": { "X-Custom": "value" } }).',
                 additionalParams: true
             }
         ]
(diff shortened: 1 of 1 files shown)


BOOPATB:
Fixes #2499

PR #6182, opened 2026-04-10
fix(evaluations): correct DoesNotContainAny/All operator logic

aviu16:
## What happened

The `DoesNotContainAny` text evaluator returns inverted results — it reports **Pass** when one of the specified values *is* found in the output, and **Fail** when none are found. This silently corrupts evaluation metrics for anyone using that operator.

`DoesNotContainAll` has a separate issue: it uses the same check-loop as `DoesNotContainAny` (break on first match) instead of verifying that *every* value is present before failing. So it behaves identically to `DoesNotContainAny` rather than checking "all present."

## Fix

- **`DoesNotContainAny`**: flip the ternary so `passed` (no values found) maps to `'Pass'`.
- **`DoesNotContainAll`**: invert the inner `includes` check so `passed` means "all values were found," then the existing ternary correctly returns `'Fail'`.

Both operators now match the semantics their names describe.

### DIFF
--- a/packages/server/src/services/evaluations/EvaluatorRunner.ts
+++ b/packages/server/src/services/evaluations/EvaluatorRunner.ts
@@ -144,21 +144,21 @@ export const runAdditionalEvaluators = async (
                             }
                             subArray.push({
                                 ...returnFields,
-                                result: passed ? 'Fail' : 'Pass'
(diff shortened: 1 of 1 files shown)


HenryHengZJ (approved):
thank you!

PR #6183, opened 2026-04-10
fix(evaluations): use correct total_cost property in CostCalculator

aviu16:
## What happened

`CostCalculator.ts` line 28 checks `costValues.total_price`, but every other part of the codebase (models.json, evaluation runners) uses `total_cost`. Since `total_price` is always `undefined`, the condition is never true and models that define a flat `total_cost` rate fall through to the `input_cost`/`output_cost` path — which are also undefined for those models, producing `NaN` costs.

## Fix

Change `total_price` → `total_cost` on line 28 to match the actual property name.


HenryHengZJ (approved):

HenryHengZJ:
thank you!

PR #6187, opened 2026-04-10
Bugfix/add compact table transformer for NotionAPILoader

### LINKED ISSUE #6180: Issue with MD table parsing in Document Store from Notion Database
### Describe the bug

When loading from a Notion Database in the doc store, pages containing tables are being parsed oddly.

Looking at the processed chunk, the width of the table columns (set in the second row of the MD table) are becoming long, apparently matching the length of the longest string in the rows below.

Space characters are also being added to each MD cell, apparently matching the width of the column set in the 2nd row.

This adds unnecessary characters, increasing tokens count, and makes splitters unable to function properly. 

### To Reproduce

1. Create a Notion Database with a page that contains a table. I've attached an example table - note that when pasted into Notion the table should be rendered.
[example table.md](https://github.com/user-attachments/files/26618180/example.table.md)

**Raw MD**
Image

**Rendered in Notion DB Page**
Image

2. Create a document store in Flowise with a Notion Database loader that loads the database created in step 1.

3. View the chunks - you'll find that the 2nd row of the md table contains many hyphens, and the rows that follow contain many space characters.

Image

### Expected behavior

Space characters should not be added to the MD cells.
And ideally the width of the columns should not be automatically updated based on the length of content in the rows.

### Screenshots

_No response_

### Flow

_No response_

### Use Method

Docker

### Flowise Version

3.1.1

### Operating System

None

### Browser

None

### Additional context

_No response_

HenryHengZJ:
## Root Cause
`notion-to-md's md.table()` function calls `markdown-table(cells)` with no options, inheriting the library's padding defaults. The `n2mClient1 in LangChain's NotionAPILoader is private, so there's no way to configure this externally — but setCustomTransformer allows overriding specific block type handlers.

## Fix
Uses `setCustomTransformer('table', ...)` on the NotionAPILoader's internal `n2mClient` to bypass the default table handler. The custom transformer replicates the same cell-fetching logic but generates compact markdown tables without padding.

Previously:
image

After:
image

### DIFF
--- a/packages/components/nodes/documentloaders/Notion/NotionDB.ts
+++ b/packages/components/nodes/documentloaders/Notion/NotionDB.ts
@@ -3,6 +3,7 @@ import { ICommonObject, IDocument, INode, INodeData, INodeParams } from '../../.
 import { TextSplitter } from '@langchain/textsplitters'
 import { NotionAPILoader, NotionAPILoaderOptions } from '@langchain/community/document_loaders/web/notionapi'
 import { getCredentialData, getCredentialParam, handleEscapeCharacters, INodeOutputsValue } from '../../../src'
+import { applyCompactTableTransformer } from './notionTableFix'
 
 class NotionDB_DocumentLoaders implements INode {
     label: string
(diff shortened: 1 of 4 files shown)


henry-heng-wd (approved):

saatchi-david:
Thank you!!

PR #6191, opened 2026-04-10
bugfix: message panel crashes via large chat volume

### LINKED ISSUE #6038: Chat Messages panel crashes with OOM when database has large volume of messages — no pagination or default limit
### Describe the bug

When accessing the Chat Messages panel in the Flowise dashboard, the application crashes with "FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory" if there are a large number of messages within the default 30-day date range.

The panel loads the full 30-day range before the user has any chance to change the date filter. With ~157,000 messages in the database, Node.js attempts to serialize all messages into JSON in memory, consuming 8-12GB of RAM before crashing.

This happens regardless of the database backend — we've tested with both SQLite and PostgreSQL 17. The atabase returns the data quickly; the issue is Node.js trying to hold all messages in memory at once.

Environment:
- Flowise: latest (self-hosted via Docker/Easypanel)
- Database: PostgreSQL 17 (also reproducible with SQLite)
- Messages in database: ~157,000

Error log:
<--- Last few GCs --->
[1:0x7fcef78f2660] Mark-Compact 8036.4 (8238.9) -> 8036.4 (8238.9) MB, 
626.75 / 0.00 ms allocation failure; scavenge might not succeed

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory

### To Reproduce

1. Accumulate ~100,000+ chat messages over time
2. Navigate to the Chat Messages panel in the dashboard
3. The panel automatically loads messages from the last 30 days
4. Node.js crashes with heap out of memory before the user can change the date filter

### Expected behavior

The Chat Messages panel should either:

1. Use server-side pagination with a configurable page size (e.g., 100-500 messages per page) instead of loading all messages at once
2. Allow configuration of the default date range via environment variable (e.g., CHATMESSAGE_DEFAULT_DAYS=7)
3. Default to a shorter date range (e.g., 7 days) for the initial load, letting the user expand if needed
4. Stream/lazy-load messages instead of loading everything into memory at once

Any of these would prevent the OOM crash for users with high message volumes.

### Screenshots

_No response_

### Flow

_No response_

### Use Method

Docker

### Flowise Version

3.1.0

### Operating System

Linux

### Browser

Chrome

### Additional context

_No response_

jchui-wd:
**Issue:**
`statsService.getChatflowStats` fetched every message row into the Node.js heap to just to calculate the `.length` of those `chat_message` rows.
When theres too many messages (like 100,000), it will overload the memory just to get the count of these messages.

**Change / Fix:**
Instead of getting ALL messages and their rows, we now just do four SQL queries (+ 1 more preceding messages if theres feedback types filter) that just gets the count / length number which is what `getChatflowStats` returns anyways.
Fixes: https://github.com/FlowiseAI/Flowise/issues/6038

**Note:**
- When dealing with this, I realize large chat message volumes is also crashing the Frontend when opening chat dialog, will open an issue / ticket and fix that seperately.

**Refactoring & Cleanup**
- Dead Code: Removed unused messageId and feedback parameters from the service signature.
- Error Handling: Migrated plain errors to InternalFlowiseError with appropriate HTTP codes (401/403).
- Testing: Added unit tests for authorization, filter operators (In, Between, etc.), and feedback subquery logic.

<details>
  <summary>Test Videos Demo</summary>

  <b>BEFORE-101k...-memory.mp4</b>
  <video src="https://github.com/user-attachments/assets/476c3b9f-9090-403d-b1f7-fe6171ea2c5e" controls width="100%"></video>

  <b>AFTER-101k-...-memory.mp4</b>
  <video src="https://github.com/user-attachments/assets/43e498b4-4c92-4d8b-b659-42c8f501cfaa" controls width="100%"></video>
</details>

<details>
  <summary>This is to show the functionality remains the same after using SQL queries, doesn't show the bug, OLD is get all chat messages method, NEW is the 4 + 1 sql query method. Both show the same thing</summary>
  <b>101k-OLD.mp4</b>
  <video src="https://github.com/user-attachments/assets/e7d6056a-d76c-4ee5-989c-06130ccbcec5" controls width="100%"></video>

  <b>101k-NEW.mp4</b>
  <video src="https://github.com/user-attachments/assets/2b82f25b-247d-4176-a8a9-4e39e68683a2" controls width="100%"></video>
</details>

Screenshot 2026-04-13 at 1 15 54 PM

### DIFF
--- a/packages/server/src/services/stats/index.ts
+++ b/packages/server/src/services/stats/index.ts
@@ -45,34 +45,15 @@ const getChatflowStats = async (
         else if (startDate) baseWhere.createdDate = MoreThanOrEqual(new Date(startDate))
         else if (endDate) baseWhere.createdDate = LessThanOrEqual(new Date(endDate))
 
-        // (sessions that contain at least one message with that feedback rating)
-        if (feedbackTypes && feedbackTypes.length > 0) {
-            const rows = await repo
-                .createQueryBuilder('cm2')
-                .select('DISTINCT(cm2.sessionId)', 'sessionId')
-                .innerJoin(ChatMessageFeedback, 'f2', 'f2.messageId = cm2.id')
-                .where('cm2.chatflowid = :chatflowid', { chatflowid })
-                .andWhere('f2.rating IN (:...feedbackTypes)', { feedbackTypes })
-                .getRawMany()
-
-            const qualifyingSessionIds = rows.map((r) => r.sessionId)
-
-            // No matching sessions - return zeros immediately
-            if (qualifyingSessionIds.length === 0) {
-                return { totalMessages: 0, totalSessions: 0, totalFeedback: 0, positiveFeedback: 0 }
-            }
-
-            baseWhere.sessionId = In(qualifyingSessionIds)
-        }
-
+        // TypeORM 0.3.x QueryBuilder accepts FindOptionsWhere objects (including FindOperators
+        // like In/Between) in .where() — the same WHERE clause machinery used by repo.find().
+        const totalMessagesQb = repo.createQueryBuilder('cm').select('COUNT(*)', 'count').where(baseWhere)
         const totalSessionsQb = repo.createQueryBuilder('cm').select('COUNT(DISTINCT(cm.sessionId))', 'count').where(baseWhere)
-
         const totalFeedbackQb = repo
             .createQueryBuilder('cm')
             .select('COUNT(*)', 'count')
             .innerJoin(ChatMessageFeedback, 'f', 'f.messageId = cm.id')
             .where(baseWhere)
-
         const positiveFeedbackQb = repo
             .createQueryBuilder('cm')
             .select('COUNT(*)', 'count')
@@ -80,15 +61,34 @@ const getChatflowStats = async (
             .where(baseWhere)
             .andWhere('f.rating = :rating', { rating: ChatMessageRatingType.THUMBS_UP })
 
-        const [totalMessages, totalSessionsRaw, totalFeedbackRaw, positiveFeedbackRaw] = await Promise.all([
-            repo.count({ where: baseWhere }),
+        if (feedbackTypes && feedbackTypes.length > 0) {
+            const sessionSubQb = repo
+                .createQueryBuilder()
+                .subQuery()
+                .select('DISTINCT cm2.sessionId')
+                .from(ChatMessage, 'cm2')
+                .innerJoin(ChatMessageFeedback, 'f2', 'f2.messageId = cm2.id')
+                .where(baseWhere)
+                .andWhere('f2.rating IN (:...feedbackTypes)', { feedbackTypes })
+
+            const subSql = sessionSubQb.getQuery()
+            const subParams = sessionSubQb.getParameters()
+
+            for (const qb of [totalMessagesQb, totalSessionsQb, totalFeedbackQb, positiveFeedbackQb]) {
+                qb.andWhere(`cm.sessionId IN ${subSql}`)
+                qb.setParameters(subParams)
+            }
+        }
+
+        const [totalMessagesRaw, totalSessionsRaw, totalFeedbackRaw, positiveFeedbackRaw] = await Promise.all([
+            totalMessagesQb.getRawOne(),
             totalSessionsQb.getRawOne(),
             totalFeedbackQb.getRawOne(),
             positiveFeedbackQb.getRawOne()
         ])
 
         return {
-            totalMessages,
+            totalMessages: parseInt(totalMessagesRaw?.count ?? '0', 10),
             totalSessions: parseInt(totalSessionsRaw?.count ?? '0', 10),
             totalFeedback: parseInt(totalFeedbackRaw?.count ?? '0', 10),
             positiveFeedback: parseInt(positiveFeedbackRaw?.count ?? '0', 10)
--- a/packages/server/src/services/stats/index.ts
+++ b/packages/server/src/services/stats/index.ts
@@ -24,7 +24,6 @@ const getChatflowStats = async (
                 `Error: statsService.getChatflowStats - activeWorkspaceId not provided!`
             )
         }
-
         const appServer = getRunningExpressApp()
 
         const chatflow = await appServer.AppDataSource.getRepository(ChatFlow).findOneBy({
@@ -39,6 +38,7 @@ const getChatflowStats = async (
 
         const repo = appServer.AppDataSource.getRepository(ChatMessage)
 
+        // Build up queries that will be used for all stats
         const baseWhere: FindOptionsWhere<ChatMessage> = { chatflowid }
         if (chatTypes && chatTypes.length > 0) baseWhere.chatType = In(chatTypes)
         if (startDate && endDate) baseWhere.createdDate = Between(new Date(startDate), new Date(endDate))
@@ -51,15 +51,20 @@ const getChatflowStats = async (
         const totalSessionsQb = repo.createQueryBuilder('cm').select('COUNT(DISTINCT(cm.sessionId))', 'count').where(baseWhere)
         const totalFeedbackQb = repo
             .createQueryBuilder('cm')
-            .select('COUNT(*)', 'count')
+            .select('COUNT(DISTINCT cm.id)', 'count')
             .innerJoin(ChatMessageFeedback, 'f', 'f.messageId = cm.id')
             .where(baseWhere)
         const positiveFeedbackQb = repo
             .createQueryBuilder('cm')
-            .select('COUNT(*)', 'count')
+            .select('COUNT(DISTINCT cm.id)', 'count')
             .innerJoin(ChatMessageFeedback, 'f', 'f.messageId = cm.id')
             .where(baseWhere)
             .andWhere('f.rating = :rating', { rating: ChatMessageRatingType.THUMBS_UP })
+
+        // When feedback filter is active, narrow all queries to sessions that contain
+        // matching feedback, restrict feedback counts to selected types, and build a
+        // precedingCount query (totalMessages = feedback msgs + their preceding user msgs).
+        let precedingCountQb: ReturnType<typeof repo.createQueryBuilder> | null = null
 
         if (feedbackTypes && feedbackTypes.length > 0) {
             const sessionSubQb = repo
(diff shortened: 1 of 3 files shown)


packages/server/src/services/stats/index.ts
```diff
@@ -1,47 +1,98 @@
...
+            .select('COUNT(*)', 'count')
+            .innerJoin(ChatMessageFeedback, 'f', 'f.messageId = cm.id')
+            .where(baseWhere)
+
+        const positiveFeedbackQb = repo
+            .createQueryBuilder('cm')
+            .select('COUNT(*)', 'count')
+            .innerJoin(ChatMessageFeedback, 'f', 'f.messageId = cm.id')
+            .where(baseWhere)
+            .andWhere('f.rating = :rating', { rating: ChatMessageRatingType.THUMBS_UP })
```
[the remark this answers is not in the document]
jchui-wd: Verified in TypeORM 0.3.20 source, QueryBuilder.where(object) explicitly handles FindOperator instances via getWherePredicateCondition() in QueryBuilder.js. In, Between, MoreThanOrEqual, and LessThanOrEqual all translate correctly to SQL through the same internal machinery as repo.find(). Added a comment to the code to make this explicit.

So it is fine

packages/server/src/services/stats/index.ts
```diff
@@ -1,47 +1,98 @@
...
+
+            const qualifyingSessionIds = rows.map((r) => r.sessionId)
+
+            // No matching sessions - return zeros immediately
+            if (qualifyingSessionIds.length === 0) {
+                return { totalMessages: 0, totalSessions: 0, totalFeedback: 0, positiveFeedback: 0 }
+            }
+
+            baseWhere.sessionId = In(qualifyingSessionIds)
         }
```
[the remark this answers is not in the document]
jchui-wd: Fixed by reusing `baseWhere` in subquery so `chatType` and date filters are consistently applied.

Changed feedbackTypes filter to use subquery to avoid hitting PSQL's parameter limit at high feedback volumes.

HenryHengZJ:
this fix turns out to be more complicated than I think, can you help summarizing what each of those 4 steps looks like

packages/server/src/services/stats/index.ts
```diff
@@ -1,47 +1,135 @@
...
+        if (!chatflow)
+            throw new InternalFlowiseError(
+                StatusCodes.FORBIDDEN,
+                `Error: statsService.getChatflowStats - chatflow ${chatflowid} not found in workspace!`
+            )
+
+        const repo = appServer.AppDataSource.getRepository(ChatMessage)
+
+        // Build up queries that will be used for all stats
+        const baseWhere: FindOptionsWhere<ChatMessage> = { chatflowid }
```
jchui-wd, line 42: `baseWhere` sets up filters that will be applied for all 4 stats `Total Session` `Total Messages` `Total Feedback Received` and `Positive Feedback`
- Specifically ChatTypes and Start and End Dates

packages/server/src/services/stats/index.ts
```diff
@@ -1,47 +1,135 @@
...
+        if (chatTypes && chatTypes.length > 0) baseWhere.chatType = In(chatTypes)
+        if (startDate && endDate) baseWhere.createdDate = Between(new Date(startDate), new Date(endDate))
+        else if (startDate) baseWhere.createdDate = MoreThanOrEqual(new Date(startDate))
+        else if (endDate) baseWhere.createdDate = LessThanOrEqual(new Date(endDate))
+
+        // TypeORM 0.3.x QueryBuilder accepts FindOptionsWhere objects (including FindOperators
+        // like In/Between) in .where() — the same WHERE clause machinery used by repo.find().
+        const totalMessagesQb = repo.createQueryBuilder('cm').select('COUNT(*)', 'count').where(baseWhere)
```
jchui-wd: `totalMessagesQb` is for `Total Messages` and just gets the **count** of all messages in `chat_message` of `chatflowid`

packages/server/src/services/stats/index.ts
```diff
@@ -1,47 +1,135 @@
...
+        const totalSessionsQb = repo.createQueryBuilder('cm').select('COUNT(DISTINCT(cm.sessionId))', 'count').where(baseWhere)
```
jchui-wd: `totalSessionsQb` is for `Total Session` stat
Just gets **count** the total **distinct** sessions from all the `chat_messages `of `chatflowid`

packages/server/src/services/stats/index.ts
```diff
@@ -1,47 +1,135 @@
...
+        const totalFeedbackQb = repo
```
jchui-wd: `totalFeedbackQb` is for `Total Feedback Received` stat.
Gets the **count** for all chat messages that is associated with `ChatMessageFeedback`

packages/server/src/services/stats/index.ts
```diff
@@ -1,47 +1,135 @@
...
+            .createQueryBuilder('cm')
+            .select('COUNT(DISTINCT cm.id)', 'count')
+            .innerJoin(ChatMessageFeedback, 'f', 'f.messageId = cm.id')
+            .where(baseWhere)
+        const positiveFeedbackQb = repo
```
jchui-wd: `positiveFeedbackQb` is for `Positive Feedback` stat.
Gets the **count** for all chat messages that is associated with `ChatMessageFeedback`
AND the rating is `THUMB UP`

packages/server/src/services/stats/index.ts
```diff
@@ -1,47 +1,135 @@
...
+            .where(baseWhere)
+            .andWhere('f.rating = :rating', { rating: ChatMessageRatingType.THUMBS_UP })
+
+        // When feedback filter is active, narrow all queries to sessions that contain
+        // matching feedback, restrict feedback counts to selected types, and build a
+        // precedingCount query (totalMessages = feedback msgs + their preceding user msgs).
+        let precedingCountQb: ReturnType<typeof repo.createQueryBuilder> | null = null
+
+        if (feedbackTypes && feedbackTypes.length > 0) {
+            const sessionSubQb = repo
```
jchui-wd, line 50: I think this is where things get complicated because I am trying to mimic the old functionality when the `feedback` filter is selected.

Old `Total Message` functionality:
1. get **every single message** for chatflow (sorted by date), WITH feedback data associated onto each row
2. Then loop through each message for AI responses that has feedback associateed with the filter feedbackType. (`totalFeedbackQb` gets this part already`
3. also keep a message row that is immediately before it. (so filtered message and + 1 more row before it) (`precedingCountQb` below will be responsible for this section`

so here,`sessionSubQb` gets distinct sessions which matches the selected filtered `feedbackTypes`.
And `totalSessionsQb, totalFeedbackQb, positiveFeedbackQb` will search within those sessions only when `feedbackTypes` search filter is selected.

`Total Message` will be calculated differently with `feedbackTypes`. To match the previous `Total Message` functionality, we can just add `totalFeedback` count + `precedingCount` which will be done down below

packages/server/src/services/stats/index.ts
```diff
@@ -1,47 +1,135 @@
...
+
+            const subSql = sessionSubQb.getQuery()
+            const subParams = sessionSubQb.getParameters()
+
+            for (const qb of [totalMessagesQb, totalSessionsQb, totalFeedbackQb, positiveFeedbackQb]) {
+                qb.andWhere(`cm.sessionId IN ${subSql}`)
+                qb.setParameters(subParams)
+            }
+
+            totalFeedbackQb.andWhere('f.rating IN (:...ratingFilter)', { ratingFilter: feedbackTypes })
```
jchui-wd: this is to make sure `totalFeedbackQb` and `positiveFeedbackQb` ONLY keeps rows that contain the selected `feedbackTypes`

packages/server/src/services/stats/index.ts
```diff
@@ -1,47 +1,135 @@
...
+            positiveFeedbackQb.andWhere('f.rating IN (:...posRatingFilter)', { posRatingFilter: feedbackTypes })
+
+            // Anti-join: count immediate predecessors of feedback messages that aren't
+            // themselves feedback messages (to avoid double-counting).
+            precedingCountQb = repo
```
jchui-wd: `precedingCountQb` is responsible for finding the rows immediately before of a feedback message, but they themselves are NOT a feedback message.

more indepth explanation below of code

packages/server/src/services/stats/index.ts
```diff
@@ -1,47 +1,135 @@
...
+                .createQueryBuilder('prev')
+                .select('COUNT(DISTINCT prev.id)', 'count')
+                .innerJoin(
+                    ChatMessage,
+                    'fb_msg',
+                    'fb_msg.chatflowid = prev.chatflowid AND fb_msg.sessionId = prev.sessionId AND fb_msg.createdDate > prev.createdDate'
+                )
+                .innerJoin(ChatMessageFeedback, 'fb', 'fb.messageId = fb_msg.id')
```
jchui-wd, line 83: find all the message of a session that is before the one with feedback

packages/server/src/services/stats/index.ts
```diff
@@ -1,47 +1,135 @@
...
+                .leftJoin(
+                    ChatMessage,
+                    'btwn',
+                    'btwn.chatflowid = prev.chatflowid AND btwn.sessionId = prev.sessionId AND btwn.createdDate > prev.createdDate AND btwn.createdDate < fb_msg.createdDate'
+                )
```
jchui-wd, line 88: This just finds the row that is the IMMEDIATE predecessor of a chat row with feedback.
<details>
<summary>Code explanation</summary>
for every 'prev', if it exists something in between 'btwn', then this 'prev' is NOT the immediate predecessor of a feedback, so discard it.

if there is no 'btwn' for the current 'prev, then this 'prev' is the predecessor of a feedback, keep it.
</details>

packages/server/src/services/stats/index.ts
```diff
@@ -1,47 +1,135 @@
...
+                .leftJoin(ChatMessageFeedback, 'prev_fb', 'prev_fb.messageId = prev.id AND prev_fb.rating IN (:...ft2)', {
+                    ft2: feedbackTypes
+                })
```
jchui-wd, line 91: Only keep if the predecessor itself is NOT a feedback type.

<details>
<summary>Code explanation</summary>
for each 'prev' left, check if its associated with a feedback 'prev_fb'
and later down 
`.andWhere('prev_fb.id IS NULL')` here

just keep the 'prev' that does not have feedback associated with it ('prev_fb.id' is null)
</details>

packages/server/src/services/stats/index.ts
```diff
@@ -1,47 +1,135 @@
...
-        return dbResponse
+        const [totalMessagesRaw, totalSessionsRaw, totalFeedbackRaw, positiveFeedbackRaw, precedingRaw] = await Promise.all([
+            totalMessagesQb.getRawOne(),
+            totalSessionsQb.getRawOne(),
+            totalFeedbackQb.getRawOne(),
+            positiveFeedbackQb.getRawOne(),
+            precedingCountQb?.getRawOne() ?? Promise.resolve(null)
+        ])
+
+        const totalMessages = precedingRaw
```
jchui-wd: So if no `feedbackTypes` filters was selected, we will just use `parseInt(totalMessagesRaw?.count ?? '0', 10)`

but if there is `feedbackTypes`, `precedingRaw` will be truthy and to match the old functionality we just add `totalFeedbackRaw` + `precedingRaw` count for `totalMessages`.

jchui-wd (commented):
Hi @HenryHengZJ, I provided an explanation of what the code is doing. Apologies it's a bit of a wall of text.

I think the four base stats are not too complicated `totalMessagesQb, totalSessionsQb, totalFeedbackQb, positiveFeedbackQb`

But once we add in the filters for `feedbackTypes` is when things get messy for `Total Message`. 

Old `Total Message` functionality:
1. get every single message for chatflow (sorted by date), WITH feedback data associated onto each row
2. Then loop through each message for AI responses that has feedback associated with the filter feedbackType. (totalFeedbackQb gets this part already`
3. also keep a message row that is immediately before it. (so filtered message and + 1 more row before it) (precedingCountQb below will be responsible for this section`

The Old code had it easy because it gets all message sorted, filter it by the selected `feedbackType` AND just also include the `index - 1`. 

This new way to query exactly what we want, we needed a lot of inner and left joins, which is explained at the bottom

Screenshot 2026-04-14 at 10 37 05 AM

HenryHengZJ:
thanks! that makes a lot of sense now

packages/server/src/services/stats/index.ts
```diff
@@ -1,47 +1,135 @@
...
+                    ft2: feedbackTypes
+                })
+                .where(baseWhere)
+                .andWhere('fb.rating IN (:...ft)', { ft: feedbackTypes })
+                .andWhere('btwn.id IS NULL')
+                .andWhere('prev_fb.id IS NULL')
         }
 
-        return dbResponse
+        const [totalMessagesRaw, totalSessionsRaw, totalFeedbackRaw, positiveFeedbackRaw, precedingRaw] = await Promise.all([
```
HenryHengZJ: These 4 queries share the same base WHERE and scan the same rows. Can we do it in one query, smtg like:

```
SELECT
  COUNT(*)                                                AS "totalMessages",
  COUNT(DISTINCT cm."sessionId")                          AS "totalSessions",
  COUNT(DISTINCT f.id)                                    AS "totalFeedback",
  COUNT(DISTINCT CASE WHEN f.rating = 'THUMBS_UP' THEN f.id END) AS "positiveFeedback"
FROM chat_message cm
LEFT JOIN chat_message_feedback f ON f."messageId" = cm.id
WHERE cm.chatflowid = :chatflowid ...
```

`LEFT JOIN` instead of `INNER JOIN` so `totalMessages` and `totalSessions` still count messages without feedback

Im not sure if this works, but pointing out just in case we can further optimize it

jchui-wd: I made a [changes](https://github.com/FlowiseAI/Flowise/pull/6191/changes/cd69c5d30961f9356f091e438d782d6dc1087ac2) to combine them all into [one query](https://github.com/FlowiseAI/Flowise/pull/6191/changes#diff-e2751e206551ad3c10ba6debb157be4ababe0ad6aa48e27d41815c9e048e3d52R107-R110)

Although feedbackTypes is still separated. But overall way less calls than before

https://github.com/user-attachments/assets/a44d5ac6-0db5-4676-9447-e0f12df2e5d6

HenryHengZJ (approved):

PR #6195, opened 2026-04-11
fix: guard against undefined user in auth permissions handler

vigneshakaviki:
## Summary

Fixes #6105 — `/api/v1/auth/resolve` returns 500 Internal Server Error on all Flowise 3.x versions, breaking the login flow.

## Root Cause

The `GET /api/v1/auth/:type` route (in `enterprise/controllers/auth/index.ts`) uses a `:type` parameter that matches **any** path segment, including `resolve`. When the UI calls `GET /api/v1/auth/resolve`:

1. The request is **whitelisted** (skips JWT authentication), so `req.user` is `undefined`
2. The `getAllPermissions` handler casts `req.user` to `LoggedInUser` and accesses `user.isOrganizationAdmin` (line 77)
3. This throws `TypeError: Cannot read properties of undefined (reading 'isOrganizationAdmin')`
4. Express catches it and returns 500

Note: The actual `POST /api/v1/auth/resolve` handler (registered in passport middleware) works fine — the issue is the `:type` catch-all route intercepting GET requests.

## Fix

Added a null check for `req.user` at the top of `getAllPermissions`. Returns `401 Unauthorized` before accessing any user properties.

```diff
-        const user = req.user as LoggedInUser
+        const user = req.user as LoggedInUser | undefined
+        if (!user) {
+            return res.status(StatusCodes.UNAUTHORIZED).json({ error: 'Unauthorized Access' })
+        }
```

## Test Plan

- [x] Added co-located test `index.test.ts` with 2 cases:
  - `req.user` undefined → returns 401 (fails before fix, passes after)
  - `req.user` present → returns permissions normally
- [x] Existing server tests pass with no regressions

cc @HenryHengZJ @chungyau97

### DIFF
--- a/packages/server/src/enterprise/controllers/auth/index.test.ts
+++ b/packages/server/src/enterprise/controllers/auth/index.test.ts
@@ -1,0 +1,78 @@
+import { describe, expect, it, beforeEach } from '@jest/globals'
+import { Request, Response, NextFunction } from 'express'
+
+// Mock logger before importing the module under test
+jest.mock('../../../utils/logger', () => ({
+    __esModule: true,
+    default: { warn: jest.fn(), info: jest.fn(), error: jest.fn(), debug: jest.fn() }
+}))
+
+// Mock getRunningExpressApp
+const mockGetRunningExpressApp = jest.fn()
+jest.mock('../../../utils/getRunningExpressApp', () => ({
+    getRunningExpressApp: () => mockGetRunningExpressApp()
+}))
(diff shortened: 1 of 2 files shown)


vigneshakaviki:
Thanks for the review feedback! Updated to use `StatusCodes` constants instead of magic numbers and switched the response key to `message` for consistency with the rest of the codebase. Also removed the unused `beforeEach` import.

PR #6196, opened 2026-04-12
fix(mcp): preserve JSON Schema types and required fields in createSchemaModel

### LINKED ISSUE #5109: OpenAI Agent Node Strict Tool Calling Mode fails on json schema validation with Custom MCP Tool
### Describe the bug

I get the following error (I believe from OpenAI's API) when trying to run an agent node that has a Custom MCP tool attached.

```
Error in Agent node: 400 Invalid schema for function 'get_table_schema': In context=('properties', 'tableName'), schema must have a 'type' key.
```

This is the json schema for my tool function:
```json
{
  "type": "object",
  "properties": {
  	"tableName": {
  		"type": "string",
  		"description": "Name of the table or view to get schema details for"
  	}
  },
  "required": ["tableName"]
}
```

I tried many variations of the JSON schema to rule out actual problems there, so I believe the issue does not lie with the schema itself.

I dug into the codebase with the help of Claude and this is eventually what I got to, basically that the issue may lie in the json schema -> zod -> json schema (for OpenAI API) pipeline. This could be wrong, but it was the best I could do without running Flowise locally and debugging there. Whatever is coming out of `packages/components/nodes/agents/OpenAIAssistant/OpenAIAssistant.ts:1159` as the `parameters` object does not have the `type` keys from my originally passed in json schema.

From Claude:
```
 Here's the complete flow of how your MCP JSON schema becomes the tool object in the OpenAI Assistant:

  Schema Conversion Pipeline

  1. MCP Server → MCP Toolkit

  - File: packages/components/nodes/tools/MCP/core.ts:90
  - Your MCP server returns tools with inputSchema (JSON Schema)
  - this._tools = await this.client.request({ method: 'tools/list' }, ListToolsResultSchema)

  2. JSON Schema → Zod Schema

  - File: packages/components/nodes/tools/MCP/core.ts:111
  - argsSchema: createSchemaModel(tool.inputSchema) converts your JSON schema
  - However, createSchemaModel() at line 159-175 loses all type information:
  const schemaProperties = Object.entries(inputSchema.properties).reduce((acc, [key, _]) => {
      acc[key] = z.any()  // <-- All properties become z.any()!
      return acc
  }, {} as Record<string, import('zod').ZodTypeAny>)

  3. Zod Schema → LangChain Tool

  - File: packages/components/nodes/tools/MCP/core.ts:135
  - tool() function from @langchain/core/tools creates a LangChain Tool object
  - This is NOT a DynamicStructuredTool, just a regular Tool

  4. Tool → OpenAI Assistant

  - File: packages/components/nodes/agents/OpenAIAssistant/OpenAIAssistant.ts:1158
  - The tool parameter is the LangChain Tool from step 3
  - zodToJsonSchema(tool.schema) converts back to JSON Schema for OpenAI
  - Since MCP tools are regular Tool objects (not DynamicStructuredTool), the strict validation logic at lines 1165 and 1194 never 
  applies

  The Problem

  You're exactly right - there's a lossy conversion:
  1. Your MCP JSON Schema (with proper types, constraints, etc.)
  2. → Zod schema with z.any() (loses all type information)
  3. → JSON Schema (basic object structure only)

  The DynamicStructuredTool.isStrict() method only applies to OpenAPI tools, not MCP tools, so your MCP tools never benefit from strict
  schema validation even when "Strict Tool Calling" is enabled.
```

### To Reproduce

1. Create an Agentflow (v2)
2. Create an Agent node - ChatOpenAI type
3. Ensure "Strict Tool Calling" property is enabled
4. Add a CustomMCP tool with a tool function that a comparable schema (an object with any simple properties)
5. Run flow, will always break with similar above error

### Expected behavior

I can't seem to use OpenAI's strict tool calling feature with my CustomMCP server tools

### Screenshots

_No response_

### Flow

_No response_

### Use Method

Docker

### Flowise Version

3.0.5

### Operating System

Linux

### Browser

Chrome

### Additional context

_No response_

publinchi:
## Problem

The `createSchemaModel` function in `packages/components/nodes/tools/MCP/core.ts` maps every MCP tool property to `z.any()`, discarding all type information from the original `inputSchema`. This causes two critical issues:

1. **`required` array is ignored** — all parameters become optional in the resulting Zod schema. LLMs then call tools with empty arguments `{}`, missing required parameters like `login` or `project_id`.

2. **Type information is lost** — `string`, `integer`, `boolean`, `array`, `anyOf` types are all silently dropped. When the Zod schema is later converted back to JSON Schema for the OpenAI API, property `type` keys are missing, causing strict mode rejections.

This issue was also reported in #5109.

## Fix

Adds a `jsonSchemaToZod()` helper function that properly maps JSON Schema types to Zod types:

- `string` → `z.string()`
- `integer` / `number` → `z.number()`
- `boolean` → `z.boolean()`
- `array` → `z.array(itemType)`
- `object` → `z.record(z.any())`
- `anyOf` with nullable → proper nullable/optional Zod type
- unknown → `z.any()` (safe fallback)

Also reads the `required` array from `inputSchema` and marks non-required fields as `.optional()` in the resulting Zod schema, so LLMs correctly identify which parameters must be provided.

## Testing

Verified with a local MCP server (Redmine) that:
- Tools with required parameters (e.g. `login: string`) now correctly appear as required in the OpenAI function calling schema
- LLMs no longer call tools with empty `{}` arguments
- Optional parameters with `anyOf: [{type: string}, {type: null}]` are correctly typed as optional strings

Fixes #5109


HenryHengZJ:
hey thanks for the PR! I've opened https://github.com/FlowiseAI/Flowise/pull/6248 that fix this in a better way, without parsing each of the types. closing this

PR #6205, opened 2026-04-13
feat: implement the custom MCP Server in tool

prd-hoang-doan:
## Ticket
https://github.com/orgs/FlowiseAI/projects/2?pane=issue&itemId=168148790

## Changes
- Implemented the MCPServersTable component for displaying MCP server data with status badges and tools count.
- Created CustomMcpServerDialog for adding and editing custom MCP servers, including validation for server URL and handling authentication headers.
- Integrated MCP server management into the Tools view with tabbed navigation for Custom Tools and Custom MCP Servers.
- Added pagination and loading states for MCP servers.

## Database

Field | Type | Nullable | Default | Description
-- | -- | -- | -- | --
id | uuid | No | Auto-generated | Primary key for the custom MCP server
name | string | No | — | Display name of the MCP server
serverUrl | string | No | — | Endpoint URL of the MCP server
iconSrc | string | Yes | null | Optional icon URL for UI display
color | string | Yes | null | Optional color used for UI presentation
authType | string | No | NONE | Authentication type (NONE, CUSTOM_HEADERS, OAUTH2 (future))
authConfig | text | Yes | null | Authentication configuration payload
tools | text | Yes | null | Serialized tool definitions returned by MCP server
status | string | No | PENDING | Connection status (PENDING, ERROR, AUTHORIZED)
createdDate | timestamp | No | Auto-generated | Record creation timestamp
updatedDate | timestamp | No | Auto-generated | Last update timestamp
workspaceId | text | No | — | Workspace identifier associated with this server

### Notes
- `authConfig` is decrypted as a string in the database.
- `tools` are stored as serialized JSON text. 
```json
{"tools":[{"name":"Github MCP Server","description":"GitHub MCP Server to search user's repositories.","inputSchema":{"type":"object","properties":{"query":{"type":"string","description":"User Input/Question content"}},"required":["query"]}}]}
```

## Testing
Platform | Auth Type | Success
-- | -- | -- 
Dify | NONE | ✅
n8n | CUSTOM_HEADERS | ✅

## Demo:
Youtube: https://youtu.be/JO8IBriEzFA


prd-hoang-doan:
Hi @harshit-flowise, @jchui-wd
Could you take your time to review this pull request? Feel free to provide feedback on this.

0xbrainkid:
The custom MCP server table with status badges and auth type handling is a solid foundation — especially the `authType` enum leaving room for OAuth2 in a future iteration.

One column worth adding to the MCP server schema: **trust attestation.** The current schema tracks connection status (PENDING/ERROR/AUTHORIZED) and auth configuration. But "AUTHORIZED" tells you the server authenticated successfully — it does not tell you whether the MCP server has been reliable in practice or whether the agent behind it is who it claims to be.

A lightweight addition:

```
Field          | Type    | Nullable | Default | Description
trustScore     | float   | Yes      | null    | Behavioral trust score (0-1) from attestation provider
attestationUrl | string  | Yes      | null    | Link to verifiable trust proof
lastVerified   | timestamp | Yes    | null    | When trust was last checked
```

This matters especially for custom (user-added) MCP servers, which have no prior vetting — unlike official Flowise integrations. A user adding a third-party MCP server endpoint should be able to see at a glance whether that server has a verified trust record before connecting it to their production flows.

The `status` badge UI already has the visual pattern: adding a trust level badge (verified/unverified/low/high) next to the connection status badge would give flow builders the signal they need without adding friction to the current UX.

packages/server/src/controllers/custom-mcp-servers/index.ts
```diff
@@ -0,0 +1,198 @@
...
+        if (!workspaceId) {
+            throw new InternalFlowiseError(
+                StatusCodes.NOT_FOUND,
+                `Error: customMcpServersController.createCustomMcpServer - workspace not found!`
+            )
+        }
+        const body = req.body
+        // Explicit allowlist — id/workspaceId/timestamps must not be overrideable by client
+        const mcpBody: Record<string, unknown> = {}
+        if (body.name !== undefined) mcpBody.name = body.name
```
HenryHengZJ, line 49: we dont have the explictly set each key, we can reuse the common function `stripProtectedFields`

packages/server/src/controllers/custom-mcp-servers/index.ts
```diff
@@ -0,0 +1,198 @@
...
+        const workspaceId = req.user?.activeWorkspaceId
+        if (!workspaceId) {
+            throw new InternalFlowiseError(
+                StatusCodes.NOT_FOUND,
+                `Error: customMcpServersController.updateCustomMcpServer - workspace not found!`
+            )
+        }
+        const body = req.body
+        // Explicit allowlist
+        const mcpBody: Record<string, unknown> = {}
```
HenryHengZJ, line 129: same here

HenryHengZJ:
> trust attestation

how do you propose we check for the trust attestation ?

HenryHengZJ:
~~When we add new MCP server, can we do Add & Authorize in one go? Instead of having users to save it first, then authorize separately~~

Updated now in this [commit](https://github.com/FlowiseAI/Flowise/pull/6205/commits/b4a8d1e8533082795072e67aeea5f5b7c181d4f9)

HenryHengZJ:
In this [commit](https://github.com/FlowiseAI/Flowise/pull/6205/commits/a2ddfd9759ec1dc39447541c00a11cb8cc1e4b15) I've made some changes for hardening the security, scalability issue.

## Security

- **SSRF**: reuse `checkDenyList` from `httpSecurity.ts` on create/update. Rejects RFC1918, loopback, link-local, IMDS, multicast, ULA.
- **Workspace isolation**: `CustomMcpServerTool` now requires and filters by `workspaceId` (from `options.workspaceId` at init, `options.searchOptions.workspaceId` in loadMethods).
- **Header validation**: new shared `validateCustomHeaders` in `packages/components/src`. RFC 7230 tokens, CRLF rejection, denies hop-by-hop / `Host` / `Cookie` / `Proxy-*` / `X-Forwarded-*` / `Sec-*`. Caps: 25 headers, 128-char keys, 2 KB values.
- **authType enum**: controller validates against `CustomMcpServerAuthType` on create + update.

## Scalability

- **Tools blob cap**: new env `CUSTOM_MCP_TOOLS_MAX_BYTES` (default 512 KB, `0` disables). Row marked `ERROR` on overflow. This is to prevent large blob of malicious JSON tools
- **Index**: single composite `IDX_custom_mcp_workspace_updated (workspaceId, updatedDate)` across all four dialects. `workspaceId` normalized to `varchar(36)` in MySQL/MariaDB.
- **Pagination**: list controller defaults to `page=1, limit=50`, hard-caps `limit` at 500.
- **`select: false` + `toolCount`**: `tools` excluded from list queries; new `toolCount` int populated at authorize. Detail/`/:id/tools` opt in via `addSelect`.

## UI

- Tools page: toolbar now on the same row as tabs — no more blank gap.
- **Discovered Tools** redesigned:
  - Theme-aware icon, friendly `annotations.title`
  - Safety chips: **READ-ONLY** / **DESTRUCTIVE** / **EXTERNAL**
  - Per-parameter rows with required/optional badge, typed chip, enum values, default, description
  - Search, expand/collapse, dark-mode contrast fixes
- Dialog fetches tools from `GET /:id/tools` (blob no longer on list responses).

image

packages/server/src/controllers/custom-mcp-servers/index.ts
HenryHengZJ: ignore me, if we swapped to `stripProtectedFields`, a client could POST `{ name, serverUrl, toolCount: 99999, status: 'AUTHORIZED', tools: '{"tools":[...fake...]}' }` and bypass the authorize flow entirely — marking a server "authorized" without ever contacting the remote MCP, and injecting arbitrary tool definitions.

HenryHengZJ (approved):

prd-hoang-doan:
Thank @HenryHengZJ for feedback and fixing some existing security issue and enhance UIUX to complete the feature.
For Trust Attestation, I think we can implement until the [SEP-2127: MCP Server Card](https://github.com/modelcontextprotocol/modelcontextprotocol/pull/2127) become final then Flowise will fetch well-known metadata endpoint from this to validate and capture TLS certificates as well. 
In next phase, I propose that we only need update the entity schema like

Field | Type | Nullable | Default | Description
-- | -- | -- | -- | --
trustLevel | string | No | USER_APPROVED | UNTRUSTED, USER_APPROVED, REGISTRY_VERIFIED
approvedByUserId | uuid | Yes | null | User who approved
approvedAt | timestamp | Yes | null | When approved

In the phase 1, the trust level basically is USER_APPROVED, we can add a checkbox for acknowledgement whenever user add a new MCP Server like:
☐ I trust `https://my-custom-mcp.example.com` and take responsibility for data shared with it.

What's your thought about this?

HenryHengZJ:
I think thats a good idea! Lets do it after server card is ready, in another PR.

PR #6206, opened 2026-04-13
Bugfix/Silent fallback to vm2 for sandbox execution

HenryHengZJ:
silent fallback to vm2 for sandbox execution


HenryHengZJ:
https://github.com/FlowiseAI/Flowise/pull/6168 introduced changes that force users to use E2B for sandbox, reverting that as its not a must for self hosted OSS users to use that, silent fallback to using patched version of vm2

henry-heng-wd (approved):

PR #6212, opened 2026-04-14
feat(agentflow): add client filtering for form input options in Start node

jocelynlin-wd:
- Enhanced Start node to include a client property for the form input option, allowing differentiation between 'agentflowv2' and 'agentflowsdk'.
- Added tests to verify that the 'formInput' option is correctly filtered based on the client type.

FLOWISE-510

https://github.com/user-attachments/assets/a0b4250f-eed8-4be8-bdb3-9b1bdedf69cb

### DIFF
--- a/packages/components/nodes/agentflow/Start/Start.ts
+++ b/packages/components/nodes/agentflow/Start/Start.ts
@@ -39,7 +39,8 @@ class Start_Agentflow implements INode {
                     {
                         label: 'Form Input',
                         name: 'formInput',
-                        description: 'Start the workflow with form inputs'
+                        description: 'Start the workflow with form inputs',
+                        client: ['agentflowv2']
                     }
                 ],
                 default: 'chatInput'
--- a/packages/server/src/services/nodes/nodes.test.ts
+++ b/packages/server/src/services/nodes/nodes.test.ts
@@ -178,6 +178,42 @@ describe('filterNodeByClient', () => {
         })
     })
 
+    describe('Start node form input filtering', () => {
+        const startNode = makeNode({
+            inputs: [
+                {
+                    label: 'Input Type',
+                    name: 'startInputType',
+                    type: 'options',
+                    options: [
+                        { label: 'Chat Input', name: 'chatInput' },
+                        { label: 'Form Input', name: 'formInput', client: ['agentflowv2'] }
+                    ],
+                    default: 'chatInput'
+                },
+                { label: 'Form Title', name: 'formTitle', type: 'string' },
+                { label: 'Form Description', name: 'formDescription', type: 'string' },
+                { label: 'Form Input Types', name: 'formInputTypes', type: 'array' },
+                { label: 'Ephemeral Memory', name: 'startEphemeralMemory', type: 'boolean' },
+                { label: 'Flow State', name: 'startState', type: 'array' }
+            ]
+        })
+
+        it('removes formInput option for agentflowsdk', () => {
+            const result = filterNodeByClient(startNode, 'agentflowsdk')
+            const optionNames = result.inputs![0].options!.map((o: any) => o.name)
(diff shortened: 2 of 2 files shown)


packages/components/nodes/agentflow/Start/Start.ts
```diff
@@ -39,7 +39,8 @@ class Start_Agentflow implements INode {
                     {
                         label: 'Form Input',
                         name: 'formInput',
-                        description: 'Start the workflow with form inputs'
+                        description: 'Start the workflow with form inputs',
+                        client: ['agentflowv2']
```
jocelynlin-wd, line 43: didn't bump the version since there is no change to the runtime schema

packages/server/src/services/nodes/nodes.test.ts
```diff
@@ -178,6 +178,42 @@ describe('filterNodeByClient', () => {
...
+                { label: 'Ephemeral Memory', name: 'startEphemeralMemory', type: 'boolean' },
+                { label: 'Flow State', name: 'startState', type: 'array' }
+            ]
+        })
+
+        it('removes formInput option for agentflowsdk', () => {
+            const result = filterNodeByClient(startNode, 'agentflowsdk')
+            const optionNames = result.inputs![0].options!.map((o: any) => o.name)
+            expect(optionNames).toContain('chatInput')
+            expect(optionNames).not.toContain('formInput')
```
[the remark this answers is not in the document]
jocelynlin-wd: Keeping the current approach — mapping to names and asserting with toContain/not.toContain reads more directly for this case since we're specifically checking presence/absence of named options. The any cast is a minor tradeoff but the assertions are more expressive about intent.

packages/server/src/services/nodes/nodes.test.ts
```diff
@@ -178,6 +178,42 @@ describe('filterNodeByClient', () => {
...
+        })
+
+        it('keeps formInput option for agentflowv2', () => {
+            const result = filterNodeByClient(startNode, 'agentflowv2')
+            const optionNames = result.inputs![0].options!.map((o: any) => o.name)
+            expect(optionNames).toContain('chatInput')
+            expect(optionNames).toContain('formInput')
```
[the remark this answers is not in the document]
jocelynlin-wd: Same as above https://github.com/FlowiseAI/Flowise/pull/6212#discussion_r3076632535 — sticking with the toContain pattern here for consistency between the two test cases.

packages/components/nodes/agentflow/Start/Start.ts
HenryHengZJ: what does `client` means? I wonder instead of making the changes here in OSS, should we create a new node in `extension` that extends the original Start class, but override the params to exclude `formInput`.

my main concern is that if we make changes here, this `cllient` value will get stored in every saved flow moving forward for every users, making it hard in future if we want to get rid/make further changes

jocelynlin-wd: (shared the tech doc offline)

The client field was introduced in PR #6107 as a general-purpose mechanism for scoping node params/options per client (agentflowv2 vs agentflowsdk). It lives on the param definition (schema), not on saved flow data — `filterNodeByClient()` on the server strips it before sending the API response, so it never gets stored in flows.

The extension approach would work too, but it means maintaining a parallel Start node that needs to stay in sync with the OSS one. Since the client filtering infrastructure is already in place and doesn't affect saved data, this felt like the lighter-weight path. 

Let me know if this makes sense to you or if you still prefer strongly to go with the extension route.

HenryHengZJ (approved):

packages/components/nodes/agentflow/Start/Start.ts
HenryHengZJ: yep agree with that

PR #6227, opened 2026-04-16
feat:533 webhook secrets

jchui-wd:
Adds opt-in secret-based authentication for webhook-triggered agentflows.
Is Tied to https://github.com/FlowiseAI/Flowise/pull/6217

## Changes
**Backend**
- New `webhookSecret` (hidden) and `webhookSecretConfigured` (boolean) columns on `ChatFlow`, with migrations for all 4 databases
- `POST /chatflows/:id/webhook-secret` and `DELETE /chatflows/:id/webhook-secret` to generate/remove secrets
- Signature verification runs before all other validation — supports HMAC-SHA256/SHA1 (GitHub, Stripe, Slack) and plain token (GitLab)

**Start Node (v1.2)**
- When webhookTrigger is selected, now shows a generate secrets section with signature header, and signature type.

**UI**
- webhookURL field renders the full trigger URL with a copy button
- webhookSecret field manages generate/regenerate/remove with the plaintext shown only immediately after generation, then masked on reload

<details>
  <summary>Test Videos Demo</summary>

  <b>WrongSignatureType-Fail.mp4</b>
  <video src="https://github.com/user-attachments/assets/8933fa03-59a1-405e-a105-88bc758053f9" controls width="100%"></video>

  <b>PlainToken-WrongHeaderName-Fail.mp4</b>
  <video src="https://github.com/user-attachments/assets/4120612e-f9fb-41db-bd4e-c46ad20e53f6" controls width="100%"></video>

  <b>PlainToken-working.mp4</b>
  <video src="https://github.com/user-attachments/assets/70117138-0267-4bbf-a0ae-74afd67e6bd4" controls width="100%"></video>

  <b>No-Secret-401.mp4</b>
  <video src="https://github.com/user-attachments/assets/e593045f-48e1-44cd-a257-5a652b02f417" controls width="100%"></video>

  <b>HMAC-non256-working.mp4</b>
  <video src="https://github.com/user-attachments/assets/1f111d54-91ed-4bbe-b3ee-c9f281b3f067" controls width="100%"></video>

  <b>HMAC-256-working.mp4</b>
  <video src="https://github.com/user-attachments/assets/d5b41de3-26ab-4c49-ba03-51a9295a101b" controls width="100%"></video>
</details>

## Screenshots
Screenshot 2026-04-16 at 12 13 14 PM

Signature Type
Screenshot 2026-04-16 at 12 13 32 PM

No Secrets Generated
Screenshot 2026-04-16 at 12 13 23 PM

Tests
Screenshot 2026-04-16 at 12 24 22 PM

### DIFF
--- a/packages/server/src/database/entities/ChatFlow.ts
+++ b/packages/server/src/database/entities/ChatFlow.ts
@@ -64,6 +64,12 @@ export class ChatFlow implements IChatFlow {
     @Column({ nullable: true, type: 'text' })
     mcpServerConfig?: string
 
+    @Column({ nullable: true, type: 'text', select: false })
+    webhookSecret?: string | null
+
+    @Column({ nullable: true, default: false })
+    webhookSecretConfigured?: boolean
+
     @Column({ nullable: false, type: 'text' })
     workspaceId: string
 }
--- a/packages/components/nodes/agentflow/Start/Start.ts
+++ b/packages/components/nodes/agentflow/Start/Start.ts
@@ -175,6 +175,45 @@ class Start_Agentflow implements INode {
                 }
             },
             {
+                label: 'Webhook Secret',
+                name: 'webhookSecret',
+                type: 'string',
+                description:
+                    'Optional secret used to verify incoming requests. When set, configure Signature Header and Signature Type below to match your sender.',
+                optional: true,
+                show: {
+                    startInputType: 'webhookTrigger'
+                }
+            },
+            {
+                label: 'Signature Header',
+                name: 'webhookSignatureHeader',
+                type: 'string',
+                description:
+                    'The request header that carries the signature. e.g. x-hub-signature-256 for GitHub, stripe-signature for Stripe, x-gitlab-token for GitLab.',
+                placeholder: 'x-webhook-signature',
+                optional: true,
+                show: {
+                    startInputType: 'webhookTrigger'
+                }
+            },
+            {
+                label: 'Signature Type',
+                name: 'webhookSignatureType',
+                type: 'options',
+                description:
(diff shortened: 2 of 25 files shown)


packages/server/src/database/entities/ChatFlow.ts
```diff
@@ -64,6 +64,12 @@ export class ChatFlow implements IChatFlow {
     @Column({ nullable: true, type: 'text' })
     mcpServerConfig?: string
 
+    @Column({ nullable: true, type: 'text', select: false })
```
HenryHengZJ, line 67: what do u think if we group these into a column like `webhookConfig`, similar to `mcpServerConfig`

jchui-wd: @HenryHengZJ 

I can group `webhookSecret` and `webhookSecretConfigured` make it as a `webhookConfig` just like `mcpServerConfig`, but I separated them intentionally because I didn't want the `webhookSecret` to be exposed when a user does a `api/chatflow` call  since the secret is stored 'as is'.

so I added `select: false` for `webhookSecrets`, but we still need to know if it is configured so the Start.node can render out the correct UI if there is already a secret hence `webhookSecretConfigured`.

Any thoughts on this?

I think I should also add `webhookSecret` and `webhookSecretConfigured` in `stripProtectedFields.ts`, but I'll hold off until we are aligned on this.

mmattu-wd: I like the idea of isolating the secret value, but are we really storing it in plain text?

henry-heng-wd: i see it make sense, lets stick to this approach

henry-heng-wd (approved):

PR #6243, opened 2026-04-17
fix: preserve metadata in RetrieverTool when using Custom Retriever

### LINKED ISSUE #6242: Metadata lost when using Custom Retriever with Retriever Tool in Agent (Qdrant)
### Describe the bug

When using a Custom Retriever with Qdrant that correctly formats metadata (e.g. {{metadata.docname}}, {{metadata.url}}, etc.), the metadata is not preserved when passing through the Retriever Tool into an Agent.

The Custom Retriever works correctly when used standalone (output = text or document), but when connected to a Retriever Tool and used inside an Agent, only the pageContent is returned and metadata is lost.

This breaks use cases where metadata is required (e.g. file names, URLs, blob type, citations).

### To Reproduce

1. Create a Qdrant vector store with documents that include metadata (docname, url, blobType, etc.)
2. Add a Custom Retriever node with the following result format:
    {{context}}
    [FileName]: {{metadata.docname}}
    Url: {{metadata.url}}
    blobType: {{metadata.blobType}}
3. Connect:
   Qdrant → Custom Retriever → Retriever Tool → Agent
4. Ask a question to the agent
5. Observe the response

### Expected behavior

The agent response should preserve metadata from the Custom Retriever.

Example:
Some document content...
[FileName]: document.pdf
Url: https://...
blobType: application/pdf

### Screenshots

Image

### Flow

_No response_

### Use Method

None

### Flowise Version

Flowise 3.1.1

### Operating System

Windows

### Browser

Chrome

### Additional context

_No response_

limbergarcia:
## 🐛 Bug Description
When using Custom Retriever with Retriever Tool, metadata from vector store (e.g. Qdrant) is lost and only `pageContent` is preserved.

## Root Cause
CustomRetriever overrides `getRelevantDocuments()` and returns formatted documents, but RetrieverTool internally calls `retriever.invoke()`, which uses LangChain's `_getRelevantDocuments()` instead.

This causes:
- Custom formatting to be ignored
- Metadata handling inconsistency

## Fix
Updated CustomRetriever to ensure compatibility with LangChain's `invoke()` flow and preserve metadata correctly.

## Steps to Reproduce
1. Use Qdrant vector store with metadata
2. Connect:
   Qdrant → CustomRetriever → RetrieverTool → Agent
3. Query documents
4. Metadata is missing in response

## Expected Behavior
Metadata fields like:
- filename
- url
- blobType

should be preserved and accessible in the final output.

## 🔗 Related Issue
Closes #6242


limbergarciavilchez:
@HenryHengZJ could you please take a look at this improvement? Your review would be appreciated.

PR #6248, opened 2026-04-19
fix(mcp): preserve tool input schema types and required fields

### LINKED ISSUE #5109: OpenAI Agent Node Strict Tool Calling Mode fails on json schema validation with Custom MCP Tool

HenryHengZJ:
## Summary

- MCP tools were losing all type information and `required` fields because `createSchemaModel` mapped every property to `z.any()`
- Rather than reimplement a JSON-Schema-to-Zod converter (lossy for `$ref`, `allOf`, `anyOf`, enums, nested objects), pass the MCP server's `inputSchema` straight through to LangChain's `tool()` — the `JsonSchema7Type` overload has been supported since `@langchain/core` 1.1.x.
- Downstream tool serializers (`Agent`, `Tool`, `OpenAIAssistant`) to uptake these changes

### Example 1 — typed params with required fields

MCP server with the following `inputSchema` for a `create_issue` tool:

```json
{
  "type": "object",
  "properties": {
    "owner":      { "type": "string",  "description": "Repo owner" },
    "repo":       { "type": "string",  "description": "Repo name" },
    "title":      { "type": "string" },
    "project_id": { "type": "integer" },
    "draft":      { "type": "boolean" }
  },
  "required": ["owner", "repo", "title"]
}
```

**Before** — what the LLM saw after `createSchemaModel`:

```json
{
  "type": "object",
  "properties": {
    "owner":      {},
    "repo":       {},
    "title":      {},
    "project_id": {},
    "draft":      {}
  }
}
```

LLM response:

```json
{ "name": "create_issue", "arguments": {} }
```

Tool call fails — `owner`, `repo`, `title` missing. OpenAI strict mode rejects the schema outright because property `type` keys are absent.

**After** — we straightaway pass the raw `inputSchema` and preserve everything. LLM response:

```json
{
  "name": "create_issue",
  "arguments": {
    "owner": "flowiseai",
    "repo":  "Flowise",
    "title": "Fix MCP schema handling"
  }
}
```

### Example 2 — enum constraints

```json
{
  "type": "object",
  "properties": {
    "priority": { "type": "string", "enum": ["low", "medium", "high"] }
  },
  "required": ["priority"]
}
```

**Before:** `priority` was `z.any()` — LLM would pass `"urgent"`, `"P1"`, or anything else.
**After:** the enum is preserved; the model provider enforces one of `low | medium | high`.

### Example 3 — nested object

```json
{
  "type": "object",
  "properties": {
    "filter": {
      "type": "object",
      "properties": {
        "status": { "type": "string" },
        "since":  { "type": "string", "format": "date-time" }
      },
      "required": ["status"]
    }
  },
  "required": ["filter"]
}
```

**Before:** nested `filter` became `z.any()` — the LLM received no guidance on inner structure and commonly sent `"filter": "open"` as a string.
**After:** full nested schema reaches the model; inner `required` and `format` are preserved.

### Example 4 — no input schema

A tool like `get_current_time` with no params.

**Before:** `createSchemaModel` threw `Invalid schema type or missing properties`; the tool was silently dropped from the toolkit because `Promise.allSettled` swallowed the rejection.
**After:** tool registers correctly with an empty object schema.

### DIFF
--- a/packages/components/src/utils.ts
+++ b/packages/components/src/utils.ts
@@ -2320,7 +2320,7 @@ type ZodToJsonSchemaInput = Parameters<typeof zodToJsonSchema>[0]
  * whichever Zod major version (`^3 || ^4`) TypeScript resolves at the call site.
  */
 export const isZodSchema = (schema: unknown): schema is ZodToJsonSchemaInput =>
-    typeof schema === 'object' && schema !== null && '_def' in schema
+    typeof schema === 'object' && schema !== null && '_def' in schema && typeof (schema as { parse?: unknown }).parse === 'function'
 
 /**
  * Normalizes a tool schema into a plain JSON Schema object.
@@ -2331,10 +2331,11 @@ export const isZodSchema = (schema: unknown): schema is ZodToJsonSchemaInput =>
  * `$schema` marker so the result is safe to embed in LLM tool definitions.
  */
 export const toolSchemaToJsonSchema = (schema: unknown): ToolJsonSchema => {
+    if (schema == null) return { type: 'object', properties: {} }
     const jsonSchema: ToolJsonSchema = isZodSchema(schema)
         ? (zodToJsonSchema(schema) as ToolJsonSchema)
         : cloneDeep(schema as ToolJsonSchema)
-    if (jsonSchema && jsonSchema.$schema) {
+    if (jsonSchema.$schema) {
         delete jsonSchema.$schema
     }
     return jsonSchema
--- a/packages/components/src/utils.ts
+++ b/packages/components/src/utils.ts
@@ -22,6 +22,7 @@ import { NodeVM } from 'vm2'
 import { Sandbox } from '@e2b/code-interpreter'
 import { secureFetch, checkDenyList, secureAxiosRequest } from './httpSecurity'
 import JSON5 from 'json5'
+import zodToJsonSchema, { type JsonSchema7Type } from 'zod-to-json-schema'
 
 export const numberOrExpressionRegex = '^(\\d+\\.?\\d*|{{.*}})$' //return true if string consists only numbers OR expression {{}}
 export const notEmptyRegex = '(.|\\s)*\\S(.|\\s)*' //return true if string is not empty or blank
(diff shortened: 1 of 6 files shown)


packages/components/src/utils.ts
```diff
@@ -2303,3 +2304,38 @@ export const isReasoningModelOpenAI = (name: string): boolean => {
...
+
+/**
+ * Normalizes a tool schema into a plain JSON Schema object.
+ *
+ * LangChain tools may expose their `schema` as either a Zod schema (has `_def`)
+ * or an already-plain JSON Schema (e.g. MCP tools). This helper handles both,
+ * deep-clones plain objects to prevent accidental mutation, and strips the
+ * `$schema` marker so the result is safe to embed in LLM tool definitions.
+ */
+export const toolSchemaToJsonSchema = (schema: unknown): ToolJsonSchema => {
```
jocelynlin-wd, line 2333: medium-priority
> `toolSchemaToJsonSchema(null/undefined)` returns null but its return type is non-nullable. Callers in Tool.ts could get a surprise null where they spread .properties. Easy fix: add an early return of `{ type: 'object', properties: {} }` for falsy input

packages/components/src/utils.ts
```diff
@@ -2303,3 +2304,38 @@ export const isReasoningModelOpenAI = (name: string): boolean => {
...
+
+type ZodToJsonSchemaInput = Parameters<typeof zodToJsonSchema>[0]
+
+/**
+ * Type guard detecting a Zod schema without importing Zod's types directly.
+ *
+ * Using `Parameters<typeof zodToJsonSchema>[0]` keeps the guard compatible with
+ * whichever Zod major version (`^3 || ^4`) TypeScript resolves at the call site.
+ */
+export const isZodSchema = (schema: unknown): schema is ZodToJsonSchemaInput =>
```
jocelynlin-wd, line 2322: medium-priority
> `isZodSchema` false positive — a plain JSON Schema object that happens to have a `_def` key gets passed to `zodToJsonSchema`, which would throw or produce garbage. Low probability but a real edge case. Easy fix: also check for `typeof schema.parse === 'function'`.

jocelynlin-wd (approved):

packages/components/src/utils.ts
HenryHengZJ: resolved https://github.com/FlowiseAI/Flowise/pull/6248/commits/8341c4ba64e9a4c09f3a902318565a56e5429a03

packages/components/src/utils.ts
HenryHengZJ: resolved https://github.com/FlowiseAI/Flowise/pull/6248/commits/8341c4ba64e9a4c09f3a902318565a56e5429a03

PR #6269, opened 2026-04-22
Fix FLOWISE-319 Sanitize User

yau-wd:
FLOWISE-319

### DIFF
--- a/packages/server/src/enterprise/routes/account.route.ts
+++ b/packages/server/src/enterprise/routes/account.route.ts
@@ -16,8 +16,6 @@ router.post(
     accountController.invite
 )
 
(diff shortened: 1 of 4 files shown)


igor-magun-wd (approved):

packages/server/src/enterprise/routes/account.route.ts
```diff
@@ -16,8 +16,6 @@ router.post(
     accountController.invite
 )
 
-router.post('/login', accountController.login)
```
HenryHengZJ, line 19: are we not using this endpoint? how can user login if we remove this

HenryHengZJ (approved):

PR #6271, opened 2026-04-22
add langsmith tracing initialized from environment variables

mmattu-wd:
FLOWISE-565

### Problem
When langchain is provided with certain environment variables it automatically starts auto tracing requests for Langsmith. See [this](https://mintlify.wiki/langchain-ai/langsmith-sdk/guides/tracing-with-langchain#automatic-tracing). This results in Agentflow logging to show single record traces, for every node execution in the agent.

### Solution 
A new tracingEnv module reads LangSmith settings from the environment (both LANGSMITH_* and legacy LANGCHAIN_* prefixes) and strips the LangChain tracing-flag vars out of process.env on first read so the built-in auto-tracer can no longer fire. applyEnvTracingProviders shapes env configs into the same provider entries the UI produces and merges them into additionalCallbacks and AnalyticHandler.init(), with UI winning on conflict and env-injected providers supplying credentials inline.

This will essentially treat the environment variable as if it were loaded via the UI. Tracing works the exact same.

### Verification
Plugged into Langsmith environment. All runs are in there but here is the data. 

#### Before
Before any changes made
<details>
  <summary>Case: Chatflow</summary>
image
</details>

<details>
  <summary>Case: Agentflow</summary>
image
</details>

#### After
<details>
  <summary>Case: Agentflow - UI Configured</summary>
42c73afb-782d-43a3-b41b-b70855988db9
</details>

<details>
  <summary>Case: Agentflow - Env Var Configured</summary>
9d59a742-7a0d-4618-874b-13367f54f803
</details>

<details>
  <summary>Case: Chatflow - UI Configured</summary>
4b9b3081-c58c-4101-a2c5-2e738084c362
</details>

<details>
  <summary>Case: Chatflow - Env Var Configured</summary>
fc7de669-deaa-45cf-9ba0-8fc249a84eb0
</details>

<details>
  <summary>Case: Multistep Agent - Env Var Configured</summary>
d81f647d-6d6a-40bb-b8e4-1c0d1b379c58
</details>

<details>
  <summary>Case: Multistep Agent - UI Configured</summary>
bd98377a-b94b-4983-8f2f-1ada56c3953b
</details>


HenryHengZJ (approved):

packages/components/src/tracingEnv.ts
```diff
@@ -0,0 +1,111 @@
...
+    tracingEnvConfigCache.clear()
+}
+
+/**
+ * Env var flags that activate LangChain's built-in auto-tracer (see @langchain/core
+ * `isTracingEnabled()`). Once Flowise adopts the tracing config, these must be cleared from
+ * `process.env` so the auto-tracer doesn't emit duplicate top-level runs alongside Flowise's
+ * manual `onLLMStart`/`onToolStart` child RunTrees.
+ */
+const LANGCHAIN_TRACING_FLAG_VARS = ['LANGSMITH_TRACING', 'LANGCHAIN_TRACING_V2', 'LANGSMITH_TRACING_V2', 'LANGCHAIN_TRACING'] as const
```
jocelynlin-wd, line 46: `LANGCHAIN_TRACING_FLAG_VARS` deletes 4 vars (`LANGSMITH_TRACING`, `LANGCHAIN_TRACING_V2`, `LANGSMITH_TRACING_V2`, `LANGCHAIN_TRACING`), but `getLangSmithEnvConfig` only checks `LANGSMITH_TRACING` and `LANGCHAIN_TRACING_V2` as activation flags (line 58). A user with `LANGSMITH_TRACING_V2=true` would have their var silently deleted with no tracing activated. Either check all 4 as valid activation flags, or remove the unchecked ones from the delete list.

Co-reviewed by Claude Sonnet 4.6

packages/components/src/tracingEnv.ts
```diff
@@ -0,0 +1,111 @@
...
+    if (!apiKey) return undefined
+
+    const endpoint = getEnvironmentVariable('LANGSMITH_ENDPOINT') ?? getEnvironmentVariable('LANGCHAIN_ENDPOINT')
+    const projectName = getEnvironmentVariable('LANGSMITH_PROJECT') ?? getEnvironmentVariable('LANGCHAIN_PROJECT')
+
+    // the four LangChain tracing-flag env vars are deleted from `process.env`. Flowise owns tracing
+    // emission from this point on; leaving the flags set would letLangChain's auto-tracer fire on
+    // every `.invoke()`/`.call()` and produce orphan top-level runs next to the manually-emitted
+    // parent/child RunTree.
+    for (const k of LANGCHAIN_TRACING_FLAG_VARS) delete process.env[k]
```
jocelynlin-wd, line 71: This function violates Command-Query Separation — it's named like a query (`get...`) but mutates global state (`process.env`) as a side effect. Callers don't expect a read function to delete env vars. Consider splitting into two functions: a pure `readLangSmithEnvConfig()` that only reads, and a separate `clearLangChainTracingFlags()` called explicitly by the caller. This also makes the side effect visible at the call site rather than hidden inside a getter.

Co-reviewed by Claude Sonnet 4.6

packages/components/src/tracingEnv.ts
```diff
@@ -0,0 +1,111 @@
...
+        const cached = tracingEnvConfigCache.get(name)
+        if (cached) return cached.value
+        const value = resolver()
+        tracingEnvConfigCache.set(name, { value })
+        return value
+    }
+}
+
+/** @internal Test-only: drop cached env-var tracing configs so a subsequent call re-reads env. */
+export const resetTracingEnvCache = (): void => {
```
jocelynlin-wd, line 36: `resetTracingEnvCache` exists only because `tracingEnvConfigCache` is hidden module-level global state. Exporting a test-only reset function is a sign the design is hard to test cleanly. Consider dependency injection — accept the cache or resolver as a parameter — so tests can pass a fresh instance without needing a reset hook.

Co-reviewed by Claude Sonnet 4.6

packages/components/src/tracingEnv.ts
```diff
@@ -0,0 +1,111 @@
...
+]
+
+export const tracingEnvEnabled = (): boolean => TRACING_ENV_PROVIDERS.some((p) => p.getEnvConfig() !== undefined)
+
+/**
+ * Merge env-var-enabled tracing providers into the analytic map. UI config wins when both sources
+ * activate the same provider. Returns a credentialData map for env-injected providers so the loop
+ * can skip DB credential loading for those entries.
+ */
+export const applyEnvTracingProviders = (
```
jocelynlin-wd, line 97: `applyEnvTracingProviders` implies mutation, but this is a pure transform — it returns a new object and explicitly does not mutate its input (there's even a test verifying this). A name like `mergeEnvTracingProviders` or `resolveTracingProviders` better reflects what it does.

Co-reviewed by Claude Sonnet 4.6

jocelynlin-wd (commented):
Thanks for the fix — the env-var tracing approach and the chainRun bug fix are both solid additions.

One thing worth flagging for maintainability: the LangSmith env var names are currently spread across four separate places — which vars activate tracing (`getLangSmithEnvConfig`), which to delete (`LANGCHAIN_TRACING_FLAG_VARS`), which to read for config (inline reads for apiKey/endpoint/projectName), and which to register as CLI flags (`base.ts`). These lists need to stay in sync manually, and they've already drifted slightly in this PR — the delete list has 4 vars but the read logic only checks 2.

A single schema object per provider could help keep everything aligned:
```ts
const LANGSMITH_ENV_SCHEMA = {
    tracingFlags: ['LANGSMITH_TRACING', 'LANGCHAIN_TRACING_V2'],
    apiKey:       ['LANGSMITH_API_KEY',  'LANGCHAIN_API_KEY'],
    endpoint:     ['LANGSMITH_ENDPOINT', 'LANGCHAIN_ENDPOINT'],
    projectName:  ['LANGSMITH_PROJECT',  'LANGCHAIN_PROJECT'],
} as const
```
Every operation — reading, deleting, CLI flag registration — would derive from this one definition, making drift between the lists impossible.

Co-reviewed by Claude Sonnet 4.6

packages/components/src/tracingEnv.ts
mmattu-wd: Will make these changes in FLOWISE-666.

PR #6298, opened 2026-04-27
Fix: Removing not null from validation

abdullah-workday:
Removing `!` since we don't want to enforce not null behavior and keeping it inline with the original code.


abdullah-workday:
Not Needed.

PR #6315, opened 2026-04-29
fix: allow absolute folder paths in Docker via FLOWISE_ALLOWED_FOLDER_PATHS

### LINKED ISSUE #5958: Title: [BUG] "Folder with Files" node is unusable in Docker due to aggressive Path Traversal validation

deepak0x:
Closes #5958

## Problem

The `isPathTraversal()` validator blocks all absolute Unix paths (regex `/^\//`). This makes the **Folder with Files** document loader unusable in Docker, where users mount volumes at paths like `/data/documents`.

The existing escape hatch `PATH_TRAVERSAL_SAFETY=false` disables *all* traversal safety — too broad for operators who only want to allow specific mount points.

## Fix

A new `FLOWISE_ALLOWED_FOLDER_PATHS` env var accepts a comma-separated list of trusted absolute base paths:

```env
FLOWISE_ALLOWED_FOLDER_PATHS=/data/documents,/mnt/shared
```

Absolute Unix paths that fall within one of those base directories pass through `isPathTraversal()` without triggering the block. All other checks (`../` traversal, percent-encoded chars, UNC/Windows paths, null bytes) remain fully enforced.

## What's unchanged

- Default behaviour is identical — absolute paths are still blocked unless `FLOWISE_ALLOWED_FOLDER_PATHS` is set
- `PATH_TRAVERSAL_SAFETY=false` still disables all checks (existing escape hatch preserved)
- `../ ` and encoded traversal through an allowed path (`/data/docs/../../../etc`) is still caught

## Files changed

- `packages/components/src/validator.ts` — `getAllowedFolderPaths()` helper + updated `isPathTraversal()`
- `packages/components/src/validator.test.ts` — 8 new test cases
- `docker/.env.example` + `packages/server/.env.example` — documented the new var next to `PATH_TRAVERSAL_SAFETY`

### DIFF
--- a/docker/.env.example
+++ b/docker/.env.example
@@ -195,6 +195,7 @@ JWT_REFRESH_TOKEN_EXPIRY_IN_MINUTES=43200
 # HTTP_DENY_LIST=
 # HTTP_SECURITY_CHECK=true
 # PATH_TRAVERSAL_SAFETY=true
+# FLOWISE_ALLOWED_FOLDER_PATHS= # Comma-separated absolute paths Docker/self-hosted users may read with the Folder node (e.g. /data/docs,/mnt/shared). Prefer this over disabling PATH_TRAVERSAL_SAFETY entirely.
 # CUSTOM_MCP_SECURITY_CHECK=true
 # CUSTOM_MCP_PROTOCOL=sse #(stdio | sse)
 # TRUST_PROXY=true #(true | false | 1 | loopback| linklocal | uniquelocal | IP addresses | loopback, IP addresses)
--- a/packages/components/src/validator.test.ts
+++ b/packages/components/src/validator.test.ts
@@ -60,6 +60,46 @@ describe('isPathTraversal', () => {
     })
 })
 
+describe('FLOWISE_ALLOWED_FOLDER_PATHS allows specific absolute paths', () => {
+    beforeEach(() => {
+        process.env.FLOWISE_ALLOWED_FOLDER_PATHS = '/data/documents,/mnt/shared'
+    })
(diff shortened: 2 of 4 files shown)


deepak0x:
@HenryHengZJ ready for review.

PR #6324, opened 2026-05-01
Fix(agentflow) - Add iteration node full functionality

j-sanaa:
Fixed the following:

- Double click on Iteration Node opens Edit Dialog

image

- Edges connecting the nodes within an Iteration Node are visible
Before:

https://github.com/user-attachments/assets/0df674c3-df39-4977-805e-77603c63b3c5

After:
image

- Node execution status added for Iteration Node

https://github.com/user-attachments/assets/69ef8884-a373-47c3-b0f7-c2b915c093f7

- Fixed errors on deleting Iteration Node

https://github.com/user-attachments/assets/744d29d3-ce37-435d-90d1-9503c60cfbf6

- Fixed color of Edit button in Dark mode:
Before:

https://github.com/user-attachments/assets/af8595aa-f4b1-4b21-90f1-3aa96f3c9f36

After:

https://github.com/user-attachments/assets/5d486267-eed5-477e-9982-2d2e015ce220


packages/agentflow/src/infrastructure/store/AgentflowContext.tsx
```diff
@@ -181,8 +181,19 @@ export function AgentflowStateProvider({ children, initialFlow }: AgentflowState
...
-            const newEdges = state.edges.filter((edge) => edge.source !== nodeId && edge.target !== nodeId)
+            const toDelete = new Set<string>([nodeId])
+            const collectDescendants = (parentId: string) => {
+                state.nodes
+                    .filter((n) => n.parentNode === parentId)
+                    .forEach((child) => {
+                        toDelete.add(child.id)
+                        collectDescendants(child.id)
+                    })
+            }
```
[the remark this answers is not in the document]
j-sanaa: Since the topology can only be 1 level deep (iteration inside iteration is not supported) the optimization does not add much value

packages/server/src/utils/buildAgentflow.ts
```diff
@@ -1349,11 +1359,10 @@ const executeNode = async ({
                     }
                 }
 
-                // Update the output with combined results
                 results.output = {
                     ...(results.output || {}),
-                    iterationResults,
-                    content: iterationResults.join('\n')
+                    iterationResults, // full record (errors included) for observability
+                    content: successfulIterationResults.join('\n') // successful only — matches what was streamed and saved to chat history
```
jocelynlin-wd: Thanks for fixing the streaming issue. The changes would affect existing-flow in the chat-visibility (errors no longer visible to user in chat) and worth reviewing:
1. Mixed success/failure run: errors silently dropped from chat content
2. All-failed run: chat content goes from "list of error strings" to empty string. If the iteration is the last node and every item errors, the user previously saw error text in the chat; now they see nothing.

Could we keep `content = iterationResults.join('\n')` (errors included, as before) and stream the same items pushed to `iterationResults`? That way the streaming addition is purely additive and doesn't change what the chat ultimately shows.

packages/agentflow/src/features/canvas/hooks/useFlowHandlers.ts
```diff
@@ -71,9 +71,14 @@ export function useFlowHandlers({
...
             }
 
+            const sourceParent = sourceNode?.parentNode
+            const targetParent = targetNode?.parentNode
+            const isWithinIterationNode = sourceParent && targetParent && sourceParent === targetParent
+
             const newEdge = {
                 ...params,
                 type: 'agentflowEdge',
+                ...(isWithinIterationNode && { zIndex: 9999 }),
```
jocelynlin-wd: nit: could we move this magic number to `tokens.zIndex.iterationEdge` (in `packages/agentflow/src/core/theme/tokens.ts`) ? and also add a comment on the new token explaining why it jumps so high?

packages/server/src/utils/buildAgentflow.ts
j-sanaa: So for every other node, the errors are not displayed in the chat content and it is only stored in the execution log and now with this change, iteration node also follows the same behavior as every other node in v2.

jocelynlin-wd (approved):

packages/server/src/utils/buildAgentflow.ts
jocelynlin-wd: Looking at the code path more carefully, the `content` change isn't actually streaming-only — content flows into `result.text` which becomes the HTTP response body for non-streaming callers (`controllers/predictions/index.ts:94-95`) and the persisted chat history row. So a caller hitting `POST /api/v1/prediction/:id` without `streaming: true` would see a different response body for the same iteration flow before vs after this PR.

Could we keep `content = iterationResults.join('\n')` (errors included, matching pre-PR behavior) and just add streaming on top from the same source? That fixes the streaming gap without altering the non-streaming API contract.

j-sanaa: Yes this makes sense to avoid changing results as it is affecting the API responses.

jocelynlin-wd (approved):

packages/server/src/utils/buildAgentflow.ts
j-sanaa: Moving this code change into a separate PR - https://github.com/FlowiseAI/Flowise/pull/6328

PR #6335, opened 2026-05-04
fix: flowise-318

0xi4o:


igor-magun-wd (approved):

packages/server/src/controllers/nodes/index.ts
```diff
@@ -94,7 +94,8 @@ const executeCustomFunction = async (req: Request, res: Response, next: NextFunc
         }
         const orgId = req.user?.activeOrganizationId
         const workspaceId = req.user?.activeWorkspaceId
-        const apiResponse = await nodesService.executeCustomFunction(req.body, workspaceId, orgId)
+        const canViewVariables = !!(req.user?.isOrganizationAdmin || req.user?.permissions?.includes('variables:view'))
```
HenryHengZJ, line 97: what if the flow contains a custom function node, and being embedded in other website.

can this work? I'd assume we automatically treat the req.user as org admin? need to confirm

0xi4o: For embeds, we call the prediction endpoint (which is whitelisted). The prediction handler calls the `node.init` directly and doesn't go through the `/api/v1/node-custom-function` endpoint (which needs authentication - cookie or api key). For API key, isOrganization is always false and the API key needs `variables:view` permissions. For cookie-based (when calling execute in the Custom JS Function node), we check both. There's no elevation of permissions.

HenryHengZJ (approved):

PR #6347, opened 2026-05-05
Feat(agentflow) Node version detection and version upgrade support 

j-sanaa:
FLOWISE-411
Changes Made                                                                                                                                                                       
  - Ported node versioning utilities (isNodeOutdated, getNodeVersionWarning, upgradeNodeData,                          
  getStaleEdgesAfterUpgrade) from the V2 canvas into the AgentFlow SDK
  - Added an orange Sync Nodes FAB on the canvas (top-left, matching V2's floating button placement),
  - Optimised syncNodes: Map-based component lookup, Set-based stale edge filter, early exit when nothing is outdated                                                                                             
  - Full test coverage for all utilities and FAB visibility branches                                                   
                                                                       
  Current Behavior mimicking legacy UI                                                                                                                                                             
  - No version field — node shows a warning badge ("Node outdated — update to latest version X") but the Sync Nodes FAB
   does not appear. The node cannot be auto-upgraded without a baseline version to compare against.
  - Outdated version (node.version < server component version) — node shows a warning badge and the orange Sync Nodes  
  FAB appears on the canvas. Clicking it upgrades all outdated nodes in one action, preserving user-configured inputs, 
  credentials, and labels.
  - Current version — no badge, no FAB.   

Demo: 

https://github.com/user-attachments/assets/575e7e07-45f7-489d-a603-12e6ea8a3171


packages/agentflow/src/Agentflow.tsx
```diff
@@ -101,7 +104,13 @@ function AgentflowCanvas({
...
 
     // Load available nodes
-    const { availableNodes } = useFlowNodes()
+    const { availableNodes, isLoading: isNodesLoading } = useFlowNodes()
+
+    useEffect(() => {
+        if (!isNodesLoading && availableNodes.length > 0) {
+            setComponentNodes(availableNodes)
+        }
+    }, [availableNodes, isNodesLoading, setComponentNodes])
```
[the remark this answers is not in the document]
j-sanaa: availableNodes is a useState so it will not be recomputed on every render

packages/agentflow/src/infrastructure/store/AgentflowContext.tsx
```diff
@@ -356,6 +362,42 @@ export function AgentflowStateProvider({ children, initialFlow }: AgentflowState
...
+    }, [])
+
+    const hasOutdatedNodes = useMemo(
+        () =>
+            state.nodes.some((node) => {
+                const cn = state.componentNodes.find((c) => c.name === node.data.name)
+                return cn ? isNodeOutdated(node.data, cn) : false
+            }),
+        [state.nodes, state.componentNodes]
+    )
```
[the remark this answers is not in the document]
j-sanaa: Map has to be rebuilt for each drag - optimization does not offer any significant gain

packages/agentflow/src/core/utils/nodeVersionUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+    return componentNode.version > nodeData.version
+}
+
+export function getNodeVersionWarning(nodeData: NodeData, componentNode: NodeDataSchema): string | null {
+    if (nodeData.version === undefined && componentNode.version !== undefined) {
+        return `Node outdated\nUpdate to latest version ${componentNode.version}`
+    }
+    if (nodeData.version !== undefined && componentNode.version !== undefined && componentNode.version > nodeData.version) {
+        return `Node version ${nodeData.version} outdated\nUpdate to latest version ${componentNode.version}`
+    }
```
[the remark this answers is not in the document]
j-sanaa: isNodeOutdated was deliberately set to return false for missing nodeData.version so the Sync Nodes FAB doesn't appear for no-version nodes matching V2's implementation

packages/agentflow/src/core/utils/nodeVersionUtils.ts
```diff
@@ -0,0 +1,71 @@
+import type { FlowEdge, NodeData, NodeDataSchema } from '../types'
+
+import { initNode } from './nodeFactory'
+
+export function isNodeOutdated(nodeData: NodeData, componentNode: NodeDataSchema): boolean {
+    if (!nodeData.version) return true
+    return componentNode.version !== undefined && componentNode.version > nodeData.version
+}
```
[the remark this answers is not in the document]
j-sanaa: isNodeOutdated was deliberately set to return false for missing nodeData.version so the Sync Nodes FAB doesn't appear for no-version nodes matching V2's implementation

packages/agentflow/src/infrastructure/store/AgentflowContext.tsx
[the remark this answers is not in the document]
j-sanaa: Map has to be rebuilt for each drag - optimization does not offer any significant gain

jocelynlin-wd:
**Re: scope — demo app and public API surface**

First off, thanks for tackling this — node versioning is a great feature to bring over to the SDK, and the utils port + reducer slice + auto FAB + badges + tests all look clean and faithful to legacy. 🙌

A couple of high-level thoughts on how this lands as an SDK port specifically, that I think would make the change even tighter:

**1. The `NodeVersioningExample` demo doesn't really model consumer code.** Everything version-related (badge, FAB, upgrade-on-click) is wired internally by the SDK — a real integrator's code is just `<Agentflow apiBaseUrl token initialFlow />` and the feature works automatically. The demo's scenario picker / reset / description text is documentation chrome rather than an integration pattern, so there's not much for a consumer to copy from it. The unit tests in `Agentflow.test.tsx > Sync Nodes FAB` already cover the show/hide behavior nicely, so the demo isn't carrying QA weight either. I'd suggest dropping the demo, its `App.tsx` entry, and the `demos/index.ts` export to keep the examples app focused on integration patterns.

**2. The new public API surface goes a bit beyond legacy parity.** Legacy V2 (`packages/ui/src/views/agentflowsv2/Canvas.jsx:441`) treats `syncNodes` as a local function called only by the FAB's `onClick` — it's never exposed externally. This PR adds `AgentFlowInstance.syncNodes()` and `hasOutdatedNodes` to the public ref, plus `onSyncNodes`/`hasOutdatedNodes` to `HeaderRenderProps` and `createHeaderProps`. Without a clear consumer use case yet, I'd lean toward keeping the surface small — once these are public we own their semantics for good (when does a programmatic sync run? how does it interact with unsaved changes? what events fire?).

Suggested trim to keep the version-sync concern fully internal, matching legacy:
- Drop `syncNodes` / `hasOutdatedNodes` from `AgentFlowInstance` and `useAgentflow`
- Drop `onSyncNodes` / `hasOutdatedNodes` from `HeaderRenderProps` and `createHeaderProps`
- Keep them on `AgentflowContextValue` — the FAB inside `Agentflow.tsx` is the only consumer and that's the right scope

Really nice work overall — just want to trim the public surface and the demo before merging so we ship the smallest correct version of this. 🚀

packages/agentflow/src/core/utils/nodeVersionUtils.ts
```diff
@@ -0,0 +1,71 @@
...
+/**
+ * Returns edges that become stale after a node upgrade.
+ * Port of updateOutdatedNodeEdge() from packages/ui/src/utils/genericHelper.js:353-416
+ * (AgentFlow V2 branch only).
+ *
+ * In AgentFlow V2, the valid targetHandle for any input edge is exactly the node ID.
+ * Edges with a suffixed handle (from an old schema-specific input) are stale.
+ * Output edges are never stale in V2.
+ */
+export function getStaleEdgesAfterUpgrade(upgradedData: NodeData, edges: FlowEdge[]): FlowEdge[] {
```
jocelynlin-wd: is this function used anywhere other than in test?  can you clear it up if unused?

packages/agentflow/src/features/canvas/containers/AgentFlowNode.tsx
```diff
@@ -61,10 +62,17 @@ function AgentFlowNodeComponent({ data }: AgentFlowNodeProps) {
...
+        if (componentNode) {
+            const versionWarning = getNodeVersionWarning(data, componentNode)
+            if (versionWarning) messages.push(versionWarning)
+        }
+
         if (data.warning) messages.push(data.warning)
         if (data.validationErrors?.length) messages.push(...data.validationErrors)
         setWarningMessage(messages.join('\n'))
-    }, [data.name, data.version, data.warning, data.validationErrors])
+    }, [data, state.componentNodes])
```
jocelynlin-wd: Minor regression: deps changed from `[data.name, data.version, data.warning, data.validationErrors]` to `[data, ...]`. Since data is a fresh object on most reducer-driven renders, the effect now re-runs more often than needed.
Suggest `[data.name, data.version, data.warning, data.validationErrors, state.componentNodes]` to restore the targeted set.

packages/agentflow/src/Agentflow.tsx
```diff
@@ -274,6 +283,27 @@ function AgentflowCanvas({
                         </StyledFab>
                     )}
 
+                    {/* Sync Nodes FAB - orange button matching V2, shown when outdated nodes exist */}
+                    {!readOnly && hasOutdatedNodes && (
+                        <StyledFab
+                            size='small'
+                            aria-label='sync'
```
jocelynlin-wd: Nit: `aria-label='sync'` is terse. `'Sync nodes'` matches the title and reads better for screen readers

packages/agentflow/src/Agentflow.tsx
```diff
@@ -274,6 +283,27 @@ function AgentflowCanvas({
...
+                            title='Sync Nodes'
+                            onClick={syncNodes}
+                            sx={{
+                                position: 'absolute',
+                                left: 20 + (showDefaultPalette ? 50 : 0) + (enableGenerator ? 50 : 0),
+                                top: 20,
+                                zIndex: tokens.zIndex.canvasButton,
+                                background: 'orange',
```
jocelynlin-wd: Nit: hard-coded color literal. Matches V2 visually, but a token (e.g. a `tokens.colors.warning` equivalent) would fit the rest of the SDK styling. Non-blocking.

packages/agentflow/src/infrastructure/store/agentflowReducer.ts
```diff
@@ -21,7 +21,8 @@ export const initialState: AgentflowState = {
     reactFlowInstance: null,
     editingNodeId: null,
     editDialogProps: null,
-    executionState: null
+    executionState: null,
+    componentNodes: [] as NodeDataSchema[]
```
jocelynlin-wd: Nit: can you confirm if the cast is unnecessary?  `AgentflowState['componentNodes']` seems to have already enforced the type. Would `componentNodes: []` be enough?

packages/agentflow/src/features/canvas/components/AgentflowHeader.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+        render(<AgentflowHeader {...baseProps} isDirty />)
+        expect(screen.getByText(/My Flow \*/)).toBeInTheDocument()
+    })
+
+    it('omits dirty marker when isDirty is false', () => {
+        render(<AgentflowHeader {...baseProps} isDirty={false} />)
+        expect(screen.queryByText(/\*/)).not.toBeInTheDocument()
+    })
+
+    it('does not render a Sync Nodes button (moved to canvas FAB)', () => {
```
jocelynlin-wd: Nit: Since `AgentflowHeader` no longer accepts any version-related prop, this test is asserting a degenerate truth (the header was never going to render that button). Either drop it, or keep as a guardrail with a comment noting the FAB lives on the canvas now — currently the test name implies the header used to render it.

jocelynlin-wd (approved):

PR #6350, opened 2026-05-06
feat - Add changesets for agentflow and observe packages

j-sanaa:
Changes
  - Added .changeset/config.json to configure changeset behavior: packages set to public access, baseBranch set to    
  main, and main Flowise app packages (flowise, flowise-ui, flowise-components, flowise-api) excluded from automated
  versioning                                                                                                          
  - Added .changeset/pre.json to put changeset in pre mode with tag dev, capturing current package versions as the
  baseline so the first release increments correctly from the existing 0.0.0-dev.x versions


.changeset/pre.json
```diff
@@ -0,0 +1,13 @@
+{
```
jocelynlin-wd: we probably don't want pre mode as this could add a lot of noise to the CHANELOG.md while iterating over dev cycles.

.changeset/all-parrots-attend.md
```diff
@@ -0,0 +1,5 @@
+---
+'@flowiseai/agentflow': patch
+---
+
+add changeset to track changes in each release
```
jocelynlin-wd: do we need this file?  I believe we should do batched release, not per PR

jocelynlin-wd (approved):
thanks for adding this!

package.json
```diff
@@ -33,12 +33,16 @@
         "lint-fix": "pnpm lint --fix",
         "quick": "pretty-quick --staged",
         "postinstall": "husky install",
-        "migration:create": "pnpm typeorm migration:create"
+        "migration:create": "pnpm typeorm migration:create",
+        "changeset": "changeset",
+        "changeset:version": "changeset version",
+        "changeset:publish": "changeset publish"
```
jocelynlin-wd, line 39: we should remove this script `changeset:publish` since publish should go via github actions still

PR #6353, opened 2026-05-06
fix(tools): pass tavilyApiKey to TavilySearch constructor

asimar007:
## Root cause
`packages/components/nodes/tools/TavilyAPI/TavilyAPI.ts` constructed `TavilySearch` with the field `apiKey`, but `@langchain/tavily@1.2.0` (and its `BaseTavilyAPIWrapper`) accept the key under the name `tavilyApiKey`. The wrong field was silently dropped, so the wrapper fell back to `process.env.TAVILY_API_KEY` and threw:

> Error in Agent node: Tavily API key not found. Please provide it as an argument or set the TAVILY_API_KEY environment variable.

## Fix
Pass the credential under the correct key name.

```diff
 const config: any = {
-    apiKey: tavilyApiKey,
+    tavilyApiKey,
     topic,
```

## How to verify
1. Add a Tavily API credential in the UI (do not set `TAVILY_API_KEY` env var).
2. Build an Agent flow with the Tavily API tool attached.
3. Run a query — it should now reach Tavily without the "API key not found" error.

## Out of scope
The reporter also mentions a second error after setting `TAVILY_API_KEY` directly (`message.content.map is not a function`). That error is downstream of this fix (in Agent ↔ tool response handling) and looks unrelated to this file, so I've kept this PR scoped to the clear credential-passing bug. Happy to open a follow-up if maintainers can point me at the right area.

Fixes #6343


packages/components/nodes/tools/TavilyAPI/TavilyAPI.ts
```diff
@@ -171,7 +171,7 @@ class TavilyAPI_Tools implements INode {
         const excludeDomains = nodeData.inputs?.excludeDomains as string
 
         const config: any = {
```
HenryHengZJ: to improve this, can we find the correct types/interface from the library?

asimar007: Good catch — switched to TavilySearchAPIRetrieverFields from the library. Would've caught the original bug at compile time.

HenryHengZJ (approved):

asimar007 (commented):
Updated in 167bd6e to use TavilySearchAPIRetrieverFields instead of any. Thanks for the review!

PR #6360, opened 2026-05-07
fix: access to invited workspace

0xi4o:


packages/server/src/enterprise/controllers/workspace-user.controller.ts
```diff
@@ -55,14 +55,7 @@ export class WorkspaceUserController {
...
             } else if (query.userId) {
-                if (query.userId !== user.id && !userMayManageOrgUsers(user)) {
-                    throw new InternalFlowiseError(StatusCodes.FORBIDDEN, GeneralErrorMessage.FORBIDDEN)
-                }
-                workspaceUser = await workspaceUserService.readWorkspaceUserByOrganizationIdUserId(
-                    user.activeOrganizationId,
-                    query.userId,
-                    queryRunner
-                )
+                workspaceUser = await workspaceUserService.readWorkspaceUserByUserId(query.userId, queryRunner)
```
HenryHengZJ: why? what went wrong

0xi4o: This guard was what was blocking workspace switching. Just reverted it temporarily. I'm working on adding a better guard so we don't reintroduce the IDOR issue.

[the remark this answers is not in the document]
0xi4o: Fixed

0xi4o: Updated the guards to allow self-workspace query while blocking cross-org queries.

HenryHengZJ (approved):

PR #6389, opened 2026-05-15
feat: add x402 payment node for paid API calls

597226617:
## Summary

Adds an **x402 Payment Node** to Flowise that enables AI agent workflows to call paid external APIs using the [x402 HTTP 402 payment protocol](https://github.com/x402-foundation/x402).

### What it does
1. Takes an endpoint URL, request params, and max price as input
2. Calls the endpoint
3. Handles HTTP `402 Payment Required` responses
4. Signs and sends USDC payment on Solana (SPL TransferChecked)
5. Retries with `X-Payment` header
6. Returns the paid API response + transaction metadata

### x402 Protocol
x402 is an open payment protocol for AI agents — pay per API call with on-chain USDC. No API keys, no accounts, no subscriptions. Supported by the Linux Foundation.

### Files Added
- `packages/components/nodes/tools/x402Payment/X402Payment.ts` — Node UI definition
- `packages/components/nodes/tools/x402Payment/core.ts` — Payment logic (x402 flow)
- `packages/components/nodes/tools/x402Payment/payment.svg` — Node icon
- `packages/components/credentials/X402Wallet.credential.ts` — Wallet credential

### Testing
- Tested with local development setup
- Payment envelope generation verified against @acedatacloud/x402-client spec

### Related
- x402 Foundation: https://github.com/x402-foundation/x402
- Pyrimid (x402 payment router): https://pyrimid.ai


TateLyman:
Ran a no-payment protocol read-through of the current x402 node implementation. I did not send payment headers, sign requests, or attempt paid calls. A few merge-blocking points still look worth fixing before this goes into Flowise:

1. The 402 parser expects top-level `maxAmountRequired`, `asset`, `payTo`, `network`, and `scheme`, but common x402 responses put one or more payment requirements under `accepts[]`. If the endpoint returns the standard body shape, `parsePaymentRequired(await res.json())` will reject valid 402 challenges instead of selecting a compatible requirement.

2. `maxAmountRequired` should be treated as an atomic-unit string, not a human USDC decimal. The current Solana path parses it as a number and then multiplies by `1e6` again before `TransferChecked`, which can double-scale the amount. The max-price check has the same unit mismatch because it compares atomic units directly to the user-facing USDC max.

3. `asset` is normally the payment asset address/mint for the selected network, not the literal string `usdc`. The current `paymentReq.asset !== "usdc"` guard will reject many valid x402 challenges, including standard Base/Solana USDC requirements that identify the token by contract/mint.

4. The Base path signs a raw ERC-20 `transfer` transaction, but the common x402 `exact` flow for EVM USDC is an EIP-3009 authorization payload that the resource server/facilitator verifies and settles. A raw transfer transaction envelope will not be accepted by standard x402 verifiers unless this node is intentionally targeting a custom server contract flow.

5. The paid-response path still returns `data: text` without applying `maxOutputLength`, so a paid response can ignore the same output bound that the free path enforces.

Patch order I would use: normalize the challenge selection around `accepts[]`, keep atomic units as `bigint`/string until formatting for display, convert the user max price into asset atomic units using selected requirement decimals, validate `asset` against known USDC mint/contract per network, and use the standard x402 payment payload builder for each supported scheme rather than hand-rolling incompatible envelopes.

597226617:
## x402 V2 Protocol Compliance Fixes

Addressed all 5 merge-blocking issues from @TateLyman's review, plus 3 code quality improvements:

### Merge-Blocking Fixes

1. **✅ Parse \`accepts[]\` array** — \`parsePaymentRequired()\` now checks for V2 \`accepts[]\` first, selects compatible requirement (Solana/Base), falls back to V1 top-level fields. Supports CAIP-2 network IDs.

2. **✅ Atomic units throughout** — \`maxAmountRequired\` is now \`string\` type, kept as atomic units (BigInt). V1 fallback converts decimal→atomic. User's \`maxPrice\` converted to atomic units for comparison.

3. **✅ Asset validation** — Validates against known USDC mint addresses per network (Solana mainnet/devnet, Base mainnet). No more \`!== "usdc"\` string comparison.

4. **✅ EIP-3009 for EVM** — Base path now uses \`transferWithAuthorization\` (EIP-712 typed data signing) instead of raw ERC-20 transfer. Signed authorization included in envelope, NOT broadcast — server/facilitator verifies and settles.

5. **✅ \`maxOutputLength\` on paid responses** — Both free and paid response paths now apply output truncation.

### Code Quality Fixes

6. **Private key format** — Supports base64, base58 (bs58), and JSON array formats with clear error messages
7. **Configurable RPCs** — \`solanaRpcUrl\` and \`baseRpcUrl\` parameters added
8. **Consistent JSON returns** — All paths return JSON (including non-payment responses)

597226617:
Hi @TateLyman 👋 

Just a friendly follow-up — I've addressed all 5 merge-blocking issues from your review in the latest push (commit on May 17). The x402 V2 protocol compliance fixes include:

1. Parse `accepts[]` array (V2 protocol)
2. Use atomic units for Solana (lamports) and EVM (wei)
3. EIP-3009 recovery signatures for gasless Base payments  
4. CAIP-2 network identifiers throughout
5. BigInt guard on payment amounts + safe JSON parsing

Happy to make any further adjustments if needed!

TateLyman:
Thanks for the update. The protocol shape is much closer now. I did a follow-up code read only; no wallet signing, payment headers, live endpoint calls, or paid calls.

A few issues still look worth fixing before merge:

1. `parseSolanaPrivateKey()` will usually never reach the base58 or JSON-array branches. `Buffer.from(this.privateKey, 'base64')` does not reliably throw for non-base64 input; it can return arbitrary bytes, then `Keypair.fromSecretKey()` fails later with a generic bad-secret-key error. I would validate base64 length/round-trip first, or try JSON-array and base58 before the loose base64 decode.

2. `validateUsdcMint()` lowercases the candidate mint, but the allowlist values are not normalized the same way. That rejects valid supported assets. For Solana, base58 mint addresses are case-sensitive and should be compared exactly. For Base/EVM, either lowercase both sides or checksum-validate then compare case-insensitively.

3. The `accepts[]` selector can choose the wrong network. It returns true if `req.extra?.mint` is any known USDC mint, even when the user selected the other network. If a response lists Solana USDC first and the node is configured for Base, the mint fallback can select Solana anyway. I would first filter by network compatibility, then validate asset/mint inside that network.

4. The V1 fallback ignores `decimals` when converting decimal `maxAmountRequired` (`* 1e6` is hard-coded). If V1 stays supported, use the parsed `decimals`; if you only want USDC, make that explicit in the error/check path.

Those are mostly input-normalization issues. The previous `accepts[]`, atomic-unit, asset-address, EIP-3009, and paid-response truncation fixes are the right direction.

TateLyman:
One more interop issue showed up while comparing this implementation against live public x402 services.

Scope: code read plus public 402 samples only. No wallet signing, payment headers, paid calls, or private endpoints.

The `accepts[]` parser currently reads `compatibleReq.price`, then returns it as `maxAmountRequired`:

```ts
maxAmountRequired: compatibleReq.price,
```

Live x402 accept entries are not consistent on a `price` field. Common shapes I saw today use either:

```json
{ "maxAmountRequired": "0.001", ... }
```

or:

```json
{ "amount": "1000", ... }
```

If an endpoint returns one of those shapes, this code selects the requirement but carries `undefined` into `BigInt(paymentReq.maxAmountRequired)` / amount signing. That means it can still fail on valid 402 challenges even after the earlier `accepts[]` fix.

Suggested patch:

- Normalize amount with an explicit helper, e.g. `req.maxAmountRequired ?? req.amount ?? req.price`.
- Treat the value as atomic units only when the field semantics are atomic. If a service uses decimal-style `maxAmountRequired` like `"0.001"`, convert by `decimals`; if it uses atomic `amount` like `"1000"`, keep it as-is. The safest path is to follow the x402 package schema/client helper instead of inferring from arbitrary marketplace fields.
- Add tests with both shapes: `accepts[0].maxAmountRequired = "0.001"` and `accepts[0].amount = "1000"`.

This is separate from the earlier private-key/mint/network issues and is probably worth fixing before merge because the failure happens at the first real paid endpoint.

597226617:
Hi @TateLyman 👋

Just a quick follow-up — the `accepts[]` amount normalization fix (commit e7173eb8, pushed May 26) should address the interop issue you flagged where live services use `maxAmountRequired` or `amount` instead of `price`.

The parser now falls back: `req.maxAmountRequired ?? req.amount ?? req.price`, with `Math.round` guarding against parseFloat precision loss. 16 independent test cases cover decimal/atomic/legacy shapes, edge cases, and priority ordering.

Let me know if anything else needs adjustment before merge!

PR #6420, opened 2026-05-21
Harden ~/.flowise/encryption.key file permissions to 0o600

ghost:
## What

Pass `{ mode: 0o600 }` to `fs.promises.writeFile` when persisting `~/.flowise/encryption.key`, so the master encryption key is created with owner-only-read permissions (`-rw-------`) instead of the umask-default `0o644` (`-rw-r--r--`).

## Why

The encryption key wraps every credential stored in the Flowise database. On a stock Linux / macOS install (umask `022`), the key file currently lands at `0o644` — readable by any other UID on the host. On multi-user systems (shared CI runners, multi-tenant servers, containers that mount the host's `~/.flowise`), that's a credential-store exposure.

Node's `fs.promises.writeFile` does not chmod by default — passing `mode` is the standard way to set the on-creation permission. The same pattern is used across the Node ecosystem for secret files (e.g., npm's `~/.npmrc` once it contains a token, GitHub CLI's hosts.yml, etc.).

This is a one-line defensive change with no API or behavioral impact: existing keys are not touched, only newly-created keys get the tighter mode.

## How to verify

```bash
rm -f ~/.flowise/encryption.key   # so the next start regenerates
npm start                          # or however you boot Flowise
stat -f '%Sp' ~/.flowise/encryption.key   # expect: -rw-------
```

Before this PR: `-rw-r--r--`. After this PR: `-rw-------`.

### DIFF
--- a/packages/server/src/utils/index.ts
+++ b/packages/server/src/utils/index.ts
@@ -1585,6 +1585,16 @@ export const getEncryptionKey = async (): Promise<string> => {
             ? path.join(process.env.SECRETKEY_PATH, 'encryption.key')
             : path.join(getUserHome(), '.flowise', 'encryption.key')
         await fs.promises.writeFile(defaultLocation, encryptKey, { mode: 0o600 })
+        // writeFile's `mode` only applies when the file is newly created; if a previous run left
+        // a file at the same path with a permissive mode, the mode is NOT downgraded. Explicit
+        // chmod after the write ensures 0o600 regardless of pre-existing state. Best-effort: a
+        // chmod failure on non-POSIX filesystems shouldn't block key generation.
+        try {
+            await fs.promises.chmod(defaultLocation, 0o600)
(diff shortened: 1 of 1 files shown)


ghost:
Thanks for the review! Just pushed 7ffd756 addressing the chmod-after-write concern. The existing-file mode-downgrade case is real (writeFile's `mode` option only applies on file creation per Node.js docs), so I added an explicit `fs.promises.chmod(defaultLocation, 0o600)` after the write, wrapped in try/catch to no-op gracefully on non-POSIX filesystems where chmod is meaningless. The parent-directory existence concern is moot in this code path since the `.flowise` directory is created earlier in initialization (see `utils/index.ts` near `mkdir(getUserHome() + '/.flowise', { recursive: true })`) — but I'll add a safety `mkdir` if maintainers want it.

PR #6423, opened 2026-05-22
feature / agent-governance

0xchikku:
- Add governance module with policy engine for tool execution control
- Implement audit logger to track tool proposals, policy decisions, and executions
- Add policy loader to parse and apply governance rules from JSON configuration
- Create governed tool wrapper for policy-gated tool execution
- Add human-in-the-loop (HITL) decision tracking in audit logs
- Include sample agent policies and audit logs for testing and documentation
- Update Agent node to integrate governance checks
- Add sqlite database to gitignore for local development


arian-gogani:
agent governance needs three layers:
1. **policy enforcement** — what the agent is allowed to do (what you're proposing)
2. **evidence generation** — proof that the agent followed the policy
3. **auditability** — an external party can verify the evidence independently

layer 2 is what's typically missing. the governance layer says "this action is allowed" but there's no tamper-evident record of that decision.

nobulex generates Ed25519 signed receipts for every governance decision. each receipt carries the policy_version that was in effect, so an auditor can verify not just what happened, but which rules were active at the time.

github.com/arian-gogani/nobulex

PR #6426, opened 2026-05-23
feat: add governance layer inside AgentExecutor ReAct loop

Shreyashgol:
- Insert policy check between tool resolution and tool.call()
- Add policyLoader, policyChecker, hitl, auditLogger modules
- policies.yaml with allow/deny/escalate rules
- Append-only audit log at project root
- Human-in-the-loop via CLI readline

### DIFF
--- a/packages/components/src/agents.ts
+++ b/packages/components/src/agents.ts
@@ -26,6 +26,10 @@ import { formatLogToString } from '@langchain/classic/agents/format_scratchpad/l
 import { IUsedTool } from './Interface'
 import { getErrorMessage } from './error'
 
+import { checkPolicy } from './governance/policyChecker'
+import { waitForHumanApproval } from './governance/hitl'
+import { writeAuditLog } from './governance/auditLogger'
+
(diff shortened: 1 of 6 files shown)


arian-gogani:
governance inside the AgentExecutor ReAct loop is the right placement. one missing piece: tamper-evident evidence of every governance decision.

when the governance layer allows or denies an action, that decision should be captured in a signed receipt. otherwise the governance check is invisible to downstream auditors.

the receipt carries: agent ID, action type, scope, verdict (ALLOW/DENY), policy_version, and an Ed25519 signature. modify anything after the fact and the signature breaks.

for Flowise specifically: a governance callback that emits a nobulex receipt at each ReAct step gives every flow execution a verifiable audit trail. EU AI Act Article 12 requires tamper-evident logs for high-risk AI systems (deadline: August 2).

github.com/arian-gogani/nobulex — open source, MIT licensed.

Shreyashgol:
Yeah , you are right , I stored in the audit.log file .

PR #6429, opened 2026-05-24
fix(server): harden tools and chatmessage route contracts

### LINKED ISSUE #6428: Ambiguous route contracts in tools and chatmessage endpoints accept / even when controller requires /:id
### Describe the bug

Some API routes are currently registered with both / and /:id, but the corresponding controller methods require req.params.id.

The route layer says / is valid, while the controller rejects it with 412 (id not provided). This is easy to miss because the current frontend mostly calls the /:id paths, so regular UI flows still work. However, the contract is inconsistent and can cause future integration issues, confusing errors, and brittle tests.

### To Reproduce

1. Start Flowise locally.
2. Call endpoints that are currently mapped to both / and /:id but need id, for example:
   * PUT /api/v1/tools
   * DELETE /api/v1/tools
   * GET /api/v1/chatmessage
   * DELETE /api/v1/chatmessage
3. Observe that request reaches controller and fails with 412 due to missing id.

#### Example Calls
curl -X DELETE http://localhost:3000/api/v1/tools
curl -X GET http://localhost:3000/api/v1/chatmessage

### Expected behavior

If an endpoint semantically requires an id, only /:id should be registered for that handler.

The / path should not be wired to id-required handlers. This keeps route contract clear and prevents avoidable 412 responses caused by ambiguous route definitions.

### Screenshots

No screenshot needed for this one (behavior can be reproduced with API calls).

### Flow

N/A

### Use Method

pnpm start

### Flowise Version

3.1.2

### Operating System

Windows

### Browser

Other

### Affected routes observed
packages/server/src/routes/tools/index.ts
router.get(['/', '/:id'], ...getToolById)
router.put(['/', '/:id'], ...updateTool)
router.delete(['/', '/:id'], ...deleteTool)

packages/server/src/routes/chat-messages/index.ts
router.get(['/', '/:id'], ...getAllChatMessages)
router.delete(['/', '/:id'], ...removeAllChatMessages)

### Related controllers already require req.params.id
packages/server/src/controllers/tools/index.ts
packages/server/src/controllers/chat-messages/index.ts

### Suggested direction
* Keep list/read-all handlers on / where appropriate.
* Keep item/update/delete handlers on /:id only when id is mandatory.

sahil2448:
### Summary
fixes : #6428 

This PR removes ambiguous / route variants for handlers that require :id, and aligns route definitions with current controller expectations. 
Even though normal frontend flows were already working, these route definitions were still inconsistent and could cause avoidable 412 errors for future integrations or refactors.

### What changed
* Updated tools routes to require :id for id-based handlers:
  * GET /tools/:id (removed / variant for get-by-id)
  * PUT /tools/:id (removed / variant for update)
  * DELETE /tools/:id (removed / variant for delete)
* Updated chatmessage routes similarly:
  * GET /chatmessage/:id (removed / variant)
  * DELETE /chatmessage/:id (removed / variant)
* Normalized abort route to PUT /chatmessage/abort/:chatflowid/:chatid

### Files changed
* packages/server/src/routes/tools/index.ts
* packages/server/src/routes/chat-messages/index.ts

### Why this matters
Controllers for these handlers already require req.params.id.
Previous route setup accepted both / and /:id, which made the API contract ambiguous.
Calling / for id-required handlers could hit controller logic and fail with 412 (id not provided) instead of being clearly non-matching at the route layer.
This change makes behavior predictable and safer for future clients and tests.

### Compatibility and risk
* Low risk, backend-only route-contract hardening.
* No change to controller/service business logic.
* Existing frontend usage remains compatible because UI already calls /:id endpoints.

### Validation
* Verified current frontend API callers still target /:id: 
  * packages/ui/src/api/tools.js
  * packages/ui/src/api/chatmessage.js
* Local commit hooks passed successfully.
* Also tested frontend side -- not giving any issue.

### Checklist
- [x] Kept existing functionality intact for current UI flows
- [x] Limited scope to route contract cleanup
- [x] No unrelated files included in commit

### DIFF
--- a/packages/server/src/routes/chat-messages/index.ts
+++ b/packages/server/src/routes/chat-messages/index.ts
@@ -7,12 +7,12 @@ const router = express.Router()
 // router.post(['/', '/:id'], chatMessageController.createChatMessage)
 
 // READ
-router.get(['/', '/:id'], chatMessageController.getAllChatMessages)
+router.get('/:id', chatMessageController.getAllChatMessages)
 
 // UPDATE
-router.put(['/abort/', '/abort/:chatflowid/:chatid'], chatMessageController.abortChatMessage)
+router.put('/abort/:chatflowid/:chatid', chatMessageController.abortChatMessage)
 
 // DELETE
-router.delete(['/', '/:id'], chatMessageController.removeAllChatMessages)
+router.delete('/:id', chatMessageController.removeAllChatMessages)
 
 export default router
--- a/packages/server/src/routes/tools/index.ts
+++ b/packages/server/src/routes/tools/index.ts
@@ -9,12 +9,12 @@ router.post('/', checkPermission('tools:create'), toolsController.createTool)
 
 // READ
 router.get('/', checkPermission('tools:view'), toolsController.getAllTools)
-router.get(['/', '/:id'], checkAnyPermission('tools:view'), toolsController.getToolById)
+router.get('/:id', checkAnyPermission('tools:view'), toolsController.getToolById)
 
 // UPDATE
-router.put(['/', '/:id'], checkAnyPermission('tools:update,tools:create'), toolsController.updateTool)
+router.put('/:id', checkAnyPermission('tools:update,tools:create'), toolsController.updateTool)
 
 // DELETE
-router.delete(['/', '/:id'], checkPermission('tools:delete'), toolsController.deleteTool)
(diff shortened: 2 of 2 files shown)


sahil2448:
@ maintainers, waiting for Review and Merge if this issue and PR is considerable...!

Issue #6421, opened 2026-05-22
Support/document OpenAI-compatible gateway base URLs for production reliability


michaelmanly:
### Feature Description

````md
## Question

Would maintainers be open to a small PR that documents or improves support for OpenAI-compatible gateway base URLs?

Some production users route LLM calls through a gateway layer so they can manage:

- retries
- fallback routing
- request tracking
- provider timeout handling
- debugging failed provider calls

For example, an OpenAI-compatible client can usually point at a gateway by changing only the API key and base URL:

```ts
import OpenAI from "openai";

const openai = new OpenAI({
  apiKey: process.env.BADGR_API_KEY,
  baseURL: "https://api.aibadgr.com/v1",
});
````

## Why I’m asking

I noticed a few provider-call paths where reliability behavior may currently depend on the underlying provider/client defaults:

* `packages/components/src/followUpPrompts.ts`

  * follow-up prompt generation appears to call providers without a consistent explicit timeout or abort signal
  * transient `429` / `5xx` provider errors may fail immediately
  * the path appears to use the selected provider only, without an optional fallback route

* `packages/components/nodes/chatmodels/ChatOpenAI/ChatOpenAI.ts`

  * OpenAI chat timeout appears to be optional and only set when the user provides one

* `packages/components/src/speechToText.ts`

  * speech-to-text provider calls appear to run without a shared timeout/retry policy

* `packages/components/nodes/retrievers/VoyageAIRetriever/VoyageAIRerank.ts`

  * reranker errors appear to fall back silently to unranked documents, which can make degraded quality harder to debug

These may already be handled elsewhere, so I wanted to ask before opening a larger PR.

## Possible small PR options

Happy to follow whatever maintainers prefer:

1. a small docs PR showing how to configure an OpenAI-compatible gateway base URL
2. a small config improvement for custom OpenAI-compatible base URLs
3. a small timeout/retry improvement in one provider path
4. improved structured error metadata for provider failures
5. no change, if this is outside the project scope

The goal would be optional production reliability support without changing default behavior.

### Feature Category

UI/UX Improvement

miguelmanlyx:
This looks like OpenAI timeout handling. `npx ai-doctor` can point out missing timeout guards and unprotected requests.

PR #6430, opened 2026-05-25
fix: improve follow-up prompt timeout and abort handling

Abhiiishek44:
## Summary

This PR improves follow-up prompt reliability for Ollama-based requests by making the request flow abort-aware and ensuring cleanup across retry attempts.

Changes included:

* wrap the Ollama follow-up prompt request in an abort-aware race so configured timeouts and retries are respected consistently
* ensure `AbortSignal` listeners are always removed after execution to avoid listener accumulation across retries
* add a small runtime validation guard for the parsed follow-up prompt response before returning the result

## Notes

* changes are intentionally scoped only to `packages/components/src/followUpPrompts.ts`
* no behavior changes for unrelated providers
* verified under the repository's Node 20 environment
* local repository-wide TypeScript issues observed during checks appear unrelated to this change

Related: #6421


Abhiiishek44:
Hi @jimjimovich, just checking in on this PR.  
Could you please review it when you get time? Happy to make any changes if needed. Thanks!

packages/components/src/followUpPrompts.ts
```diff
@@ -20,6 +24,85 @@ export interface FollowUpPromptResult {
...
+const isTimeoutError = (error: any): boolean => {
+    const errorCode = error?.code
+    if (error?.name === 'TimeoutError' || error?.name === 'AbortError') return true
+    if (typeof error?.message === 'string' && /timeout|timed out/i.test(error.message)) return true
+    return ['ETIMEDOUT', 'ECONNRESET', 'EAI_AGAIN', 'ENOTFOUND', 'ECONNREFUSED'].includes(errorCode)
+}
+
+const isRetryableError = (error: any): boolean => {
+    const status = getErrorStatus(error)
+    if (status && [429, 500, 502, 503].includes(status)) return true
```
sviat-barbutsa: This is a great improvement, but I think the retry classifier should also include timeout HTTP statuses here, especially `504`.

Right now `429`, `500`, `502`, and `503` are retryable, and timeout-like messages/codes are retryable, but a provider/proxy error like as Axios' common `Request failed with status code 504` can have `status = 504` without a timeout-shaped code or message. In that case this would not retry even though `504 Gateway Timeout` is one of the potential failures this PR is trying to handle.

Could you, please, include `504` in this list, and maybe `408` as well, with a focused test for at least the 504 path?

Abhiiishek44: Thanks for the review. I’ve added 504 and 408 to the retryable HTTP status list and included a focused test for the Axios-style 504 status path.

Abhiiishek44:
Hi @sviat-barbutsa, just following up on this PR.

I’ve addressed your review feedback by adding 504 and 408 to the retryable HTTP status list, along with a focused test for the Axios-style 504 path.

Could you please re-review it when you get time?

Please let me know if any further changes are needed. Thanks!

AbhilashG12:
@Abhiiishek44 Hey! I was looking into issue #6421 and tested your branch on my machine. The retry logic for 504/408 works perfectly!

I noticed the PR might be stuck on that automated bot feedback, so I tried applying it on my end to see what happens:

Ignore the bot about executeWithRetry: The bot tells you to delete the throw new Error at the bottom, but if you do, TypeScript gets mad and won't compile. Your original code was right.

Fix the Ollama block: The bot is right about the Ollama part. You can safely delete all that Promise.race and abortPromise code and just await ollamaClient.chat(...). Since your executeWithRetry function is already handling the timeout, you don't need to do it twice.

I tested that Ollama cleanup and the build passes perfectly. Let me know if you want me to paste the exact code snippet here for you to copy/paste. Awesome work getting this built!

Abhiiishek44:
@AbhilashG12 Thanks for checking this and testing the build locally.

Yes, please paste the snippet here. I’ll review it with my current Ollama block and update the PR accordingly.
Also, thanks for clarifying the executeWithRetry part.

AbhilashG12:
@Abhiiishek44 No problem at all! Here is the cleaned-up block for the OLLAMA case.

You can replace your entire current const chatPromise = ... block (including the try/finally logic and the Promise.race) with this: 

// Ollama client does not accept AbortSignal directly in this SDK call.
// However, executeWithRetry already handles the timeout and abort logic
// via Promise.race at the outer level, making this wrapper redundant.
const response = await ollamaClient.chat({
    model: providerConfig.modelName,
    messages: [
        {
            role: 'user',
            content: followUpPromptsPrompt
        }
    ],
    format: {
        type: 'object',
        properties: {
            questions: {
                type: 'array',
                items: {
                    type: 'string'
                },
                minItems: 3,
                maxItems: 3,
                description: 'Three follow-up questions based on the conversation history'
            }
        },
        required: ['questions'],
        additionalProperties: false
    },
    options: {
        temperature: parseFloat(`${providerConfig.temperature}`)
    }
})

const result = FollowUpPromptType.parse(JSON.parse((response as any).message.content))
return { questions: result.questions }


This effectively removes the redundant nested timeout logic while keeping the actual API call logic exactly as you had it. Once you swap that in, your build should pass the linter and the TS compiler perfectly.

Good luck with the PR!

Abhiiishek44:
Hi @sviat-barbutsa,

In the earlier update, I added `408` and `504` to the retryable status list and included a focused test for the Axios-style `504` case.
In this latest update, I cleaned up the Ollama block based on the automated feedback. I removed the redundant inner timeout handling and kept the outer `executeWithRetry` flow unchanged, since it already handles timeout and retry behavior.

I also kept the fallback `throw new Error(...)` in `executeWithRetry`, as removing it causes TypeScript compilation issues.

Could you please re-review it when you get time?

Thanks!

PR #6437, opened 2026-05-26
Add MCP tool trust verifier hook

RitwijParmar:
## Summary
- add an optional MCPToolkit trustVerifier hook that runs before tools/call dispatch
- pass transport type, server params, server URL, tool name, and tool input into the verifier
- support allow/warn/deny object decisions and boolean allow/deny decisions
- block denied calls before creating a new MCP client so untrusted calls do not reach the server

## Why
This addresses the trust-boundary gap described in #6433 with an opt-in core hook and no behavior change when the verifier is not configured.

## Tests
- npx -y -p node@20 -p pnpm@10.26.0 pnpm --filter flowise-components test -- nodes/tools/MCP/core.test.ts --runInBand
- npx -y -p node@20 -p pnpm@10.26.0 pnpm --filter flowise-components exec eslint nodes/tools/MCP/core.ts nodes/tools/MCP/core.test.ts --max-warnings 0
- npx -y -p node@20 -p pnpm@10.26.0 pnpm --filter flowise-components exec tsc --noEmit

### DIFF
--- a/packages/components/nodes/tools/MCP/core.test.ts
+++ b/packages/components/nodes/tools/MCP/core.test.ts
@@ -76,6 +76,21 @@ describe('MCP tool trust verifier', () => {
         await expect(mcpTool.invoke({})).rejects.toThrow('MCP tool call blocked by trust verifier for "export_data"')
         expect(createClient).not.toHaveBeenCalled()
     })
+
+    it('fails securely when the verifier returns nullish decisions', async () => {
+        const { toolkit, createClient } = createToolkit()
+        toolkit.trustVerifier = jest.fn().mockResolvedValue(null)
+
+        const mcpTool = await MCPTool({
+            toolkit,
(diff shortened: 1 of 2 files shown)


RitwijParmar:
Addressed the Gemini review in 4a59a6e: trust verifier decisions now fail closed for null/undefined, reason handling guards null objects, and a regression test verifies no MCP client is created when the verifier returns null.

arian-gogani:
MCP tool trust verifier is the pre-execution gate. the post-execution layer: a signed receipt for every tool call.

a trusted MCP server can still return unexpected results. the trust check passes the server at admission. the receipt proves what the server actually did during execution. both layers are necessary.

we built this receipt layer. Ed25519 signed, JCS canonical, hash-linked chains. cross-validated across 5 implementations. open source.

github.com/arian-gogani/nobulex

PR #6439, opened 2026-05-27
Add Baidu Qianfan rerank retriever

jimmyzhuu:
## Summary

This PR adds a Baidu Qianfan Rerank Retriever node for reranking vector store retrieval results with Qianfan rerank models.

## Changes

- Add `Baidu Qianfan Rerank Retriever` under Retrievers.
- Add a dedicated `Baidu Qianfan API Key` credential for Qianfan bearer API keys.
- Call Qianfan v2 rerank API directly at `/v2/rerank`.
- Use `bce-reranker-base` as the default rerank model.
- Preserve Flowise output modes:
  - Retriever
  - Document
  - Text
- Attach `relevance_score` metadata to reranked documents.
- Fall back to original documents when the Qianfan API call fails or returns invalid indexes.

## Why

Flowise already supports Baidu Qianfan chat and embeddings. Adding Qianfan rerank support completes a common RAG workflow for Baidu users: retrieve candidates from a vector store, then rerank them with a dedicated cross-encoder reranker before sending context downstream.

## Verification

- `node packages/components/node_modules/jest/bin/jest.js --config packages/components/jest.config.js BaiduQianfanRerank.test.ts BaiduQianfanRerankRetriever.test.ts`
  - 2 test suites passed
  - 13 tests passed
- `git diff --check origin/main...HEAD`
  - passed
- Manual smoke test against Qianfan `/v2/rerank`
  - returned 2 reranked documents
  - ranked the weather-related document first for a weather query
  - returned numeric relevance scores


jimmyzhuu:
Resolved the actionable review items in `84891fc`:

- Added an explicit credential validation error when the Qianfan API key is missing.
- Changed `topN` parsing to use integer parsing and added regression coverage for decimal input.
- Added tests for both cases; the targeted rerank test suite now has 13 passing tests.

For the other two suggestions:

- The compressor still falls back to the original documents on rerank API errors without logging, matching the existing Jina/Cohere/Voyage rerank retriever behavior in this package.
- I did not switch to `handleDocumentLoaderOutput` because this is a retriever node, not a document loader. Existing rerank retrievers format `document` and `text` outputs directly, so this keeps the node consistent with the local retriever patterns.

PR #6467, opened 2026-06-02
Fix frontend axios error handling

sahil2448:
### Linked issue

Closes #6466 

---

### Description

This PR improves frontend API error handling by making the shared `getErrorMessage` utility safer for Axios-style errors and using it in `ToolDialog`.

Previously, `ToolDialog` directly accessed `error.response.data` inside several catch blocks. That can fail when Axios does not include a `response` object, such as during network failures, backend downtime, CORS issues, timeouts, or cancelled requests. In those cases, the error handler could throw another runtime error while trying to display the original API failure.

---

### Changes made

- Updated `packages/ui/src/utils/errorHandler.js`
  - Safely reads `response.data.message`
  - Safely reads `response.data.error`
  - Supports string response bodies
  - Falls back to response status/statusText when response data is unavailable
  - Keeps the existing generic fallback for normal JavaScript errors

- Updated `packages/ui/src/views/tools/ToolDialog.jsx`
  - Replaced repeated direct `error.response.data` access with `getErrorMessage(error)`
  - Applied this to tool export, create, save, and delete error handlers

- Added `packages/ui/src/utils/errorHandler.test.js`
  - Covers Axios response messages
  - Covers Axios `error` fields
  - Covers string response bodies
  - Covers status-only responses
  - Covers normal JavaScript errors without Axios response data

---

### Why this is needed

Instead of each component manually reading nested Axios fields, the shared helper now handles the common API error shapes safely. Components can show useful error messages without risking a secondary crash inside the catch block.

---

### Testing

Ran the focused UI test:

```sh
pnpm -C packages/ui test errorHandler.test.js --runInBand
```

Result:

```txt
Test Suites: 1 passed
Tests: 5 passed
```

---

### Screenshots

Not applicable. This is an error-handling utility change and does not alter the normal UI layout.

---

### Additional notes

This PR keeps the scope focused on `ToolDialog` and the shared error extractor. Other frontend components still contain similar inline Axios error parsing and can be migrated in follow-up PRs.


sahil2448:
Hii @0xi4o ....waiting for PR review, please tell me any changes if needed...!

PR #6472, opened 2026-06-03
fix(chatopenai): support stop in base options

### LINKED ISSUE #2499: [BUG] Additional Parameters did not take effect.

Javi-vahe:
## Summary

- Support `stop` when provided through ChatOpenAI Base Options.
- Extract `stop` from parsed base options and apply it to the model options instead of sending it as a default header.
- Preserve the existing `Stop Sequence` field behavior by only using Base Options `stop` when `obj.stop` is not already set.

Fixes #2499

## Tests

- `git diff --check`

### DIFF
--- a/packages/components/nodes/chatmodels/ChatOpenAI/ChatOpenAI.ts
+++ b/packages/components/nodes/chatmodels/ChatOpenAI/ChatOpenAI.ts
@@ -280,28 +280,23 @@ class ChatOpenAI_ChatModels implements INode {
 
         if (baseOptions) {
             try {
-                parsedBaseOptions = typeof baseOptions === 'object' ? baseOptions : JSON.parse(baseOptions)
+                parsedBaseOptions = typeof baseOptions === 'object' ? { ...baseOptions } : JSON.parse(baseOptions)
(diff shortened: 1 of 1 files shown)


Javi-vahe:
Hi maintainers, I opened this PR to address #2499.

The issue happens because Base Options are currently parsed into the OpenAI configuration/default headers path, so a stop value provided there does not reach the model invocation options. This PR extracts stop from parsed Base Options and applies it to obj.stop, while preserving the existing Stop Sequence field behavior.

Please let me know if you'd prefer a different approach.

Javi-vahe:
Thanks for the review. I updated the implementation to avoid mutating the parsed Base Options object and to avoid applying stop sequences to OpenAI reasoning models.

The PR still removes `stop` from Base Options so it is not sent as a default header, while only applying it to `obj.stop` for non-reasoning models when the dedicated Stop Sequence field is not already set.

PR #6474, opened 2026-06-04
Upgrade FireCrawl document loader to the firecrawl JS SDK v4 (v2 API)

rakshith48:
Upgrades `packages/components/nodes/documentloaders/FireCrawl/FireCrawl.ts` from the hand-rolled v1 axios calls to the official **`firecrawl`** JS SDK v4 (v2 API): `.scrape`/`.crawl`/`.search`/`.extract` + typed responses. Node label/inputs/modes/output shape preserved; legacy `maxDepth` mapped to `maxDiscoveryDepth` for back-compat. Note: HTTP now goes through the SDK client instead of `secureAxiosRequest` — flagged for reviewer preference.

**Verification:** static + SDK-introspection/mocked against the real v2 SDK (`firecrawl-py 4.28.2` / `@mendable firecrawl v4`); not run against the live API. Happy to address review/CI feedback.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/components/nodes/documentloaders/FireCrawl/FireCrawl.ts
+++ b/packages/components/nodes/documentloaders/FireCrawl/FireCrawl.ts
@@ -7,7 +7,7 @@ import Firecrawl, {
     type CrawlOptions,
     type SearchRequest,
     type SearchResultWeb
-} from '@mendable/firecrawl-js'
+} from 'firecrawl'
 import { INode, INodeData, INodeParams, ICommonObject, INodeOutputsValue } from '../../../src/Interface'
 import { getCredentialData, getCredentialParam, handleEscapeCharacters } from '../../../src/utils'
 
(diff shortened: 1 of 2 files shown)


rakshith48:
Thanks for the review! I applied one of these and respectfully declined the others — here's the reasoning:

**Applied — extract `success`/`error` check.** `extract()` returns `Promise<ExtractResponse>`, which genuinely has `success`/`status`/`data`/`error`, so checking for failure there is a real improvement. (I used `response.success === false || response.status === 'failed'` rather than `!response.success`, since `success` is optional and a completed job may omit it — `!response.success` would false-throw.)

**Declined — `success` checks on scrape / search / crawl.** These suggestions assume the **v1** `{ success, data, error }` envelope, which the **v4** SDK removed. In v4 these methods return the data objects directly and **throw** on error:
- `scrape(): Promise<Document>` — no `.success`/`.data` (the doc fields are top-level: `.markdown`, `.metadata`)
- `search(): Promise<SearchData>` — results are under `.web`/`.news`/`.images`, no `.success`
- `crawl(): Promise<CrawlJob>` — has `.status`/`.data`, no `.success`

So `response.success`/`response.data` don't exist on those types — the suggested patches would fail the TypeScript build. The current code reads the correct fields (verified via `tsc` against the real `@mendable/firecrawl-js` v4 types), and failures propagate as thrown errors.

**Declined — `== null` style nits.** Keeping the existing strict-equality checks for consistency; purely cosmetic.

PR #6494, opened 2026-06-08
Fix Flowise 606

yau-wd:
FLOWISE-606


yau-wd:
We’re no longer proceeding with this approach and will be moving forward with an alternative solution in https://github.com/FlowiseAI/Flowise/pull/6499.

PR #6507, opened 2026-06-12
Preserve MCP tool input schema metadata

almahfuz777:
## Summary
- Convert MCP tool JSON input schemas to Zod schemas before passing them to LangChain tools
- Preserve argument descriptions, required fields, enums, arrays, nested objects, and union-like schemas
- Add regression coverage for MCP schema conversion and LangChain tool creation

## Why
Custom MCP tools can expose parameter metadata through JSON Schema. Passing that schema through directly can drop useful tool argument metadata before it reaches the model/tool interface. Converting the schema into the Zod format expected by LangChain preserves that metadata and keeps validation behavior intact.

Fixes #6497.

## Validation
- `corepack pnpm --dir packages/components exec eslint nodes/tools/MCP/core.ts nodes/tools/MCP/core.test.ts --ext ts --report-unused-disable-directives --max-warnings 0`
- `$env:NODE_OPTIONS='--max-old-space-size=4096'; corepack pnpm --dir packages/components exec jest nodes/tools/MCP/core.test.ts --runInBand`


almahfuz777:
Thanks for the review. I addressed the schema converter gaps in `packages/components/nodes/tools/MCP/core.ts`:

- supports non-string and mixed enums
- supports array-based `type` definitions like `['string', 'null']`
- infers object/array schemas from `properties` and `items`
- preserves schema `default` values
- throws on invalid or unsupported schema definitions instead of falling back to `z.any()`

I also added regression coverage in `packages/components/nodes/tools/MCP/core.test.ts`.

Verified with:
`node --max-old-space-size=4096 .\node_modules\jest\bin\jest.js nodes/tools/MCP/core.test.ts --runInBand`

Commit pushed: `6e596e0`

Issue #5501, opened 2025-11-20
Flow execution stops at Condition node when multiple branches converge into single LLM node


Sohorev:
### Describe the bug

Flow execution stops at a Condition node without proceeding to any branch when multiple edges converge into a single LLM node. The flow terminates silently without error.

### To Reproduce

1. Import attached `bug-flow.json` into Flowise
2. Start chat and send: `hi`

### Expected behavior

Condition evaluates to false (not empty), takes output 1, executes `to_operator`

### Actual behavior:
- Flow execution stops at the Condition node (`flow.stopReason is empty`)
- No branch is executed (neither output 0 nor output 1)
- No error is thrown
- Flow terminates silently
- Flow state shows:
  ```json
  {
    "nextAgent": "",
    "intsructions": "",
    "stopReason": "not_empty"
  }
  ```

## Workaround

If you duplicate the `to_operator` LLM node and route different branches to separate (but identical) nodes, the bug disappears. This confirms the issue is related to multiple edges converging into a single node.

### Screenshots

<img width="1574" height="689" alt="Image" src="https://github.com/user-attachments/assets/d1244c7a-2d98-48e2-a3dd-5daf002313a6" />
<img width="1631" height="576" alt="Image" src="https://github.com/user-attachments/assets/4a7c846f-9690-479e-8cc3-4bb13452b11d" />

### Flow

[bug-flow.json](https://github.com/user-attachments/files/23650246/bug-flow.json)

### Use Method

Docker

### Flowise Version

3.0.10

### Operating System

Linux

### Browser

Firefox

### Additional context

[bug-flow.json](https://github.com/user-attachments/files/23650246/bug-flow.json)

Sohorev:
I have no objections if someone fixes this bug; I won't be able to do it myself.

HenryHengZJ:
why do you have the output of conditions merge into same node? then it wouldn't really make sense to have a condition?

<img width="1008" height="712" alt="Image" src="https://github.com/user-attachments/assets/3428d7b9-6f3e-4a1e-8612-3f9de76d51a1" />

thecacho:
> why do you have the output of conditions merge into same node? then it wouldn't really make sense to have a condition?
> 
> ....

@HenryHengZJ I think that is just for simplicity, it's the minimum to show when the bug occurs...

[...]

Issue #5358, opened 2025-10-23
Condition and Human Input nodes stall when their output connects to an input that already has a value/connection


LarsEbt:
## Description
Flows stall at:
- Condition Node
- Human Input Node

When their output is connected to a target node input that already has an existing value or an existing connection. In other words, fan-in works for most nodes, but if the upstream node is a Condition or Human Input node and the downstream input is already populated/connected, execution stops at the Condition/Human Input node.

This looks related to #4660 but is not the same: it also happens with Human Input nodes and specifically when the target input already contains a value/connection.

<img width="1238" height="421" alt="Image" src="https://github.com/user-attachments/assets/77c02be1-9d79-458c-b87e-cd85ec0c9505" />

maximusbirkefeld:
vote up!

paulius-blip:
Hi friend! I had similar issue and found out that this is mainly IF (Condition) nodes problem. When there it is condition after condition in series in my case it went even to the Agent it is not connected to it :D Solved by leaving only one Condition node (problem appears only when they are in series). Hope it helps!

guomengjia618-dot:
Hi maintainers, I’m investigating #5358. In the current Agentflow V2 scheduler, decision routing is reduced from output edges to target node IDs, while waiting inputs are keyed by source node ID. When active and inactive Condition/Human Input edges converge on the same target, the valid route can be suppressed or the target can remain waiting.

I propose a small, backward-compatible server-side change: select active outgoing edges by `sourceHandle`, enqueue each active target once, and preserve the existing dependency and input-merging behavior. I would add regression tests for Condition, resumed Human Input, ordinary fan-in, and inactive branches.

I noticed #6532 changes nearby routing logic. Would you accept a separate PR for the edge-convergence case, or is this already covered by an existing plan?

Issue #4660, opened 2025-06-15
[BUG] Agentflows v2 Condition not works when 2 or more condition connects same node


okurayhan:
Agentflows v2 Condition not works when 2 or more condition connects same node.

![Image](https://github.com/user-attachments/assets/f4da9868-dcc4-4710-8128-251fc0bae07b)

Please see the screenshot. When I write "hi", Report Value Collection Agent not works. When I duplicate Report Value Collection Agent 5 times and connect all nodes to different agents, it works. But I don't want to duplicate same agents.

toi500:
I don't think the Condition Node is intended to be used like that (I could be wrong).

The issue is that you're using the Condition Node as if it has multiple OR paths, but it's really an IF/ELSE IF chain. For your use case (for 2 agents), you would to need to combine all your "false" cases into a single condition That way, you only have two outputs: one for True and one for False.

Edit: Also, since you are using the LLM + Condition Node, you could use the Conditional Agent Node directly for a very reliable route.

![Image](https://github.com/user-attachments/assets/0fbbecd9-6521-4a05-b9a4-c2bd7cdd34a4)

okurayhan:
It works in Agentflows v1 version. I want to combine all cases into one node because I am checking every state value is not empty. When all state values are filled another agent will work.

toi500:
Yeah, I know, in v1 you can also add custom code to the node. Condition Node in V2 is more a simplification of the V1 one, this is why I am unsure if this is a current bug or a limitation.

If what you want is to collect data with one agent first and then to move to the next agent, you also do it like this.

![Image](https://github.com/user-attachments/assets/1ea2a6a0-89ad-4221-810d-9d2fa70f92bd)

okurayhan:
First agent collects more than one data. It collects 5 user input value. So I need to run same agent to get 5 value.

toi500:
Ok...

guivern:
I am experiencing a similar problem here, in the scenarios where the "LomiBot Orders" agent should be launched, the flow stops

![Image](https://github.com/user-attachments/assets/87c1a0eb-aa35-4be1-8fa2-fb6510094c82)

howardtokka:
same thing happens to my flow
it stops at the first node
![Image](https://github.com/user-attachments/assets/3b76b4d5-3a5c-4207-a364-316c914cfa96)

iparker:
Same with loop-nodes with input from different nodes:

[...]

PR #6532, opened 2026-06-19
fix(agentflow): condition node leaks parallel branches when an output fans out to multiple nodes

Kimyongari:
## What & Why

In AgentFlow v2, when a **single Condition output** is connected to **two or more nodes** (parallel branches), selecting a *different* branch still executes one of the nodes wired to the unselected branch.

### Root cause

`determineNodesToIgnore` in `packages/server/src/utils/buildAgentflow.ts` uses `edges.find()` to locate the edge for each unfulfilled condition output:

```ts
for (const index of unfulfilledIndexes) {
    const ignoreEdge = edges.find((edge) => edge.source === nodeId && edge.sourceHandle === `${nodeId}-output-${index}`)
    if (ignoreEdge) {
        ignoreNodeIds.push(ignoreEdge.target)
    }
}
```

`find()` returns only the **first** matching edge. When one output (`-output-${index}`) fans out to multiple targets, only the first target is added to `ignoreNodeIds`. Downstream, child nodes are executed unless they are in that list (`if (ignoreNodeIds.includes(childId)) continue`), so the remaining parallel targets are **never gated** and leak into other branches.

Because the leaked node depends on edge ordering, the same bug also surfaces as the opposite symptom (a matched branch appearing to "stop"), which matches the reports in the linked issues.

## Reproduction

- Start → Condition
- Condition `output-0` → **LLM A** and **LLM B** (two edges on the same output)
- Condition `output-1` → **LLM C**

Expected: selecting `output-1` runs only **LLM C**.
Actual: **LLM C** *and* one of **LLM A / LLM B** both run.

Minimal flow JSON used to reproduce:

[Bug Test Agents.json](https://github.com/user-attachments/files/29121328/Bug.Test.Agents.json)

🎥 **Before (bug)** — selecting `output-1` wrongly runs an `output-0` node too:

before

## Fix

Use `edges.filter()` and ignore **every** target connected to an unfulfilled output:

```ts
for (const index of unfulfilledIndexes) {
    const ignoreEdges = edges.filter(
        (edge) => edge.source === nodeId && edge.sourceHandle === `${nodeId}-output-${index}`
    )
    for (const ignoreEdge of ignoreEdges) {
        ignoreNodeIds.push(ignoreEdge.target)
    }
}
```

🎥 **After (fixed)** — selecting `output-1` runs only the correct node:

after

## Related issues

- #4660 — Agentflows v2 Condition not works when 2 or more condition connects same node
- #5501 — Flow execution stops at Condition node when multiple branches converge into single LLM node
- #5358 — Condition / Human Input nodes stall when output connects to an input that already has a value/connection

## Notes

- Scope limited to the AgentFlow v2 condition routing. A similar `find()` pattern exists in the legacy v1 IfElse path (`packages/server/src/utils/index.ts`), but it involves a different node type / routing model and is intentionally left out of this PR.


Kimyongari:
@jocelynlin-wd @0xi4o @HenryHengZJ @yau-wd
Could you please review this PR when you have a chance?

packages/server/src/utils/buildAgentflow.ts
```diff
@@ -829,10 +829,14 @@ async function determineNodesToIgnore(
             .filter((index) => index !== -1)
 
         // Find nodes to ignore based on unfulfilled conditions
+        // Note: a single condition output can fan out to multiple target nodes
+        // (e.g. parallel branches sharing the same `-output-${index}` handle), so we
+        // must collect ALL matching edges. Using find() here would only ignore the
+        // first target and let the remaining parallel nodes leak into other branches.
         for (const index of unfulfilledIndexes) {
-            const ignoreEdge = edges.find((edge) => edge.source === nodeId && edge.sourceHandle === `${nodeId}-output-${index}`)
+            const ignoreEdges = edges.filter((edge) => edge.source === nodeId && edge.sourceHandle === `${nodeId}-output-${index}`)
```
sviat-barbutsa, line 837: Nice fix. I think this deserves a focused regression test because this branch execution bug is easy to bring back accidentally.

The case I would like to highlight in is a condition node where one unfulfilled output handle fans out to two target nodes, while another output is fulfilled. The test should assert that both targets connected to the unfulfilled handle are ignored and neither is enqueued/executed so that would specifically protect the difference between the old `find(...)` behavior and the new `filter(...)` behavior here.

Kimyongari: Thanks for the review, and I appreciate the suggestion. You're right that this branch execution bug is easy to reintroduce, so a focused regression test makes a lot of sense.

I looked into adding one, but determineNodesToIgnore currently lives inside buildAgentflow.ts, which imports the full execution and runtime dependency graph (storage providers and so on) at module load time. Because of that, unit testing the function in isolation would mean either extracting it into its own module or mocking a lot of unrelated internals in the test. Both of those go well beyond the small fix in this PR.

Since that's a more structural change, I'd rather not expand the scope here on my own. @HenryHengZJ, would you prefer I extract the function and add the regression test in this PR, or keep this one minimal and follow up separately? I'm happy to go either way based on your call.

Kimyongari:
@HenryHengZJ
Could you please review this PR when you have a chance?

Kimyongari:
@HenryHengZJ
Could you please review this PR when you have a chance?

PR #6542, opened 2026-06-22
Add Keenable Search tool node (keyless by default)

ilya-bogin-keenable:
Adds a **Keenable Search** tool node, alongside the existing
SearXNG/Tavily/Brave/Exa/Serper search nodes.

The difference from the other search nodes is that it doesn't require an API
key. By default it uses Keenable's keyless public endpoint, so it works out of
the box. Providing a key (in the node config) switches to the authenticated
endpoint and lifts the rate limits.

### Changes

- `packages/components/nodes/tools/KeenableSearch/`: a custom `Tool` (via the
  SSRF-safe `secureFetch`), following the SearXNG node shape. Maps results to
  `{title, link, snippet}` JSON.
- Optional API key, `mode` (pro/realtime), base-URL override, max results.
  `realtime` mode is validated to require a key.

### Config

- API Key: optional. Blank = keyless free tier; set it to lift limits / enable realtime.
- Base URL: optional override (defaults to `https://api.keenable.ai`).

Verified the node's search path against the public endpoint (keyless).

### DIFF
--- a/packages/components/nodes/tools/KeenableSearch/KeenableSearch.ts
+++ b/packages/components/nodes/tools/KeenableSearch/KeenableSearch.ts
@@ -1,0 +1,212 @@
+import { Tool } from '@langchain/core/tools'
+import { INode, INodeData, INodeParams } from '../../../src/Interface'
+import { getBaseClasses } from '../../../src/utils'
+import { secureFetch } from '../../../src/httpSecurity'
+
+const defaultName = 'keenable-search'
+const defaultDesc =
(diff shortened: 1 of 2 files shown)


ilya-bogin-keenable:
Fxed.

ilya-bogin-keenable:
@HenryHengZJ @0xi4o gentle ping on this one. It is additive: a new tool node, no changes to existing nodes, and it works with no API key by default (a key only lifts rate limits). Happy to rebase or split it if that makes review easier.

PR #6552, opened 2026-06-24
fix: preserve credential on document store loader update

putramkti:
## Problem

When updating a document store loader that already has a credential saved, submitting the update form with no credential selected overwrites the existing credential with `undefined`. This causes the loader to lose its authentication on every settings update unless the user re-selects the credential each time.

## Fix

Add a guard to only overwrite `data.credential` with the existing value when the incoming payload does not already contain a credential:

```diff
- if (found.credential) {
+ if (!data.credential && found.credential) {
      data.credential = found.credential
  }
```

This preserves any credential the user explicitly set in the update while still falling back to the stored value when none is provided.

## Scope

- `packages/server/src/services/documentstore/index.ts` (modified, +1/-1)

## Testing

- Verified that updating a document store loader without changing the credential preserves the existing credential value.
- Verified that explicitly setting a new credential in the update form correctly replaces the old one.


putramkti:
Friendly ping — this is ready for review.

PR #6555, opened 2026-06-24
Add Eden AI Gmail Assistant chatflow template

MVS-source:
Adds a marketplace chatflow template that retrieves unread Gmail, summarizes/analyzes the content with Eden AI, and creates a Gmail draft reply.


Activeleds (commented):
Thank you, we also appreciate

MVS-source:
Thanks for taking a look and for the kind words, I appreciate it. I'll take your feedback into account. I'm also going to share this chatflow in Discussions (Show and Tell) since that looks like the better fit for this kind of contribution.

PR #6592, opened 2026-07-02
feat: add LoopQuest human-in-the-loop tool node

TomPhillipsLabs:
## New tool node: LoopQuest Human Review

Adds a tool node for [LoopQuest](https://loopquest.tomphillips.uk) — a human-in-the-loop step for agents. Before a consequential action, the agent calls `request_human_review`; a person reviews it, and the verdict returns to the agent.

### How it works
Flowise chatflows can't natively pause, so the tool runs a **blocking gate**: it creates a review task in `gate` mode (`POST /api/v1/tasks`) and polls `GET /api/v1/tasks/{id}` until it resolves (or a fail-closed timeout escalates), then returns a readable verdict string. A **Monitor** mode creates the review and returns immediately without waiting.

### Files
- `packages/components/nodes/tools/LoopQuest/LoopQuest.ts` — the `INode` tool; returns a `DynamicStructuredTool` (`request_human_review`).
- `packages/components/nodes/tools/LoopQuest/core.ts` — pure helpers (task body + verdict formatting), unit-tested in the source repo.
- `packages/components/nodes/tools/LoopQuest/loopquest.svg` — icon.
- `packages/components/credentials/LoopQuestApi.credential.ts` — API key (`loopQuestApiKey`) + optional `baseUrl`.

### Config
Node inputs: Game (swiper/versus/sorter/detective/fixer/redact/grounding), Mode (gate/monitor), gate timeout, max wait, poll interval. The LLM supplies `content` (+ `title`, and `claim`/`source` for grounding).

### Notes
- Auth is a Bearer API key; verifiable with `GET /api/v1/me`.
- Happy to provide a test API key for review (dedicated, disposable workspace) — see the comment below.
- If CI needs `pnpm-lock.yaml` regenerated for a new credential/node, flag it and I'll follow up.
- Docs: https://loopquest.tomphillips.uk/docs · OpenAPI: https://loopquest.tomphillips.uk/openapi.json

### DIFF
--- a/packages/components/credentials/LoopQuestApi.credential.ts
+++ b/packages/components/credentials/LoopQuestApi.credential.ts
@@ -1,0 +1,33 @@
+import { INodeParams, INodeCredential } from '../src/Interface'
+
+class LoopQuestApi implements INodeCredential {
+    label: string
+    name: string
+    version: number
+    description: string
+    inputs: INodeParams[]
+
+    constructor() {
+        this.label = 'LoopQuest API'
+        this.name = 'loopQuestApi'
+        this.version = 1.0
+        this.description = 'Your LoopQuest workspace API key (Workspaces → API keys).'
+        this.inputs = [
+            {
+                label: 'API Key',
+                name: 'loopQuestApiKey',
+                type: 'password'
+            },
+            {
+                label: 'Base URL',
+                name: 'baseUrl',
+                type: 'string',
+                optional: true,
+                default: 'https://loopquest.tomphillips.uk',
+                description: 'Only change this for a self-hosted LoopQuest deployment.'
+            }
+        ]
+    }
+}
+
+module.exports = { credClass: LoopQuestApi }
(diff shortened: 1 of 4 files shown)


TomPhillipsLabs:
### Test key for review

A dedicated, disposable LoopQuest workspace for validating the node — please feel free to test all paths; I'll delete it after review:

```
lq_EO33Ufr7JRzvd0hFY_nax8iNFizfNKmc
```

Auth is a Bearer API key (the `loopQuestApiKey` credential field). Verify with:
```
GET https://loopquest.tomphillips.uk/api/v1/me
Authorization: Bearer <key>
→ 200 { "org_id": "...", "workspace": "Flowise Review" }
```

TomPhillipsLabs:
Thanks @gemini-code-assist — applied 4 of the 5 in the latest commit:

- ✅ **API key check** — `init` now throws a clear error if the credential's key is missing.
- ✅ **Clamp intervals** — `timeoutSeconds`/`maxWaitSeconds`/`pollSeconds` are `Math.max(1, …)`, so a 0/negative poll interval can't tight-loop.
- ✅ **try/catch** — the whole flow is wrapped (returns `handleErrorMessage(error)` on failure), and the poll loop swallows transient network errors so one blip doesn't abort the wait window.
- ✅ **`handleErrorMessage`** imported from `../../../src/utils`.

Skipped one, with reasoning:

- **`verdictToString` throwing on invalid data** — its `null` return is an intentional *"still pending, keep polling"* sentinel, not a silent default. Changing it to throw / never return `null` would break the poll loop (it treats `null` as "not resolved yet"). Malformed/missing responses are already handled: the poll `try/catch` ignores them and the loop falls through to the fail-closed timeout return. Keeping `null` semantics.

PR #6609, opened 2026-07-11
fix(ui): sync sidebar active state when navigating via logo

iruzen-dono:
## What

Sidebar active menu item remains highlighted after navigating via the Flowise logo back to Chatflows.

## Root cause

The logo uses a `<Link to={config.defaultPath}>` which navigates to `/chatflows` but does not dispatch `MENU_OPEN` to sync the sidebar active state. The active state (`customization.isOpen`) still holds the previously clicked item id.

## Fix

Add a click handler on the logo button that dispatches `MENU_OPEN` with `id: "chatflows"`, so the Chatflows nav item becomes active when navigating via the logo.

## References

Closes #6510

### DIFF
--- a/packages/ui/src/layout/MainLayout/LogoSection/index.jsx
+++ b/packages/ui/src/layout/MainLayout/LogoSection/index.jsx
@@ -1,4 +1,5 @@
 import { Link } from 'react-router-dom'
+import { useDispatch } from 'react-redux'
 
 // material-ui
 import { ButtonBase } from '@mui/material'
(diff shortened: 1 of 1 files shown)


iruzen-dono:
Hi maintainers 👋 tiny UX fix — when navigating via the logo/header, the sidebar highlighted state doesn't update to match the active route. This syncs it, preventing confusion about which page the user is on. Would appreciate a review when possible.

iruzen-dono:
Hi maintainers 👋 following up on this tiny UX fix — the sidebar active state doesn't update when navigating via the logo/header. This syncs it to prevent confusion about which page the user is on. Minimal +1/-1 line change. Would appreciate a review 🙏

PR #6610, opened 2026-07-11
fix(redis): escape all RediSearch special chars in metadata, not just -

iruzen-dono:
## What

Redis vector store `escapeSpecialChars` only escapes the `-` character, but Redis Search reserves many more characters as operators/syntax: `,.<>{}[]"':;!@#$%^&*()-+=~`

When document metadata contains any of these unescaped characters (e.g. `:`, `"`), JSON.parse fails on retrieval.

## Fix

- `escapeSpecialChars`: now escapes all 26 RediSearch reserved characters
- `unEscapeSpecialChars`: now reverses all 26 escaped characters (single regex)

## Reproduction

1. Setup Redis vector store node
2. Upsert document with metadata containing colon or double quote
3. Run similarity search in Playground → JSON error
4. After fix: works

## References

Closes #6598

### DIFF
--- a/packages/components/nodes/vectorstores/Redis/utils.ts
+++ b/packages/components/nodes/vectorstores/Redis/utils.ts
@@ -1,12 +1,37 @@
 import { isNil } from 'lodash'
 
 /*
- * Escapes all '-' characters.
- * Redis Search considers '-' as a negative operator, hence we need
- * to escape it
+ * Escapes all RediSearch special characters.
+ * Redis Search considers these characters as operators or special syntax,
+ * so they must be escaped to be treated as literals.
  */
 export const escapeSpecialChars = (str: string) => {
     return str.replaceAll('-', '\\-')
+        .replaceAll(',', '\\,')
+        .replaceAll('.', '\\.')
+        .replaceAll('<', '\\<')
+        .replaceAll('>', '\\>')
+        .replaceAll('{', '\\{')
+        .replaceAll('}', '\\}')
+        .replaceAll('[', '\\[')
+        .replaceAll(']', '\\]')
+        .replaceAll('"', '\\"')
+        .replaceAll("'", "\\'")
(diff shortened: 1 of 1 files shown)


packages/components/nodes/vectorstores/Redis/utils.ts
```diff
@@ -1,12 +1,37 @@
...
+        .replaceAll('$', '\\$')
+        .replaceAll('%', '\\%')
+        .replaceAll('^', '\\^')
+        .replaceAll('&', '\\&')
+        .replaceAll('*', '\\*')
+        .replaceAll('(', '\\(')
+        .replaceAll(')', '\\)')
+        .replaceAll('+', '\\+')
+        .replaceAll('=', '\\=')
+        .replaceAll('~', '\\~')
```
sviat-barbutsa, line 34: Good catch. I think we’re still missing one Redis query separator: `|`.

Since `buildQuery` uses `|` as the OR operator to join metadata filters, a literal value like **foo | bar** will be incorrectly parsed as two separate alternatives instead of a single metadata value. As noted in the Redis documentation, `|` is a special character that needs to be escaped in tag filters.

Could you, please, update both escapeSpecialChars and unEscapeSpecialChars to include `|`, and add a regression test for a metadata filter value containing a pipe? I think it'll be really useful!

iruzen-dono:
Hi maintainers 👋 this patch escapes all RediSearch special characters in metadata fields (previously only the '-' char was handled). Without this, metadata searches silently return wrong results for values containing punctuation. Would appreciate a review when possible.

iruzen-dono:
Hi maintainers 👋 following up on this PR that escapes all RediSearch special characters in metadata fields (previously only '-' was handled). Without this, metadata searches silently return wrong results. Minimal +1 line change. Would appreciate a review 🙏

PR #6611, opened 2026-07-11
fix(ui): prevent duplicate chatflow creation on rapid save clicks

iruzen-dono:
## What

Prevents multiple chatflows from being created when the Save/Create button is clicked repeatedly.

## Root cause

The save dialog button has no loading state — it fires `onConfirm` on every click regardless of whether a save request is already in flight.

## Fix

1. **SaveChatflowDialog**: Added `loading` prop — when true, the confirm button and Enter key are disabled, and the cancel button is also disabled to prevent closing while saving
2. **CanvasHeader**: Added `flowDialogLoading` state — set to true when save starts, waits for `handleSaveFlow` to resolve (if it returns a promise), then resets

## References

Closes #6502

### DIFF
--- a/packages/ui/src/ui-component/dialog/SaveChatflowDialog.jsx
+++ b/packages/ui/src/ui-component/dialog/SaveChatflowDialog.jsx
@@ -5,7 +5,7 @@ import PropTypes from 'prop-types'
 import { Button, Dialog, DialogActions, DialogContent, OutlinedInput, DialogTitle } from '@mui/material'
 import { StyledButton } from '@/ui-component/button/StyledButton'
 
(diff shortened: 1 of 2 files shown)


iruzen-dono:
Hi maintainers 👋 this is a minimal fix preventing duplicate chatflow creation on rapid save button clicks. The debounce guard is already present server-side but missing at the UI layer — this adds client-side protection. Would appreciate a review when possible.

iruzen-dono:
Hi maintainers 👋 following up on this PR that prevents duplicate chatflow creation on rapid save clicks. Minimal client-side fix complementing the existing server-side debounce — +2 lines. Would appreciate a review when possible 🙏

PR #6627, opened 2026-07-15
Fix Flowise 6602: add HTTP_ALLOW_LIST for trusted SSRF hosts

laraib-W:
Closes #6602
## Summary                                                             
  - Adds `HTTP_ALLOW_LIST` env var — comma-separated hostnames whose SSRF deny-list check is skipped `172.16.0.0/12` and are rejected by `checkDenyList`; the only prior  workaround was `HTTP_SECURITY_CHECK=false`, which disables SSRF protection globally                                                    
  - Applied in both `checkDenyList` and the pinned-agent  `resolveAndValidate` path used by `secureFetch`/`secureAxiosRequest`, so redirect chains remain safe — a hop off an allowed host into an  unrelated private IP is still blocked                                  
  - Entries are exact, case-insensitive hostname matches (no wildcards)
                                                                         
  ### Naming note
  Chose `HTTP_ALLOW_LIST` over the reporter-suggested `MCP_ALLOW_LIST` to mirror the existing `HTTP_DENY_LIST` naming and cover every  SSRF-guarded call site (not only MCP).                      
                                          
  ## Test plan
  - [x] `pnpm --filter flowise-components test -- httpSecurity` — 79/79  pass, including 13 new tests for the allow list                      
  - [ ] Reviewer: verify docker-compose deployment can now reach an MCP sibling at `HTTP_ALLOW_LIST=<service-name>` without setting `HTTP_SECURITY_CHECK=false`                                
  - [ ] Reviewer: confirm redirect from an allowed host to a private IP is still blocked


packages/components/src/httpSecurity.ts
```diff
@@ -128,6 +174,9 @@ export async function checkDenyList(url: string): Promise<void> {
         hostname = hostname.slice(1, -1)
     }
 
+    // Explicitly allowed hostnames bypass the deny list
+    if (httpAllowList.includes(hostname)) return
```
Drabc: For an allow list, it would be ideal to give the option to allow more restrictive urls. So, rather than allowing `example.internal` one could decide to be more restrictive and only allow `http://example.internal:8080/mcp`. Here we can do something like 
```js
 if (isHttpAllowListed(urlObj, httpAllowList)) return
```
that implementation could look something like this 
```ts
interface HttpAllowRule {
  protocol?: 'http:' | 'https:'
  hostname: string
  port?: string
  pathname: string
}

function normalizeHostname(hostname: string): string {
  let normalized = hostname.trim().toLowerCase()

  // URL.hostname may retain brackets around IPv6 addresses.
  if (normalized.startsWith('[') && normalized.endsWith(']')) {
    normalized = normalized.slice(1, -1)
  }

  // Treat fully qualified names with trailing dots as equivalent.
  if (normalized.endsWith('.')) {
    normalized = normalized.slice(0, -1)
  }

  return normalized
}

function normalizePathname(pathname: string): string {
  if (!pathname || pathname === '/') return '/'

  const normalized = pathname.replace(/\/+$/, '')

  return normalized || '/'
}

function getEffectiveHttpPort(url: URL): string | null {
  if (url.protocol === 'http:') {
    return url.port || '80'
  }

  if (url.protocol === 'https:') {
    return url.port || '443'
  }

  return null
}

function isPathAllowed(
  targetPathname: string,
  allowedPathname: string
): boolean {
  const target = normalizePathname(targetPathname)
  const allowed = normalizePathname(allowedPathname)

  // A rule without a path allows every path on the destination.
  if (allowed === '/') return true

  /*
   * Require a path-segment boundary so a rule for "/mcp"
   * does not also allow "/mcp-admin".
   */
  return target === allowed || target.startsWith(`${allowed}/`)
}

function hasExplicitHttpProtocol(entry: string): boolean {
  const normalized = entry.toLowerCase()

  return (
    normalized.startsWith('http://')
    || normalized.startsWith('https://')
  )
}

function hasExplicitPort(entry: string): boolean {
  /*
   * This function is only used for entries without a protocol.
   * Extract the authority before any path, query, or fragment.
   */
  const authority = entry.split(/[/?#]/, 1)[0]

  if (authority.startsWith('[')) {
    const closingBracketIndex = authority.indexOf(']')
    if (closingBracketIndex === -1) return false

    const remainder = authority.slice(closingBracketIndex + 1)

    return remainder.startsWith(':') && remainder.length > 1
  }

  const colonIndex = authority.lastIndexOf(':')

  return colonIndex !== -1 && colonIndex < authority.length - 1
}

function parseHttpAllowRule(entry: string): HttpAllowRule | null {
  const trimmed = entry.trim()

  /*
   * Returning null silently ignores an empty entry and causes it to grant
   * no access. User-defined entries should be validated when submitted or
   * during startup so empty rules can be reported or intentionally removed.
   */
  if (!trimmed) return null

  const hasExplicitProtocol = hasExplicitHttpProtocol(trimmed)

  /*
   * Handle bare IPv4 and IPv6 addresses directly.
   *
   * Examples:
   *   10.20.30.40
   *   ::1
   *   fe80::1
   *
   * Canonicalizing the address ensures equivalent IP representations match.
   * For example, "0:0:0:0:0:0:0:1" and "::1" normalize to the same value.
   *
   * IPv4 addresses with ports and bracketed IPv6 addresses with
   * ports are handled by the URL parser below.
   */
  if (!hasExplicitProtocol && ipaddr.isValid(trimmed)) {
    return {
      hostname: ipaddr.parse(trimmed).toString().toLowerCase(),
      pathname: '/'
    }
  }

  /*
   * URL removes default ports during parsing. For example,
   * parsing "http://example.internal:80" produces an empty
   * URL.port value.
   *
   * Track whether a scheme-less rule explicitly included a port
   * so ":80" is not mistakenly treated as allowing every port.
   */
  const schemeLessRuleHasExplicitPort = (
    !hasExplicitProtocol
    && hasExplicitPort(trimmed)
  )

  let allowed: URL

  try {
    /*
     * Entries without a scheme are hostname or IP rules.
     *
     * Supported examples:
     *   example.internal
     *   example.internal:8080
     *   example.internal/mcp
     *   example.internal:8080/mcp
     *   10.20.30.40:8080/mcp
     *   [fd00::1]:8080/mcp
     *
     * Optional ports and paths restrict the destination.
     * The injected "http://" scheme is used only for parsing.
     * Protocol is enforced only when the entry explicitly
     * includes http:// or https://.
     */
    allowed = new URL(
      hasExplicitProtocol
        ? trimmed
        : `http://${trimmed}`
    )
  } catch {
    /*
     * Returning null silently ignores malformed entries and causes them
     * to grant no access. User-defined entries should be validated when
     * submitted or during startup so parsing errors can be reported.
     */
    return null
  }

  /*
   * Only HTTP and HTTPS destinations are supported.
   *
   * Returning null silently ignores unsupported protocols and causes the
   * entry to grant no access. User-defined entries should be validated when
   * submitted or during startup so unsupported protocols can be reported.
   */
  if (
    allowed.protocol !== 'http:'
    && allowed.protocol !== 'https:'
  ) {
    return null
  }

  /*
   * Credentials are not valid destination authorization rules.
   *
   * Returning null silently ignores the entry and causes it to grant no
   * access. User-defined entries should be validated when submitted or
   * during startup so embedded credentials can be reported clearly.
   */
  if (allowed.username || allowed.password) {
    return null
  }

  /*
   * Queries and fragments are not used for allow-list matching.
   *
   * Returning null silently ignores entries containing either value and
   * causes them to grant no access. User-defined entries should be validated
   * when submitted or during startup so unsupported URL components can be
   * reported clearly.
   */
  if (allowed.search || allowed.hash) {
    return null
  }

  let port: string | undefined

  if (hasExplicitProtocol) {
    /*
     * An explicit protocol with no port restricts the rule to
     * that protocol's default port.
     */
    const effectivePort = getEffectiveHttpPort(allowed)

    /*
     * This should be unreachable because the protocol was validated above.
     * Returning null still fails closed if that invariant changes.
     * Configuration validation should report the entry if this occurs.
     */
    if (!effectivePort) return null

    port = effectivePort
  } else if (schemeLessRuleHasExplicitPort) {
    /*
     * The injected parsing protocol is HTTP, so an explicitly
     * provided port that URL normalized away must have been 80.
     */
    port = allowed.port || '80'
  }

  return {
    protocol: hasExplicitProtocol
      ? allowed.protocol as 'http:' | 'https:'
      : undefined,
    hostname: normalizeHostname(allowed.hostname),
    port,
    pathname: normalizePathname(allowed.pathname)
  }
}

function isHttpAllowListed(
  target: URL,
  allowList: string[]
): boolean {
  const targetPort = getEffectiveHttpPort(target)

  // Non-HTTP targets are never authorized by the HTTP allow list.
  if (!targetPort) return false

  const targetHostname = normalizeHostname(target.hostname)

  return allowList.some((entry) => {
    const allowed = parseHttpAllowRule(entry)

    /*
     * Invalid user-defined entries are silently ignored here and grant
     * no access. They should already have been validated and reported
     * when submitted or during application startup.
     */
    if (!allowed) return false

    if (allowed.hostname !== targetHostname) {
      return false
    }

    // Enforce protocol only when explicitly included in the rule.
    if (
      allowed.protocol
      && allowed.protocol !== target.protocol
    ) {
      return false
    }

    // A rule without a port allows any HTTP or HTTPS port.
    if (allowed.port && allowed.port !== targetPort) {
      return false
    }

    return isPathAllowed(
      target.pathname,
      allowed.pathname
    )
  })
}

Drabc (commented):
Thanks for working on this. Would be really useful!

packages/components/src/httpSecurity.ts
laraib-W: Thanks, implemented in the latest commit, following your design closely.                                               
                                                                
  Supported shapes now range from coarse to strict: `example.internal` · `example.internal:8080/mcp` · `http://example.internal:8080/mcp` · `[fd00::1]:8080/mcp`     
                                                                
  Path matching respects segment boundaries (`/mcp` allows  `/mcp/tools` but not `/mcp-admin`), protocol is only enforced when explicit, and hostnames-only keep their existing behavior so this isn't breaking. Rules with credentials, queries,fragments, or non-http(s) schemes fail closed.
 
Covered by 18 new unit tests and verified end-to-end against a Docker Compose MCP setup.

PR #6640, opened 2026-07-17
fix: honor telemetry disable flag

AkhilTrivediX:
## Summary

Honors `DISABLE_FLOWISE_TELEMETRY=true` when initializing PostHog telemetry.

## Root cause

The variable was exposed through the CLI and deployment configuration, but the telemetry constructor only checked `POSTHOG_PUBLIC_API_KEY`. As a result, an explicitly configured opt-out did not prevent telemetry initialization.

## Changes

- skip PostHog initialization when `DISABLE_FLOWISE_TELEMETRY` is `true`
- add a regression test covering an enabled PostHog key with telemetry explicitly disabled

## Validation

- `pnpm --filter ./packages/server test -- --runInBand src/utils/telemetry.test.ts`
- `pnpm exec prettier --check packages/server/src/utils/telemetry.ts packages/server/src/utils/telemetry.test.ts`
- `pnpm exec eslint packages/server/src/utils/telemetry.ts packages/server/src/utils/telemetry.test.ts`

Fixes #6618

### DIFF
--- a/packages/server/src/utils/telemetry.test.ts
+++ b/packages/server/src/utils/telemetry.test.ts
@@ -70,6 +70,16 @@ describe('utils/telemetry.ts', () => {
             expect(posthogNode.PostHog).not.toHaveBeenCalled()
         })
 
+        it('does not create PostHog when telemetry is explicitly disabled', () => {
+            process.env.POSTHOG_PUBLIC_API_KEY = 'ph-key'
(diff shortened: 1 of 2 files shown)


AkhilTrivediX:
Closing this because an earlier pull request already covers the same issue. Keeping the discussion focused on the existing proposal.

PR #6653, opened 2026-07-22
feat: add SERPdive API tool node

edendalexis:
## What this adds

A new tool node for [SERPdive](https://serpdive.com), a web search API built for AI agents: one call returns extracted, answer-ready page content instead of a list of links to fetch. This gives Flowise users a third search-API option next to Tavily and Exa.

Measured, not asserted: on a public 1,000-question benchmark judged blind by an independent model, SERPdive won 60.7% of decided quality duels against Tavily's default (basic) search while returning 20.2% fewer tokens at the same speed (the advanced tier was not benchmarked). Methodology and per-question results are open: [serpdive-benchmark](https://github.com/edendalexis/serpdive-benchmark). Free tier: 1,000 credits/month, no card.

## Changes

- `packages/components/nodes/tools/SerpdiveAPI/SerpdiveAPI.ts` — node + tool class in one file, following the Searxng pattern (self-contained `Tool` subclass, `secureFetch` from `src/httpSecurity` for the outbound call). Inputs mirror the API contract exactly: model (mako/moby), include answer, max results (1-10). **No new dependency.**
- `packages/components/credentials/SerpdiveApi.credential.ts` — API key as `type: 'password'`, per the credential security rule in CONTRIBUTING.
- `packages/components/nodes/tools/SerpdiveAPI/serpdive.svg` — vector icon.

## Validation

- Typechecked both files against `packages/components/tsconfig.json` (strict) and formatted with the repo's prettier config.
- Live-invoked the node end to end (init with credential, then `tool.call(...)` as an agent would): real search returns results + optional written answer in ~2 s; an invalid key surfaces the API's error message.
- I could not run the full monorepo build locally, so please let CI have the final word; happy to fix anything it flags.

I kept the diff minimal and self-contained. If this node is not something you want to carry, no hard feelings, feel free to close. Also glad to add a docs page in FlowiseDocs if the node is accepted.


edendalexis:
Update while this is open: SERPdive now has a free tier, which changes the calculus for this integration.

`model: "krill"` is **free and unlimited under fair use** — no card, no credits to spend, no trial clock. It returns about 700 tokens per search (roughly half what the usual alternatives send), one request at a time, at low priority. The paid models are unchanged.

Why it matters here: the usual objection to adding a search provider is that reviewers and users need a paid key just to try it. That objection is gone — anyone can run this integration end to end for free, including whoever reviews this PR.

Nothing in this PR needs to change for that: the default stays `mako`, and `krill` is opt-in via the existing model parameter. Happy to add a line to the docs here if you'd like the free tier called out explicitly.

